PR #6, opened 2018-03-01
added responsive <meta> + some description and keywords.

dev-matan-tsuberi:
for mobile friendliness


.gitignore
```diff
@@ -5,4 +5,5 @@ ropsten_pass.txt
 PRIVATE_README.md
 
 *.log*
-*sublime*
\ No newline at end of file
+*sublime*
+yarn*
```
tibetsprague: why hiding yarn stuff? to push us to use npm?

dev-matan-tsuberi: yea, we decided `npm` was more friendly to newcomers, right? (you can still use `yarn` if you want)

PR #16, opened 2018-03-06
Correctly show votes

tibetsprague:
- Show whether current user already voted on each proposal and what their vote was
- In proposal history list show the total DAO reputation at the time the proposal was executed


tibetsprague:
This also removes automatic redeeming of passed proposals

PR #40, opened 2018-03-21
Display errors

dev-matan-tsuberi:
display notification upon any `REJECTED` actions. 
Reset staking loading animation upon failure.

### DIFF
--- a/src/actions/arcActions.ts
+++ b/src/actions/arcActions.ts
@@ -28,6 +28,7 @@ export function getDAOs() {
     const newOrgEvents = daoCreator.InitialSchemesSet({}, { fromBlock: 0 });
     newOrgEvents.get(async (err: Error, eventsArray: any[]) => {
       if (err) {
+        dispatch({type: 'Notification/Show', payload: {message: 'Could not get DAOs: ' + err.message}});
         dispatch({ type: arcConstants.ARC_GET_DAOS_REJECTED, payload: "Error getting new daos from genesis contract: " + err.message });
       }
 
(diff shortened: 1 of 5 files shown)


dev-matan-tsuberi:
@tibetsprague good idea, where is this adopted in the app?
under this convention, we should get rid of `REJECTED` action types. 
shall we also get rid of `PENDING` by regarding actions with no `payload` and no `error` as pending?

tibetsprague:
@fmatan why should we get rid of the REJECTED actions? its still nice to know that that action failed. the errors in the payload are more for displaying to the user

tibetsprague:
also this is not adopted in the app yet

dev-matan-tsuberi:
@tibetsprague 
as described in https://github.com/redux-utilities/flux-standard-action#errors-as-a-first-class-concept
the point of the `error` property is to avoid action types such as `..._SUCCESS` and `..._FAILURE` (in our case `FULFILLED`/`REJECTED`).

tibetsprague:
thats true. well, if you want to do that even bigger refactor go for it :) I think we could start by just adding the errors to the payload but keeping the separate actions, which do still provide better logging. up to you

dev-matan-tsuberi:
if wer'e going to refactor this to use FSA, two other things are:
1. use the `meta` property for stuff like `proposalId` which is shared across `PENDING`/`REJECTED`/`FULFILLED`
2. make actions more typed.

dev-matan-tsuberi:
Let's keep this refactor as an issue or a trello card for now, as it's not urgent and we have other stuff to do. https://trello.com/c/XSFvuSsY

A question: if we follow what you suggested, how would you handle the automatic dismissal of the notification after timeout (which is what `showAlert` does)?

tibetsprague:
sounds good. good question, id have to ponder more

PR #41, opened 2018-03-21
detect gananche network name

dev-matan-tsuberi:
As I said in trello, `ganache-cli` doesn't have a const `networkId`, but arc.js does provide one.

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -11,17 +11,19 @@ export default class Util {
   public static networkName(id: number) {
     switch (id) {
       case 1:
-        return "mainnet"
+        return "Mainnet"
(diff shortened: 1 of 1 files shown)


dev-matan-tsuberi:
[Display which network you are running on in the header](https://trello.com/c/F0qeRC92/105-display-which-network-you-are-running-on-in-the-header)

tibetsprague:
yes but we could detect a large number that looks like a ganache network. for now that seems easiest. or yes we could always use the fixed networkId from arc.js

dev-matan-tsuberi:
I actually don't think it's really needed to detect ganache since it's used only in development.

PR #42, opened 2018-03-22
Cache blockchain on S3

tibetsprague:
Create a script that runs in a loop in the background on Heroku parsing the blockchain data used by Alchemy (kovan for now) and storing it as JSON on S3. This is used to preload the state when connected to Kovan.

A few other things:
- Had to remove the Git Commit in the header functionality because for some reason it broke on Heroku
- Fixed some bugs in the disabling of voting and staking from https://github.com/daostack/alchemy/commit/c2e20dc4e02e02ef642701463abbade89514d62a
- The code for truffle-hdwallet-provider I copied from a repo and converted to typescript because i couldnt figure out how to add ambient declarations for that particular module. We should really fork and create our own module


src/actions/arcActions.ts
```diff
@@ -8,16 +8,32 @@ import * as Redux from "redux";
...
 import * as schemas from "../schemas";
 
-import { IAccountState, IDaoState, IProposalState, IStakeState, ProposalStates, TransactionStates, VoteOptions } from "reducers/arcReducer";
-import { showAlert } from "actions/notificationsActions";
+export function loadCachedState() {
+  return async (dispatch: Redux.Dispatch<any>, getState: Function) => {
+    dispatch({ type: arcConstants.ARC_LOAD_CACHED_STATE_PENDING, payload: null });
+    try {
+      const cachedState = await axios.get('https://s3-us-west-2.amazonaws.com/daostack-alchemy/initialArcState-' + Arc.Config.get('network') + '.json');
+      dispatch({ type: arcConstants.ARC_LOAD_CACHED_STATE_FULFILLED, payload: cachedState.data });
```
dkent600: I'm having trouble figuring out where cachedState.data is being used to initialize the application data.  Can you please point me to it?

src/bg_cache_worker.ts
```diff
@@ -0,0 +1,151 @@
...
+  s3.putObject(s3Params, (err, data) => {
+    if (err) {
+      console.error("Error writing data to S3 = ", err, err.stack);
+    } else {
+      // tslint:disable-next-line:no-console
+      console.log("Successfully wrote cached data for " + network + " to S3. ", data);
+    }
+  });
+
+  setTimeout(cacheBlockchain, 1000);
```
dkent600: How long does the function take to execute?  Just wondering if one second is a good value.  (particularly since the function is still caching the entire blockchain)

dkent600:
What happens if the cached data is requested at the same time it is being created?

src/bg_cache_worker.ts
tibetsprague: maybe 30-40 seconds? basically I just restart it when it ends :)

src/actions/arcActions.ts
tibetsprague: in the reducer https://github.com/daostack/alchemy/pull/42/files#diff-10f59e3a1f2937ef1befd8f188e5345d

tibetsprague:
hmm, well the file will already exist, its just being overwritten. I guess it depends on how well S3 handles writes...

tibetsprague:
... with simultaneous reads

src/actions/arcActions.ts
dkent600: all I see is `return payload`, but don't see what happens with it

tibetsprague: thats how Redux reducers work, that replaces that part of the state :)

dkent600: But where is the payload used?

src/bg_cache_worker.ts
dev-matan-tsuberi: Shouldn't the script create a redux `store` and dispatch actions to the already existing reducers to update it (eg. `STAKE_FULFILLED`, `VOTE_FULFILLED`)?

src/actions/arcActions.ts
dev-matan-tsuberi: @dkent600 [here](https://github.com/daostack/alchemy/pull/42/files#diff-10f59e3a1f2937ef1befd8f188e5345dR124) where the reducer just returns the payload as the new state, discarding the old state.

src/bg_cache_worker.ts
```diff
@@ -0,0 +1,151 @@
...
+import Util from "./lib/util";
+
+const network = Arc.Config.get('network');
+
+if (process.env.NODE_ENV == 'production') {
+  // Use Infura on production
+  const infuraNetwork = network == 'live' ? 'mainnet' : network;
+
+  // The default account in reading from Infura, I'm not totally sure why this is needed (but it is), or what account it should be
+  const mnemonic = process.env.DEFAULT_ACCOUNT_MNEMONIC;
```
dev-matan-tsuberi: Where is this defined?

src/bg_cache_worker.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  const provider = new HDWalletProvider(mnemonic, "https://" + infuraNetwork + ".infura.io/UeW8cwaou03qFgsAHoDP");
```
dev-matan-tsuberi: Where is `DEFAULT_ACCOUNT_MNEMONIC` defined?

tibetsprague: Heroku

src/bg_cache_worker.ts
```diff
@@ -0,0 +1,151 @@
...
+    const voteEventWatcher = genesisProtocol.VoteProposal({}, { fromBlock: lastBlock });
+    const getVoteEvents = promisify(voteEventWatcher.get.bind(voteEventWatcher));
+    const voteEvents = await getVoteEvents();
+    let voteEventArgs;
+    for (let cnt = 0; cnt < voteEvents.length; cnt++) {
+      voteEventArgs = voteEvents[cnt].args;
+      proposalDetails = initialState.proposals[voteEventArgs._proposalId];
+
+      // If voter not already added as a member on the DAO set them up (this should not happen right?)
+      if (!initialState.daos[proposalDetails.daoAvatarAddress].members[voteEventArgs._voter]) {
```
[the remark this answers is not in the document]
tibetsprague: how/why would it do that? we want to import the data as quickly and easily as possible. remember that everything else is running on the client. this is running on the server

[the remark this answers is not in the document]
tibetsprague: oh I see, you are saying create the state object through the reducers. hmm. that might make sense... seems like extra unnecessary work right now :) ill have to ponder if that makes sense as a refactor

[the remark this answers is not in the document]
dev-matan-tsuberi: This can be brittle if we change stuff in the app. If we use the already defined reducers, there can't be inconsistencies between the app's state and this script's state. In terms of performance, there's no difference.

src/bg_cache_worker.ts
dev-matan-tsuberi: So, this is the secret key for the node that's running in infura?

dev-matan-tsuberi: Seems weird that we can just access it

dev-matan-tsuberi: Oh, it's actually the secret key for the account that interacts with the infura rpc.

tibetsprague: no its a mnemonic for a default account. the key is actually embedded in the URL, but really that should also be an ENV variable :) ill do that. what do you mean weird that we can access it? thats how Heroku handles environment variables. how else would it work?

dev-matan-tsuberi: A mnemonic is just a "pretty printer" secret key.
nevermind, all is good :)

PR #44, opened 2018-03-25
fix staking bounds check with bignumber

dev-matan-tsuberi:
There was a bug with the bounds checking of the stakes, due to a `bignumber` issue. This fixes it by properly using `.gt` and `.lt` from the `bignumber.js` library


src/actions/arcActions.ts
```diff
@@ -711,9 +712,9 @@ export function stakeProposal(daoAvatarAddress: string, proposalId: string, pred
       const stakingToken = await StandardToken.at(await votingMachineInstance.contract.stakingToken());
       const balance = await stakingToken.balanceOf(currentAccountAddress);
 
-      const amount = web3.toWei(stake, "ether");
-      if (amount < minimumStakingFee) { throw new Error(`Staked less than the minimum: ${minimumStakingFee}!`); }
-      if (amount > balance) { throw new Error(`Staked more than than the balance: ${balance}!`); }
+      const amount: BigNumber = new BigNumber(web3.toWei(stake, 'ether'));
+      if (amount.lt(minimumStakingFee)) { throw new Error(`Staked less than the minimum: ${Number(web3.fromWei(minimumStakingFee, 'ether'))}!`); }
```
tibetsprague: I created a function Util.fromWei that casts to a number and doesnt requite passing in 'ether'

src/actions/arcActions.ts
```diff
@@ -712,9 +712,9 @@ export function stakeProposal(daoAvatarAddress: string, proposalId: string, pred
       const stakingToken = await StandardToken.at(await votingMachineInstance.contract.stakingToken());
       const balance = await stakingToken.balanceOf(currentAccountAddress);
 
-      const amount: BigNumber = new BigNumber(web3.toWei(stake, 'ether'));
-      if (amount.lt(minimumStakingFee)) { throw new Error(`Staked less than the minimum: ${Number(web3.fromWei(minimumStakingFee, 'ether'))}!`); }
-      if (amount.gt(balance)) { throw new Error(`Staked more than than the balance: ${Number(web3.fromWei(balance, 'ether'))}!`); }
+      const amount: BigNumber = Util.toWei(stake);
+      if (amount.lt(minimumStakingFee)) { throw new Error(`Staked less than the minimum: ${Number(Util.fromWei(minimumStakingFee))}!`); }
```
tibetsprague: can still remove the "Number(..." Util.fromWei does that for you

dev-matan-tsuberi: fixed: https://github.com/daostack/alchemy/pull/44/commits/4e486495a693d62270c25e74dbccc2792cd9cfa3

dkent600: `fromWei` returns a bignumber, not a number.  In some cases javascript will treat a bignumber like a number, but not in all.

In any case,  I think it is better to use `fromWei().toNumber()'.

src/actions/arcActions.ts
```diff
@@ -654,9 +655,9 @@ export function stakeProposal(daoAvatarAddress: string, proposalId: string, pred
       const stakingToken = await StandardToken.at(await votingMachineInstance.contract.stakingToken());
       const balance = await stakingToken.balanceOf(currentAccountAddress);
 
-      const amount = web3.toWei(stake, "ether");
-      if (amount < minimumStakingFee) { throw new Error(`Staked less than the minimum: ${minimumStakingFee}!`); }
-      if (amount > balance) { throw new Error(`Staked more than than the balance: ${balance}!`); }
+      const amount = new BigNumber(Util.toWei(stake));
+      if (amount.lt(minimumStakingFee)) { throw new Error(`Staked less than the minimum: ${Util.fromWei(minimumStakingFee)}!`); }
```
dkent600: In these computations using `.lt` there is no need to cast to a number at all.  I think it would be better that `Utils.fromWei` return a `BigNumber` and let the caller decide whether it needs to cast the result to a number.  (And when doing so, best to use `.toNumber()` rather than `Number()`).

dev-matan-tsuberi: Thought about this, this would also have the benefit of making `toWei`/`fromWei` inverses of each other.
But returning a string is more convenient for debugging, e.g. in the `redux-devtools` you'll see the number as a string instead of this weird `BigNumber` representation.
(Also I don't know if there parts in the code which assume it's a string, and this would screw them up)

dkent600: Your choice, but I think it merits a refactoring, at an appropriate time.  In the long run I think it'll be a better design.

dev-matan-tsuberi: added a todo: https://github.com/daostack/alchemy/pull/44/commits/b5df9a0f38f4b860160736e6baafce21f29c75b8

src/lib/util.ts
```diff
@@ -3,9 +3,9 @@ import { BigNumber } from "bignumber.js";
 
 export default class Util {
 
-  public static fromWei(amount: BigNumber): number {
+  public static fromWei(amount: BigNumber): BigNumber {
     const web3 = Arc.Utils.getWeb3();
-    return Number(web3.fromWei(amount, "ether"));
+    return web3.fromWei(amount, "ether");
```
dkent600: I am pretty sure that `web3.fromWei(amount, "ether")` returns `BigNumber|string`, hence I think you should do:

`return web3.toBigNumber(web3.fromWei(amount, "ether"));`

dkent600: What about `toWei`?  I think it should return BigNumber too.

src/actions/arcActions.ts
```diff
@@ -654,9 +655,9 @@ export function stakeProposal(daoAvatarAddress: string, proposalId: string, pred
       const stakingToken = await StandardToken.at(await votingMachineInstance.contract.stakingToken());
       const balance = await stakingToken.balanceOf(currentAccountAddress);
 
-      const amount = web3.toWei(stake, "ether");
-      if (amount < minimumStakingFee) { throw new Error(`Staked less than the minimum: ${minimumStakingFee}!`); }
-      if (amount > balance) { throw new Error(`Staked more than than the balance: ${balance}!`); }
+      const amount = new BigNumber(Util.toWei(stake));
```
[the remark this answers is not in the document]
dev-matan-tsuberi: `toWei` returns a string, to use `.lt` we need a `BigNumber`.

src/lib/util.ts
dev-matan-tsuberi: `web3.fromWei` only returns a string if you give it a `number | string`, we are giving it a `BigNumber`.

about the type of `toWei`, see my previous comment, added a todo for future refactoring.

PR #46, opened 2018-03-26
Track running ops

dev-matan-tsuberi:
Track pending, failed, and succesful operations and display them as stacked notifications.

This PR:
- Defines a new type safe action type for async operations `actions/async.ts`.
- Defines a new `reducer/operations` module (that supersedes `reducer/notifications`) with a reducer that keeps track of running ops and a middleware that automatically resets failed ops.
- Refactors `components/Notification` to use the new `IOperation` type.
- Refactors the `ARC_STAKE...` actions to use this new action type.

Still left to do: 

- [x] Refactor other actions to also use this framework so they also get displayed (I probably would want to also do this in this PR, but I wanted to first get some feedback before doing the work).


dev-matan-tsuberi:
[Notifications for transactions pending, failed and succeeded](https://trello.com/c/u7XhgvhJ/56-notifications-for-transactions-pending-failed-and-succeeded)

dev-matan-tsuberi:
@dkent600 Not sure. anyway, it's unrelated. It's used to keep track of actions, not to do anything related to blockchain/solidity.

src/actions/arcActions.ts
```diff
@@ -612,30 +609,44 @@ export function voteOnProposal(daoAvatarAddress: string, proposalId: string, vot
...
-        avatarAddress: daoAvatarAddress,
-        proposalId,
-        stake,
-        prediction,
-        stakerAddress: currentAccountAddress,
-        transactionState: TransactionStates.Unconfirmed,
-      },
+    const proposal: IProposalState = getState().arc.proposals[proposalId];
+
+    const meta = {
```
tibetsprague: Do you really need all of these for the meta? seems like proposalId and stakerAddress would be enough to identify, since you shouldn't be able to stake twice at the same time on the same proposal

src/actions/arcActions.ts
```diff
@@ -612,30 +609,44 @@ export function voteOnProposal(daoAvatarAddress: string, proposalId: string, vot
...
+      stake,
+      prediction,
+      stakerAddress: currentAccountAddress,
     };
 
-    dispatch({ type: arcConstants.ARC_STAKE_PENDING, payload });
+    dispatch({
+      type: arcConstants.ARC_STAKE,
+      sequence: AsyncActionSequence.Pending,
+      operation: {
```
tibetsprague: aren't there two steps for staking? so shouldn't we say that here?

src/actions/web3Actions.ts
```diff
@@ -21,7 +20,6 @@ export function initializeWeb3() {
         type: ActionTypes.WEB3_CONNECTION_REJECTED,
         action: "Can't connect to web3",
       };
-      dispatch(showAlert("Can't connect to web3"));
```
tibetsprague: so this will be replaced by something else in this PR?

src/reducers/arcReducer.ts
```diff
@@ -199,46 +201,56 @@ const arcReducer = (state = initialState, action: any) => {
...
-            [payload.stakerAddress]: {
-              stakes : { $unset : [payload.proposalId] },
+      switch (sequence) {
+        case AsyncActionSequence.Pending:
+          return update(state, { daos: {
+            [avatarAddress] : {
+              members: {
+                [stakerAddress]: {
+                  stakes : { [proposalId] : { $set : {
+                    avatarAddress,
```
tibetsprague: shouldn't this all be coming from the payload? and cant you use the spread operator here?

src/reducers/operations.ts
```diff
@@ -0,0 +1,120 @@
...
+        switch (action.sequence) {
+          case 'reset':
+          case 'success':
+            return update(state, { $unset: [hash] });
+          case 'pending':
+            return update(state, {
+              [hash]: {
+                $merge: {
+                  status: OperationsStatus.Pending,
+                  step: state[hash].status === OperationsStatus.Pending ? state[hash].step + 1 : 0,
```
tibetsprague: im a little unclear how we will use this step functionality. steps for us are multiple blockchain transactions, its not like we will be dispatching multiple actions for each one

src/actions/web3Actions.ts
dev-matan-tsuberi: I intend to refactor other actions as well the same as the `ARC_STAKE...` actions in this PR. Just wanted to get some feedback before doing so

dev-matan-tsuberi:
@tibetsprague I used this architecture in another app I wrote which also needed notifications for running operations.

src/reducers/operations.ts
dev-matan-tsuberi: Steps increment for each `pending` action. So if you have multiple transactions, you would dispatch a `pending` action after completing each transaction.

dev-matan-tsuberi:
@tibetsprague BTW, if you'll look at `failureResetter` in `reducers/operations.ts`, that's a simple example of what I mean by encapsulating effects in middlewares.

src/actions/arcActions.ts
dev-matan-tsuberi: Currently both transactions (approve and transfer) happen inside `arc.js` in the `stake` function. So I can't really put something after the approve and before the transfer.
Probably needs some solution in arc.js.

src/actions/arcActions.ts
dev-matan-tsuberi: at a minimum I think what characterizes a staking operation uniquely is:
1. `avatarAddress` - aka on which DAO?
2. `proposalId` - aka on which proposal?
3. `stakerAddress` - aka who is staking?

But I thought adding `prediction` and `stake` also makes sense and if we use a different `payload` type for the `pending` action from the `success` action, it breaks type-safty. (although we can probably invent a more complicated `AsyncAction` type that would work)

src/reducers/arcReducer.ts
dev-matan-tsuberi: It's all available in the `meta`. Sure we can use the speard operator, or better: `immutability-helper` :).

src/actions/arcActions.ts
tibetsprague: but we want to pass in the totalSteps at the beginning (2) so we know how many more there are to come

src/reducers/operations.ts
tibetsprague: and you are saying that we cant really do that right now. seems like we need to ask MetaMask for transaction results somehow

src/actions/arcActions.ts
dev-matan-tsuberi: If I put `totalSteps: 2` and have only one `pending` action it will seem weird that it succeeds while showing that it only did 1 step out of 2. So until we can dispatch an action after the approve and before the transfer, I think it's better leaving it this way.

dev-matan-tsuberi: added `totalSteps: 1`

dev-matan-tsuberi:
refactored `voteOnProposal`(https://github.com/daostack/alchemy/pull/46/commits/9a89ca0c0198dc5c24cce46e2d254734512587eb) and `createProposal`(https://github.com/daostack/alchemy/pull/46/commits/a886bd14f04d1ac21d7aee896e850204c9e388fe)

Is there anything else we want to display a notification for?

src/actions/arcActions.ts
tibetsprague: well might be more confusing to have steps = 1 when there really are two steps, so maybe leave out for now?

src/actions/web3Actions.ts
tibetsprague: but this is a different kind of alert, we will still want those, for example if a proposal passes after voting on we want a notification for that, and actually im not seeing that here

src/reducers/operations.ts
tibetsprague: yeah that would be nice, lets think through how to articulate it to Doug. notifications on completed transactions somehow

tibetsprague:
oh also you didn't answer my concern about 'reset' not being clear, why not use 'cancel'?

src/actions/arcActions.ts
dev-matan-tsuberi: removed in https://github.com/daostack/alchemy/pull/46/commits/efae6596a110cca203d88f089cb58452f384127f

dev-matan-tsuberi:
@tibetsprague `reset` => `cancel` at: https://github.com/daostack/alchemy/pull/46/commits/36d9332920ebf6efbdb6642dc00b43973ffaa76e

dev-matan-tsuberi:
refactored: `WEB3_CONNECT` (https://github.com/daostack/alchemy/pull/46/commits/fab3944e2edcd86b4fe858cac502ed232d6d4832), `ARC_CREATE_PROPOSAL`(https://github.com/daostack/alchemy/pull/46/commits/a886bd14f04d1ac21d7aee896e850204c9e388fe), `ARC_CREATE_DAO`(https://github.com/daostack/alchemy/pull/46/commits/9096b0cff4f75f7414912ce1e643ed1f924999c9)

src/reducers/operations.ts
dev-matan-tsuberi: Not sure what is a good solution for this. A callback? A global `pendingTransactions` function for all transactions? Is this going to be needed anywhere except token approval?

Maybe we should just handle the token approval in alchemy instead of Arc.js.

What do you think @dkent600 @tibetsprague ?

dev-matan-tsuberi:
@tibetsprague 
How can `reputationWhenExecuted` change when you vote on a proposal?

tibetsprague:
@fmatan reputation when executed is set on the proposal when it passes, which happens when  you vote on it. but sounds like we should probably listen for proposal executed events and set it then

PR #53, opened 2018-04-05
watch rep and tokens

### LINKED ISSUE #29: User account balance should automatically update as it changes
User account balance should automatically update as it changes.  My account just received 5 ETH but the balance shown below did not reflect it until I refreshed the browser page.

image

dev-matan-tsuberi:
Resolve #29


tibetsprague:
do we actually need this? when would these events happen without an action from the user first? only if they have Alchemy open in multiple browsers i think. i guess there isn't a big problem with adding this, but maybe we should wait until there's a real use case.

dev-matan-tsuberi:
I don't know, I was just resolving the issue.
Don't you want the rep and tokens to update at the header?

dev-matan-tsuberi:
@tibetsprague I don't know, I was just resolving the issue.
Don't you want the rep and tokens to update at the header?

tibetsprague:
right now we already update the tokens and rep for the DAO and the account when any action is taken. thats why im saying i cant think of a situation right now where this would be valuable. that may change. so maybe lets leave this here for now and not merge yet

dev-matan-tsuberi:
Well, for example, `tokens` doesn't update right after staking.

We can just fix that in `stake(..)`, but I think it's better to have an event handler rather than trying to anticipate the change in every place it could happen:

- We could easily miss a place where it updates (like with staking).
- It ties your code to the implementation in Arc, basically your'e replicating the logic that's already in Arc which means if something changes in Arc you need to update the code in multiple places.
- It does not account for outside actions, like a different user transferring tokens to the current user.

Having an event handler takes care of all of that and makes the code more straightforward.

tibetsprague:
ok, im down to merge this

PR #57, opened 2018-04-17
Proposal execution event 

dev-matan-tsuberi:


src/reducers/arcReducer.ts
```diff
@@ -333,6 +334,26 @@ const arcReducer = (state = initialState, action: any) => {
...
+      const { avatarAddress, proposalId, executionState, decision, reputationWhenExecuted } = payload;
+
+      if (executionState === ExecutionState.PreBoostedBarCrossed || executionState === ExecutionState.BoostedBarCrossed) {
+        return update(state, {
+          proposals: {
+            [proposalId]: {
+              $merge: {
+                reputationWhenExecuted,
+                winningVote: decision,
+                state: ProposalStates.Executed
```
dev-matan-tsuberi: Does anything else need to be updated?

tibetsprague: votesYes and votesNo on the proposal? actually that should be handled by the vote event. also once redeeming is in there could be new redemptions for the current user. ill have to do that. hmm, cant think of anything now

dev-matan-tsuberi:
[watch for proposal executions events](https://trello.com/c/JpBrHwyS/154-watch-for-proposal-executions-events)

PR #60, opened 2018-04-18
remove large error messages

dev-matan-tsuberi:


tibetsprague:
I wonder if we should console.log the full error, mostly for development purposes?

dev-matan-tsuberi:
But I don't want them to appear in the final build. Do we have a webpack variable that's injected into the app that indicates if its a `prod` or `dev` build?

dkent600:
As long as we're in alpha/beta testing mode it might be helpful to keep those console messages.

dev-matan-tsuberi:
Done in: https://github.com/daostack/alchemy/pull/60/commits/e6c9b3adbf7d24357fae85e679b4085cc22ab777

PR #61, opened 2018-04-19
Use `TransactionService` from Arc.js

dev-matan-tsuberi:


src/lib/util.ts
```diff
@@ -32,4 +32,24 @@ export default class Util {
...
+   * @param cb A callback to execute on each pending transaction.
+   */
+  public static onPendingTransactions(topic: string, key: symbol, cb: (txCount: number) => any) {
+    let count = 0;
+    const sub = Arc.TransactionService.subscribe(topic, (topic: string, txEventInfo: any) => {
+      if (txEventInfo.options.key === key && txEventInfo.tx) {
+        cb(txEventInfo.txCount);
+        count++;
+        if (count >= txEventInfo.txCount) {
+          sub.unsubscribe();
```
dkent600: Would be wise to be able to call `unsubscribe` in cases where an error or tx decline occurs before the events have finished firing.

src/lib/util.ts
```diff
@@ -32,4 +32,24 @@ export default class Util {
...
   }
+
+  /**
+   * Subscribes to pending transactions for a specific operation.
+   * Unsubscribes automatically.
+   * @param topic Arc.js TransactionService topic
+   * @param key A unique `key` to look for in `txEventInfo.options`
+   * @param cb A callback to execute on each pending transaction.
+   */
+  public static onPendingTransactions(topic: string, key: symbol, cb: (txCount: number) => any) {
```
dkent600: the tslint rules should require that function return types be declared

src/actions/arcActions.ts
```diff
@@ -797,7 +800,20 @@ export function stakeProposal(daoAvatarAddress: string, proposalId: string, pred
       if (amount.lt(minimumStakingFee)) { throw new Error(`Staked less than the minimum: ${Util.fromWei(minimumStakingFee).toNumber()}!`); }
       if (amount.gt(balance)) { throw new Error(`Staked more than than the balance: ${Util.fromWei(balance).toNumber()}!`); }
 
-      const stakeTransaction = await votingMachineInstance.stake({ proposalId, vote: prediction, amount });
+      const key = Arc.TransactionService.generateInvocationKey('Stake');
```
dkent600: Would it make sense to put `const key = Arc.TransactionService.generateInvocationKey('Stake'); ` inside of `onPendingTransactions`?

src/lib/util.ts
```diff
@@ -32,4 +32,24 @@ export default class Util {
...
+    let count = 0;
+    const sub = Arc.TransactionService.subscribe(topic, (topic: string, txEventInfo: any) => {
+      if (txEventInfo.options.key === key && txEventInfo.tx) {
+        cb(txEventInfo.txCount);
```
dkent600: Could be good for the user to supply `count` to `cb` here in addition to the total count.  Then when you handle the event, in the message you can tell the user how far along they are (like "completed 2nd of 8" transactions)

src/actions/arcActions.ts
dev-matan-tsuberi: done in: https://github.com/daostack/alchemy/pull/61/commits/366f5bd5e619d50fa3ea85893029b5ec18f705ad

src/lib/util.ts
dev-matan-tsuberi: That's already done automatically. Each `dispatch` of a pending operation increments the `step` until the operation terminates with either `success` or `failure`.

src/lib/util.ts
dev-matan-tsuberi: How would you detect this using `TransactionService`?

src/lib/util.ts
dev-matan-tsuberi: I don't think that's always wanted. Though, the return type for this specific function was declared in https://github.com/daostack/alchemy/pull/61/commits/366f5bd5e619d50fa3ea85893029b5ec18f705ad

dev-matan-tsuberi:
[Show progress bar of number of transactions left to sign](https://trello.com/c/KxmaYoeF/119-show-progress-bar-of-number-of-transactions-left-to-sign)

src/lib/util.ts
dkent600: Hmmm, using `TransactionService` it is possible you could check `tsEventInfo.ts.status` to see if the tx succeeded.  According to the specs on `TransactionReceipt`, "'0x0' indicates transaction failure , '0x1' indicates transaction succeeded".  I haven't tried this to see how it works.

However, this would not handle the case where an exception is thrown by the invoked function.

src/lib/util.ts
dkent600: @dev-matan-tsuberi  Just curious:  what are the cases where a return type declaration would not be wanted?

src/lib/util.ts
dkent600: It may be be better to unsubscribe in a `finally` block of a `try` around the function invocation in your code.

dkent600: I think `performAction` has a lot of responsibilities.

My suggestion:

1. revert back to the structure you had before this change.
2. wrap each Arc.js function call in a try/catch/finally (already done I think except for the `finally`).  The code that invokes the Arc.js functions is best suited to directly invoking Arc.js functions and in handling errors.
3. have `onPendingTransactions` return the subscription
4. in your `finally` block, unsubscribe.
5. remove the `unsubscribe` from `onPendingTransactions`.  This will ensure that if for some reason the total count given by Arc.js is wrong, you will unsubscribe in a timely appropriately.

If you want you can abstract-away the `unsubscribe` in a function sibling to `onPendingTransactions`.

How does that sound?

dev-matan-tsuberi: > 1. revert back to the structure you had before this change.

This means that key generation is duplicated and not abstracted away.

> 2. wrap each Arc.js function call in a try/catch/finally (already done I think except for the finally). The code that invokes the Arc.js functions is best suited to directly invoking Arc.js functions and in handling errors.
> 3. have onPendingTransactions return the subscription
> 4. in your finally block, unsubscribe.

We still need an `onError` callback to handle failed transactions in the listener. So moving the `try/catch` outside the function doesn't give you much.

> remove the unsubscribe from onPendingTransactions. This will ensure that if for some reason the total count given by Arc.js is wrong, you will unsubscribe appropriately.

Can you clarify a bit more?

dkent600: >> revert back to the structure you had before this change.

> This means the caller has to know to include the key in the options.

In any case the caller is already jumping through hoops that entail knowledge about this subsystem for tracking transactions, so I don't see it as a great concern that they would have to include `key` in the options.

However can you add `options` as an argument to `Util.onPendingTransactions` which in turn would call `Arc.TransactionService.generateInvocationKey(key)` and poke the result into the `options `object?

src/lib/util.ts
```diff
@@ -32,4 +32,45 @@ export default class Util {
...
+          if (Arc.Utils.getWeb3().toDecimal(info.tx.receipt.status) !== 1) {
+            console.error(info.tx);
+            const err = new Error(`Transaction '${info.tx.receipt.transactionHash}' failed on '${topic}' with options '${JSON.stringify(opts, undefined, 2)}'`);
+            console.error(err);
+            onError(err);
+            sub.unsubscribe();
+          } else {
+            onPending(info.txCount);
+            count++;
+            if (count >= info.txCount) {
```
dkent600: How could count be greater than txCount?

src/lib/util.ts
dkent600: >> remove the unsubscribe from onPendingTransactions. This will ensure that if for some reason the total count given by Arc.js is wrong, you will unsubscribe appropriately.

> Can you clarify a bit more?

I believe the code as you've currently proposed will not handle cases where the total count given by Arc.js is wrong.

1. if the total count given by Arc.js is too small then you will unsubscribe too soon
2. if the total count given by Arc.js is too high then you will never unsubscribe

The way I've proposed I believe you only unsubscribe in one place.  So it would not just be more robust, but also cleaner.

dev-matan-tsuberi: > However can you add options as an argument

I prefer to minimize implicit mutation of arguments. This means the order of calling `onPendingTransactions` and the contract function matters, and if you mix it up by accident it will be very confusing and hard to debug:

```
const options = ...;
const promise = contract.doSomething(options);
const sub = Util.onPendingTransactions(..., options, (txCount) => {
  /* never called since `key` was not included in `options` */
});
const result = await promise;
```

> Do you think the caller will not always receive an exception when transactions fail?

The `try/catch` cannot catch exceptions in the listener since it's asynchronous:
```
try{
someAsyncStuff(() => { 
   throw new Error() 
}); 
} catch(e) { 
  /* This line is never reached */ 
}
```

> I believe the code as you've currently proposed will not handle cases where the total count given by Arc.js is wrong.

That's correct. It's impossible to know when to unsubscribe if the `txCount` is wrong. The best thing you can do is to unsubscribe after a long timeout, hoping there are no more transactions pending.

src/lib/util.ts
dev-matan-tsuberi: Can't hurt to check for `>` too :)

dkent600: Thing is, it is confusing, raises questions in the programmers mind as to whether there is a non-obvious case (like it did with me, and I know more about this code than a newbie would).  I think is best in this case to comment as such, or change the code.

dev-matan-tsuberi: fixed: https://github.com/daostack/alchemy/pull/61/commits/16f87cd8e057c404cd058d247b3d2cd0bc04f1eb

src/lib/util.ts
dev-matan-tsuberi: For example standard `React` functions like `render` etc... where you know what your'e doing and the types just make things more cluttered.

src/lib/util.ts
dkent600: If you can't `await`, then can you do:

```
someAsyncStuff
    .catch(() => { onError(); } )
    .finally(() => { sub.unsubscribe(); })
```

Regarding order of calling:  I don't understand the concern because pretty much everything depends on args being passed in the correct order and on functions being called in the correct order.  I understand it is nice to be able to abstract and centralize such knowledge in a single location, but doing so should result in clean code that doesn't tradeoff too much.  You called this proposed code "a bit ugly" and I felt it was abstracting too much, and is also not as robust as it could be.  So we must question whether this abstraction is really worth avoiding an unlikely coding error.  TypeScript will handle the order of arguments. The order of function calls, and the robustness, would be handled if you follow the catch/finally pattern from above.

I totally understand the desire for abstraction when it makes things cleaner, prettier and easier to code and maintain.  At the very least, I think it needs to be robust.

dev-matan-tsuberi: @tibetsprague What do you think?

PR #63, opened 2018-04-26
Change Arc.js version to get rid of mock DAOs

dev-matan-tsuberi:
This version of arc.js is the same as `.47` but with a a fresh migration of the contracts, so we don't have all the mock DAOs (Skynet thing).


dev-matan-tsuberi:
Arc.js tests are failing for `48.1`

PR #65, opened 2018-05-07
show relative rep with tooltip

dev-matan-tsuberi:
Question: `ViewProposalContainer` contains duplicate code with `ProposalContainer`. Shouldn't we just use `ProposalContainer` from `ViewProposalContainer`?

### DIFF
--- a/src/components/Proposal/ProposalContainer.tsx
+++ b/src/components/Proposal/ProposalContainer.tsx
@@ -17,15 +17,15 @@ import * as css from "./Proposal.scss";
 
 interface IStateProps {
   currentAccountAddress: string;
-  currentAccountRedemptions: IRedemptionState;
-  dao: IDaoState;
-  proposal: IProposalState;
+  currentAccountRedemptions?: IRedemptionState;
+  dao?: IDaoState;
+  proposal?: IProposalState;
 }
 
-const mapStateToProps = (state: IRootState, ownProps: any) => {
+const mapStateToProps = (state: IRootState, ownProps: any): IStateProps => {
   const proposal = state.arc.proposals[ownProps.proposalId];
-  const dao = state.arc.daos[proposal.daoAvatarAddress];
-  const currentAccountRedemptions = dao.members[state.web3.ethAccountAddress] ? dao.members[state.web3.ethAccountAddress].redemptions[proposal.proposalId] : false;
+  const dao = proposal && state.arc.daos[proposal.daoAvatarAddress];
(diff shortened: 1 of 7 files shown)


dev-matan-tsuberi:
[Show relative reputation throughout the app](https://trello.com/c/nk0OCPnd/103-show-relative-reputation-throughout-the-app)

tibetsprague:
@dev-matan-tsuberi yeah I do want to merge those as much as possible. the only thing is we dont really know what ViewProposalContainer will end up looking like, I think it could be significantly different looking. still, its true that they should probably just be the same for now.

src/layouts/HeaderContainer.tsx
```diff
@@ -96,7 +97,11 @@ class HeaderContainer extends React.Component<IProps, null> {
                     ? <div>
                         <AccountBalance tokenSymbol={dao.tokenSymbol} balance={member.tokens} accountAddress={web3State.ethAccountAddress} />
                         &nbsp; | &nbsp;
-                        <span>{member.reputation.toFixed(2).toLocaleString()} {dao.name} reputation</span>
+                        <Tooltip placement="bottom" trigger={["hover"]} overlay={<span style={{minHeight: 'inherit !important'}}>{member.reputation} reputation</span>}>
+                          <span style={{cursor: 'pointer'}}>
```
tibetsprague: a pointer usually means you can click on the element and have it do something, I dont think we want that here

src/components/Proposal/ProposalContainer.tsx
```diff
@@ -114,12 +114,18 @@ class ProposalContainer extends React.Component<IProps, null> {
         rewards.push(proposal.nativeTokenReward + " " + dao.tokenSymbol);
       }
       if (proposal.reputationChange) {
-        rewards.push(proposal.reputationChange + " reputation");
+        rewards.push(
+          <Tooltip placement="top" trigger={["hover"]} overlay={<span style={{minHeight: 'inherit !important'}}>{proposal.reputationChange} reputation</span>}>
+            <span style={{cursor: 'pointer'}}>
```
tibetsprague: a pointer usually means you can click on the element and have it do something, I dont think we want that here

dev-matan-tsuberi: Idk, Feels more natural to me, do you have another suggestion? Or shall I remove it?

tibetsprague: I would just remove it. the tooltip appears immediately anyway right?

dev-matan-tsuberi: Yes

dev-matan-tsuberi:
Should we have a link from each proposal on "Proposals" to it's individual page?

tibetsprague:
we do, on the view-proposal branch.

PR #68, opened 2018-05-11
remove uneeded status check

dev-matan-tsuberi:


dev-matan-tsuberi:
Another thing: fixed another console error in https://github.com/daostack/alchemy/pull/68/commits/5c2cd45d8ac78f01d3e8a98858536d82234acdc2
React was complaining about not having a `key` in the static array used, changed to `<React.Fragment>`.

PR #73, opened 2018-05-14
Fix 'stuck on pending'

dev-matan-tsuberi:
This happens because of https://github.com/daostack/arc.js/issues/215. A work-around is possible until there's a new version. But in certain cases it might be solved with a simple UI fix.

- When staking, disable staking on a different prediction than the one already staked on.
- Is there other cases in which this happens?


dev-matan-tsuberi:
[Staking 100 GENs fails](https://trello.com/c/m3bHnUrm/144-staking-100-gens-fails)

PR #74, opened 2018-05-15
move expired proposals to history

dev-matan-tsuberi:
Question: Should there be a separate section in history for expired proposals vs successfully executed?

### DIFF
--- a/src/selectors/daoSelectors.ts
+++ b/src/selectors/daoSelectors.ts
@@ -20,21 +20,21 @@ const getDaoProposals = createSelector(
   },
 );
 
-export const boostedProposals = createSelector(
+export const createBoostedProposalsSelector = () => createSelector(
   [ getDaoProposals ],
   (proposals) => proposals.filter((proposal: IProposalState) =>
     proposal.state === ProposalStates.Boosted && (+moment() / 1000) <= proposal.boostedTime + proposal.boostedVotePeriodLimit
   )
 );
 
-export const preBoostedProposals = createSelector(
+export const createPreBoostedProposalsSelector = () => createSelector(
   [ getDaoProposals ],
   (proposals) => proposals.filter((proposal: IProposalState) =>
     proposal.state === ProposalStates.PreBoosted && (+moment() / 1000) <= proposal.submittedTime + proposal.preBoostedVotePeriodLimit
   ),
 );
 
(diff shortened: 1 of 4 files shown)


tibetsprague:
@dev-matan-tsuberi I will leave that up to @brodeur to decide if he wants to change the design that way

src/selectors/daoSelectors.ts
```diff
@@ -19,26 +20,31 @@ const getDaoProposals = createSelector(
...
-  );
-};
-
-export const makeDaoExecutedProposalsSelector = () => {
-  return createSelector(
-    [ getDaoProposals ],
-    (proposals) => proposals.filter((proposal: IProposalState) => (proposal.state == ProposalStates.Executed || proposal.state == ProposalStates.Closed)),
-  );
-};
+export const boostedProposals = createSelector(
```
tibetsprague: I used functions because if there are multiple components using the same selector with different proposals (multiple DAOs for example) then it won't memoize correctly: https://github.com/reduxjs/reselect I think we still need to do that to future proof

dev-matan-tsuberi: Correct. Another problem I noticed is that, since the selectors are time-dependent there's no point in memoization anyway. 

Do we actually need this performance benefit in the app? seems to me that it's negligible (unless there are thousands of proposals, in which case I think we should display and load them in pages, i.e. infinite scroll or similar)

What do you think?

tibetsprague: I think we get some benefit from memoization, unless we expect there to be a constant flood of new proposals all the time. And in the future there will likely be multiple components using these selectors, which will make them more valuable. At the very least we need similar functions anyway to filter the proposals, so why not make them selectors :)

dev-matan-tsuberi: But since it's time-dependent, memoization can lead to stale views, that should be updated (because time passes), but are not. I guess that's not such a big concern.

dev-matan-tsuberi: fixed: https://github.com/daostack/alchemy/pull/74/commits/723b55a98ed4cfda68596736869992856a104b5a

tibetsprague: the views should never be stale, reselect invalidates the memoization if the proposal list changes at all

PR #89, opened 2018-05-30
update user's eth in real time

dev-matan-tsuberi:


dev-matan-tsuberi:
[Update user balances in real time as they change](https://trello.com/c/dtpcDNSM/121-update-user-balances-in-real-time-as-they-change)

PR #91, opened 2018-05-31
Lowercase all account addresses used throughout the app

tibetsprague:
To make sure they always match each other

https://trello.com/c/4E9ypJRH

### DIFF
--- a/src/actions/arcActions.ts
+++ b/src/actions/arcActions.ts
@@ -127,21 +127,21 @@ export async function getDAOData(avatarAddress: string, getDetails: boolean = fa
     const getMintTokenEvents = promisify(mintTokenEvents.get.bind(mintTokenEvents));
     let eventsArray = await getMintTokenEvents();
     for (let cnt = 0; cnt < eventsArray.length; cnt++) {
-      memberAddresses.push(eventsArray[cnt].args.to);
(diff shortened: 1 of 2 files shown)


src/actions/arcActions.ts
```diff
@@ -127,21 +127,21 @@ export async function getDAOData(avatarAddress: string, getDetails: boolean = fa
     const getMintTokenEvents = promisify(mintTokenEvents.get.bind(mintTokenEvents));
     let eventsArray = await getMintTokenEvents();
     for (let cnt = 0; cnt < eventsArray.length; cnt++) {
-      memberAddresses.push(eventsArray[cnt].args.to);
+      memberAddresses.push(eventsArray[cnt].args.to.toLowerCase());
```
dkent600: Just wondering:  Would it make more sense for the member addresses to be accessors (via `get`) that convert the value, when requested, to lowercase?   Just wondering.  Maybe define a type that is an address that encapsulates the actual string (and maybe provides other functionality too)?

tibetsprague: hmm. you mean create a class for members? im not quite sure how that would work easily with redux state stuff, ill think about it.

tibetsprague: yeah unfortunately there is not a clean easy way to do this with Redux.

PR #95, opened 2018-06-05
If a DAO does not have enough ETH for a reward, don’t redeem it

tibetsprague:
And show the user that it won’t be redeemed


tibetsprague:
@dev-matan-tsuberi yeah probably. I can add that

tibetsprague:
@dev-matan-tsuberi added the same functionality for staker bounties

src/actions/arcActions.ts
```diff
@@ -1064,7 +1065,7 @@ export function redeemProposal(daoAvatarAddress: string, proposal: IProposalStat
         eventWatcher.stopWatching()
       }
 
-      if (redemption.stakerBountyTokens) {
+      if (redemption.stakerBountyTokens && dao.genCount >= redemption.stakerBountyTokens) {
```
dev-matan-tsuberi: `stakerBountyTokens` should already be `0` if greater than the dao's funds.
https://github.com/daostack/alchemy/blob/ea2160a9ca8aced422feea76ee24e4e758a32581/src/actions/arcActions.ts#L395

tibetsprague: oh, ok so that check isn't needed then

tibetsprague: oh but actually we don't want that line, because then we are showing the user a staking bounty of 0, when actually it should be more.

dev-matan-tsuberi: Correct. Then we should remove that line.

tibetsprague:
@dev-matan-tsuberi ok removed it

PR #96, opened 2018-06-05
Validate create proposal form

dev-matan-tsuberi:
using [Formik](https://github.com/jaredpalmer/formik).

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
   },
   "dependencies": {
     "@daostack/arc.js": "0.0.0-alpha.58",
-    "@types/prop-types": "^15.5.3",
     "aws-sdk": "^2.212.1",
     "axios": "^0.17.1",
     "bignumber.js": "^5.0.0",
@@ -107,6 +106,7 @@
     "@types/lodash": "^4.14.74",
     "@types/node": "^8.0.26",
     "@types/object-hash": "^1.2.0",
+    "@types/prop-types": "^15.5.3",
     "@types/rc-tooltip": "^3.4.11",
     "@types/react": "^16.0.35",
     "@types/react-dom": "^16.0.3",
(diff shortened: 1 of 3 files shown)


dev-matan-tsuberi:
[Consistent form validation](https://trello.com/c/QoFi2IHg/125-consistent-form-validation)

dev-matan-tsuberi:
["Create Proposal" form, URL is not validated](https://trello.com/c/0ToKXm10/223-create-proposal-form-url-is-not-validated)

dev-matan-tsuberi:
["Create Proposal" form: Reputation reward & ETH reward are not validated to be positive numbers](https://trello.com/c/LE7oTNYZ/224-create-proposal-form-reputation-reward-eth-reward-are-not-validated-to-be-positive-numbers)

package.json
```diff
@@ -60,6 +60,7 @@
   },
   "dependencies": {
     "@daostack/arc.js": "0.0.0-alpha.58",
+    "@types/prop-types": "^15.5.3",
```
tibetsprague: should be a dev-dependency

src/components/CreateProposal/CreateProposalContainer.tsx
```diff
@@ -159,125 +117,189 @@ class CreateProposalContainer extends React.Component<IProps, IState> {
...
+              description,
+              ethReward,
+              externalTokenReward,
+              nativeTokenReward,
+              reputationReward,
+              title
+            } = values;
+            const errors: any = {};
+
+            const require = (name: string) => {
```
tibetsprague: seems really weird that we have to define require here. so we have to do this for every form we want to validate?

dev-matan-tsuberi: just for convenience really... Formik doesn't provide built in validation rules. We can use HTML5's `required` but the error message would look different than the others. 
If it gets used more than once we'll extract it to a separate module.

PR #99, opened 2018-06-08
Dont act on pending transactions

tibetsprague:
Don't show proposals in the UI until they have been confirmed on the blockchain.

### DIFF
--- a/src/actions/arcActions.ts
+++ b/src/actions/arcActions.ts
@@ -568,7 +568,23 @@ export function createProposal(daoAvatarAddress: string, title: string, descript
       const votingMachineParamsHash = await dao.controller.getSchemeParameters(votingMachineInstance.contract.address, dao.avatar.address)
       const votingMachineParams = await votingMachineInstance.contract.parameters(votingMachineParamsHash)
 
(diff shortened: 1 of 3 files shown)


src/actions/arcActions.ts
```diff
@@ -568,7 +568,23 @@ export function createProposal(daoAvatarAddress: string, title: string, descript
       const votingMachineParamsHash = await dao.controller.getSchemeParameters(votingMachineInstance.contract.address, dao.avatar.address)
       const votingMachineParams = await votingMachineInstance.contract.parameters(votingMachineParamsHash)
 
-      const submitProposalTransaction: any = await Util.performAction(
+      const descriptionHash = Arc.Utils.SHA3(description);
```
dev-matan-tsuberi: Why hash the description?

tibetsprague: because we dont have the proposalId until the transaction completes so we look up the proposal by descriptionHash onProposalCreateEvent

dev-matan-tsuberi: ok

PR #101, opened 2018-06-12
Make sure any confirmed proposals have their server data updated with the correct proposalId

tibetsprague:
Make sure any confirmed proposals have their server data updated with the correct proposalId

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "auto-start-ganache": "cross-env arcjs_network=ganache run-with-ganache --ganache-cmd ganache-cli \"npm run migrate-ganache && npm run start-ganache\""
   },
   "dependencies": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -61,7 +61,7 @@
     "auto-start-ganache": "cross-env arcjs_network=ganache run-with-ganache --ganache-cmd ganache-cli \"npm run migrate-ganache && npm run start-ganache\""
   },
   "dependencies": {
-    "@daostack/arc.js": "0.0.0-alpha.58",
+    "@daostack/arc.js": "0.0.0-alpha.59",
```
dev-matan-tsuberi: Why do you need the latest arc.js?

dkent600: To fix a bug in retrieving the contribution reward proposal description hash

dev-matan-tsuberi: ok

PR #103, opened 2018-06-14
fix 240 sec issue

dev-matan-tsuberi:
This PR silences the 240 sec issue by reaching into the underlying truffle contracts and setting `synchronization_timeout` to `0` (no timeout).

It's important to set it also when obtaining a new contract instance via `.new(..)` or `.at(..)` (but not `.deployed(..)`), though no such instances are present (that are used to fire transactions).


dev-matan-tsuberi:
[Do not send "operation failed" notification after 240 seconds passed.](https://trello.com/c/42pw11QG/251-do-not-send-operation-failed-notification-after-240-seconds-passed)

src/actions/arcActions.ts
```diff
@@ -587,7 +587,7 @@ export function createProposal(daoAvatarAddress: string, title: string, descript
       const schemeParamsHash = await dao.controller.getSchemeParameters(contributionRewardInstance.contract.address, dao.avatar.address);
       const schemeParams = await contributionRewardInstance.contract.parameters(schemeParamsHash);
       const votingMachineAddress = schemeParams[2];
-      const votingMachineInstance = await Arc.GenesisProtocolFactory.at(votingMachineAddress);
```
tibetsprague: we can remove the votingMachineAddress const no?

src/index.tsx
```diff
@@ -10,6 +10,17 @@ import "./assets/styles/global.scss";
 async function renderApp() {
   try {
     await Arc.InitializeArcJs();
+    // Silence 240 sec error
+    Arc.ContractWrappers.AbsoluteVote.contract.constructor.synchronization_timeout = 0;
```
tibetsprague: I hope this is fixed by the work @dkent600 is doing to change how arc.js handles transactions

src/actions/arcActions.ts
dev-matan-tsuberi: done https://github.com/daostack/alchemy/pull/103/commits/e6e5ebf9f63cfa24e4c88bb950ee25afcc9f9e51

src/index.tsx
dkent600: @tibetsprague Without changing the timeout, truffle will still timeout waiting for mined transactions.

PR #104, opened 2018-06-15
Alchemy aurum

brodeur:


src/components/Home/Home.scss
```diff
@@ -1,3 +1,122 @@
...
+  display: table;
+  clear: both;
+}
+
 .homeWrapper {
   width: 100%;
+}
+
+.topTriangle {
+	position: absolute;
```
tibetsprague: seems like maybe you have tabs turned on in Sublime? please switch to two spaces

PR #127, opened 2018-07-18
revert to using env var instead of arc.js for the cache script

dev-matan-tsuberi:


dkent600:
Just curious why this change is needed.

dev-matan-tsuberi:
@dkent600 The scripts needs to take the `arcjs_network` from the heroku configuration. which is `kovan` for `staging` and `live` for production.

PR #137, opened 2018-07-23
Redeem button text

dev-matan-tsuberi:


dev-matan-tsuberi:
[User can manually execute a proposal or redeem any proposal for proposal beneficiary if it passed](https://trello.com/c/XfpcLhfa/313-user-can-manually-execute-a-proposal-or-redeem-any-proposal-for-proposal-beneficiary-if-it-passed)

dev-matan-tsuberi:
[Proposal execution error](https://trello.com/c/UXywL3Fu/377-proposal-execution-error)

scripts/mintGen.js
```diff
@@ -8,7 +8,9 @@ async function main() {
   const DAOToken = await Arc.Utils.requireContract('DAOToken');
   const genToken = await DAOToken.at(genTokenAddress)
 
-  await Promise.all(web3.eth.accounts.map(async account => {
+  const daoAddress = '0xf7b7be05d6c115184f78226f905b643dd577fa6b';
```
tibetsprague: this is the ganacheDb DAO right? what if they don't use ganacheDb? I guess we just encourage that

src/actions/arcActions.ts
```diff
@@ -446,25 +446,24 @@ async function getRedemptions(avatarAddress: string, votingMachineInstance: Arc.
...
-
-      redemptions.beneficiaryEth = numberOfPeriods * proposal.ethReward;
-      redemptions.beneficiaryNativeToken = numberOfPeriods * proposal.nativeTokenReward;
-      redemptions.beneficiaryReputation = numberOfPeriods * proposal.reputationChange;
-    } else {
-      redemptions.beneficiaryEth = (await proposalInstance.contract.getPeriodsToPay(proposalId, avatarAddress, ContributionRewardType.Eth)) * proposal.ethReward;
-      redemptions.beneficiaryNativeToken = (await proposalInstance.contract.getPeriodsToPay(proposalId, avatarAddress, ContributionRewardType.NativeToken)) * proposal.nativeTokenReward;
-      redemptions.beneficiaryReputation = (await proposalInstance.contract.getPeriodsToPay(proposalId, avatarAddress, ContributionRewardType.Reputation)) * proposal.reputationChange;
-    }
+  if (proposal.state == ProposalStates.Boosted && proposal.winningVote === VoteOptions.Yes) {
```
tibetsprague: I think this is going to be confusing. how do we track if the beneficiary rewards are for the current user or for someone else? actually this will be easier with my new architecture coming right up

src/components/Proposal/ProposalContainer.tsx
```diff
@@ -108,18 +111,26 @@ class ProposalContainer extends React.Component<IProps, IState> {
       isVotingYes
     } = this.props;
 
-    const redeemable = currentAccountRedemptions && (
+    const beneficiaryHasRewards = currentAccountRedemptions && (
       currentAccountRedemptions.beneficiaryReputation ||
       currentAccountRedemptions.beneficiaryNativeToken ||
+      (currentAccountRedemptions.beneficiaryEth && dao.ethCount >= currentAccountRedemptions.beneficiaryEth)
```
tibetsprague: I think we always want to show the rewards and just indicate if there isn't enough ETH to actually carry out the reward

src/reducers/arcReducer.ts
```diff
@@ -450,12 +453,9 @@ const arcReducer = (state = initialState, action: any) => {
...
               members: {
                 [accountAddress]: {
-                  $merge : payload.beneficiary,
+                  $merge: payload.beneficiary,
                   // remove pending redemptions from this account
-                  redemptions :
-                    payload.redemptions ?
-                      { [proposalId]: {$merge: payload.redemptions}} :
-                      { $unset: [proposalId] },
+                  redemptions: { $unset: [proposalId] },
```
tibetsprague: couldn't there still be redemptions left, if the DAO doesnt have enough funds to cover them all? I think we still want it the way it was

src/components/Proposal/ProposalContainer.tsx
dev-matan-tsuberi: We do, in [this](https://github.com/daostack/alchemy/pull/137/files/dab40ee32d0b023a076f8e76a03161c4b10663b5#diff-c02babe5ab2a47081fe5b019ee9f62f2R432) line. This boolean only indicates if there are rewards that the beneficiary can actually redeem.

If there are rewards that can't be redeemed, the button and tooltip will show up but will be disabled.

src/reducers/arcReducer.ts
dev-matan-tsuberi: I tried to update redemptions on the redeem event via `getRedemptions`, but it seems to return old values.

src/components/Proposal/ProposalContainer.tsx
tibetsprague: right

src/reducers/arcReducer.ts
tibetsprague: hmm, so we will still end up with the weird row disappearing then reappearing... we may have to manually figure out which ones are being redeemed

dev-matan-tsuberi: i'll try to fix that

PR #138, opened 2018-07-23
Persist only pending operations

dev-matan-tsuberi:


dev-matan-tsuberi:
[There's a ton of operations left hanging around in the state](https://trello.com/c/QV7dFMgC/379-theres-a-ton-of-operations-left-hanging-around-in-the-state)

PR #144, opened 2018-07-25
Boosted -> (Boosted || QuietEnding)

dev-matan-tsuberi:


dev-matan-tsuberi:
[Proposal in QuietEndingPeriod state don't appear in Alchemy at all](https://trello.com/c/Ri31uW1f/370-proposal-in-quietendingperiod-state-dont-appear-in-alchemy-at-all)

PR #145, opened 2018-07-25
round second decminal up

dev-matan-tsuberi:


tibetsprague:
hmm, this doesnt solve the issue does it? I thought the point was to round up the second decimal of what the user entered. though I guess either way could solve it, because if we show them that they need to enter the rounded up amount then I guess they would. ok, works for me

dev-matan-tsuberi:
The problem is that we show `X MORE GEN REQUIRED` and the user enters X and expects the proposal to be executed and it isn't.

PR #147, opened 2018-07-26
fixes

dev-matan-tsuberi:


src/reducers/arcReducer.ts
```diff
@@ -466,11 +466,27 @@ const arcReducer = (state = initialState, action: any) => {
       switch (sequence) {
         case AsyncActionSequence.Pending:
           return update(state, {
-            redemptions: { [redemptionsKey] : { transactionState: { $set: TransactionStates.Unconfirmed }} }
+            redemptions: {
+              [redemptionsKey]: {
+                $apply: (redemptions: IRedemptionState) => ({
+                  ...redemptions,
```
tibetsprague: why do we need to update the redemption here?

src/reducers/arcReducer.ts
```diff
@@ -487,17 +503,40 @@ const arcReducer = (state = initialState, action: any) => {
...
-                  redemptions: (arr: string[]) => arr.filter((item) => item != redemptionsKey)
+                  redemptions: (arr: string[]) => arr.filter((item) => item != redemptionsKey),
                 }
               },
               redemptions: { $unset: [redemptionsKey] }
             });
           }
 
-          // Also update the beneficiary account and the dao
+          if (accountRedemptions) {
```
tibetsprague: we shouldn't merge this in if its broken, lets leave it on a separate branch so we can both look at it more

dev-matan-tsuberi: It's currently just deleting the redemption once it's complete because I'm not passing `redemptions` and `accountRedemptions`. So it's not broken.

src/reducers/arcReducer.ts
dev-matan-tsuberi: I was getting `cannot read transactionState of undefined` because immutability-helper tries to read it from an undefined redemption. 
This simply does the same thing but avoids this error.

src/reducers/arcReducer.ts
dev-matan-tsuberi: Feel free to uncomment the two lines at `onRedeemEvent` if you find a way to update it correctly.

src/reducers/arcReducer.ts
tibetsprague: well yeah, but that error should not be happening right? the redemption should be there. I guess this is a decent failsafe, but I worry that it could hide another error then. can you think of a valid reason why the redemption wouldn't be there?

dev-matan-tsuberi: I think only `transactionState` is not defined on the redemption, need to check. Though it makes sense that it's undefined until the first redeem. Or you can have an `idle` state if you want.

tibetsprague:
I dont think we can do this will or right without the RedeemerRedeem event telling us who the beneficiary is/was

PR #149, opened 2018-07-27
Added overtime language and helper

brodeur:


src/components/Proposal/ProposalContainer.tsx
```diff
@@ -357,9 +357,52 @@ class ProposalContainer extends React.Component<IProps, IState> {
             }
             <h3>
               <span>
+
+                    <strong>
```
tibetsprague: i dont understand what this is doing here. looks like a duplicate of below?

PR #155, opened 2018-07-31
use GP & CR events to properly update redemptions state

dev-matan-tsuberi:


dev-matan-tsuberi:
[Redemption rows disappear when there are blocked rewards (Insufficient GEN/ETH)](https://trello.com/c/twhAIkqB/384-redemption-rows-disappear-when-there-are-blocked-rewards-insufficient-gen-eth)

dev-matan-tsuberi:
[Track GenesisProtocol and ContributionReward Redeem events to update redemptions](https://trello.com/c/IpLJmPiM/440-track-genesisprotocol-and-contributionreward-redeem-events-to-update-redemptions)

src/actions/arcActions.ts
```diff
@@ -1173,6 +1160,29 @@ export function onRedeemEvent(proposalId: string) {
   };
 }
 
+export enum RewardType {
```
tibetsprague: this should go in arcReducer.ts (actually i desperately want to create a types or models file)

src/layouts/AppContainer.tsx
```diff
@@ -82,11 +82,7 @@ const mapDispatchToProps = {
 type IProps = IStateProps & IDispatchProps;
 
 class AppContainer extends React.Component<IProps, null> {
-  public proposalEventWatcher: Arc.EventFetcher<Arc.NewContributionProposalEventResult>;
-  public stakeEventWatcher: Arc.EventFetcher<Arc.StakeEventResult>;
-  public voteEventWatcher: Arc.EventFetcher<Arc.VoteProposalEventResult>;
-  public redeemEventWatcher: Arc.EventFetcher<Arc.RedeemerRedeemEventResult>;
-  public executeProposalEventWatcher: Arc.EntityFetcher<Arc.ExecutedGenesisProposal, Arc.ExecuteProposalEventResult>;
+  public watchers: Array<Arc.EventFetcher<any> | Arc.EntityFetcher<any, any>> = []
```
tibetsprague: thanks

src/bg_cache_worker.ts
```diff
@@ -176,15 +176,40 @@ async function updateCache() {
...
-    const redeemEventWatcher = redeemerInstance.RedeemerRedeem({ }, { fromBlock: lastCachedBlock, toBlock: latestBlock });
-    const getRedeemEvents = promisify(redeemEventWatcher.get.bind(redeemEventWatcher, -1));
-    const redeemEvents: Array<Arc.DecodedLogEntryEvent<Arc.RedeemerRedeemEventResult>> = await getRedeemEvents();
-    for (let index = 0; index < redeemEvents.length; index++) {
-      const event = redeemEvents[index];
-      console.log("Proposal redeemed through Redeemer", event);
-      await store.dispatch(arcActions.onRedeemEvent(event.args._proposalId));
-    }
+    const redeemEth = await contributionRewardInstance.RedeemEther({}, {fromBlock: lastCachedBlock, toBlock: latestBlock}).get(undefined, -1);
+    redeemEth.forEach(({args: {_proposalId, _avatar, _beneficiary}}) =>
```
tibetsprague: i don't think we can use foreach because im pretty sure we want to await each dispatch and carry out each event one by one in the cache script. so need to use a regular for loop

src/actions/arcActions.ts
```diff
@@ -1100,6 +1090,8 @@ export function redeemProposal(daoAvatarAddress: string, proposal: IProposalStat
     try {
       const redeemerInstance = await Arc.RedeemerFactory.deployed();
       const redeemTx = await redeemerInstance.redeem({ avatarAddress: daoAvatarAddress, beneficiaryAddress: accountAddress, proposalId: proposal.proposalId });
+      await redeemTx.getTxConfirmed();
```
tibetsprague: hmm, this does not seem like a good idea to me. don't we need to wait for the RedeemerRedeem event to fire off the onRedeemEvent? this wont work if they refresh the page right?

dev-matan-tsuberi: The `RedeemerRedeem` event was removed, no?

src/bg_cache_worker.ts
dev-matan-tsuberi: right, also I noticed that there are some places we can use `Promise.all` instead of a `for` loop with an `await` to make it parallel (e.g. in `getDAOs`). I think it's also the case here.

tibetsprague: yeah thats true

src/actions/arcActions.ts
tibetsprague: oh boy, yeah forgot to mention that we actually could still use it for this case. what do you suggest we do?

dev-matan-tsuberi: Well, we can use the `operations` module that gets persisted through refreshes.

src/actions/arcActions.ts
dev-matan-tsuberi: Sure. And we'll make the refactoring a seperate PR.

dev-matan-tsuberi: https://github.com/daostack/alchemy/pull/155/commits/14a02fc8da2301b4506f6ffaca6419dd700d0cbc

src/bg_cache_worker.ts
dev-matan-tsuberi: https://github.com/daostack/alchemy/pull/155/commits/859188fae85d97f760f2cc184f8a987fb1f4eb97

src/actions/arcActions.ts
dev-matan-tsuberi: https://github.com/daostack/alchemy/pull/155/commits/50a3d168b968097e2b386c2b0639fd4956c64f02

dev-matan-tsuberi:
One question: Should we enable the redeem button when a redeem tx is in progress (to enable users to increase gas price)?

PR #158, opened 2018-08-03
fix quiet ending

dev-matan-tsuberi:


dev-matan-tsuberi:
[Proposal in QuietEndingPeriod state don't appear in Alchemy at all](https://trello.com/c/Ri31uW1f/370-proposal-in-quietendingperiod-state-dont-appear-in-alchemy-at-all)

src/reducers/arcReducer.ts
```diff
@@ -229,9 +231,10 @@ export const initialState: IArcState = {
 };
 
 export const closingTime = (proposal: IProposalState) => {
-  const { state, boostedTime, submittedTime, preBoostedVotePeriodLimit, boostedVotePeriodLimit, executionTime } = proposal;
+  const { state, boostedTime, submittedTime, preBoostedVotePeriodLimit, boostedVotePeriodLimit, quietPeriod, executionTime } = proposal;
+
   const start = boostedTime ? boostedTime : submittedTime;
-  const duration = boostedTime ? boostedVotePeriodLimit : preBoostedVotePeriodLimit;
+  const duration = boostedTime ? (state === ProposalStates.QuietEndingPeriod ? quietPeriod : boostedVotePeriodLimit) : preBoostedVotePeriodLimit;
```
tibetsprague: yeah i don't think this is right, i think my fix in a9136f3e9b64eba68d07b7440e93a0c0e336e57c handles it, because Arc actually updates the boostedTime and the boostedVotePeriodLimit when it hits the quiet ending period

src/actions/arcActions.ts
```diff
@@ -327,6 +327,7 @@ async function getProposalDetails(daoInstance: Arc.DAO, votingMachineInstance: A
     beneficiaryAddress: contributionProposal.beneficiaryAddress.toLowerCase(),
     boostedTime,
     boostedVotePeriodLimit,
+    quietPeriod: votingMachineParams.quietEndingPeriod,
```
tibetsprague: we dont need this

src/components/Proposal/ProposalContainer.tsx
```diff
@@ -363,7 +363,7 @@ class ProposalContainer extends React.Component<IProps, IState> {
               <span>
                 { proposal.state == ProposalStates.QuietEndingPeriod ?
                     <strong>
-                      <img src="/assets/images/Icon/Overtime.svg"/> OVERTIME: CLOSES IN {closingTime(proposal).fromNow().toUpperCase()}
+                      <img src="/assets/images/Icon/Overtime.svg"/> OVERTIME: CLOSES {closingTime(proposal).fromNow().toUpperCase()}
```
tibetsprague: well take this though :)

src/actions/arcActions.ts
```diff
@@ -283,10 +283,10 @@ async function getProposalDetails(daoInstance: Arc.DAO, votingMachineInstance: A
   const avatarAddress = daoInstance.avatar.address;
 
   const votingMachineParamsHash = await daoInstance.controller.getSchemeParameters(votingMachineInstance.contract.address, avatarAddress);
-  const votingMachineParams = await votingMachineInstance.contract.parameters(votingMachineParamsHash);
+  const votingMachineParams = await votingMachineInstance.getParameters(votingMachineParamsHash);
 
   const proposalDetails = await votingMachineInstance.getProposal(proposalId);
-  const preBoostedVotePeriodLimit = Number(votingMachineParams[1]);
+  const preBoostedVotePeriodLimit = Number(votingMachineParams.preBoostedVotePeriodLimit);
```
tibetsprague: and this is nice

src/reducers/arcReducer.ts
dev-matan-tsuberi: `boostedTime` time is updated to be the time when the decision was flipped. `boostedVotePeriodLimit` is a DAO param though and does not change. Once the the proposal is in quiet mode the calculation should be `boostedTime + quietPeriod`.

src/actions/arcActions.ts
dev-matan-tsuberi: How would you know the length of the quiet ending then?

src/reducers/arcReducer.ts
tibetsprague: proposal.currentBoostedVotePeriodLimit is the one we care about, and Arc does update it

src/actions/arcActions.ts
tibetsprague: because currentBoostedVotePeriodLimit gets set to it by Arc and we update proposal.boostedVotePeriodLimit on our end to match it (we should really match all the names)

src/reducers/arcReducer.ts
dev-matan-tsuberi: correct, was confused by the naming because `boostedVotePeriodLimit` is a DAO param in Arc.

tibetsprague: yeah, that would be a good thing to change

PR #163, opened 2018-08-09
disallow empty tooltip

dev-matan-tsuberi:


dev-matan-tsuberi:
[Main net: After staking tx is confirmed the loading animation is still covering the staking buttons](https://trello.com/c/v3GKafaE/482-main-net-after-staking-tx-is-confirmed-the-loading-animation-is-still-covering-the-staking-buttons)

src/components/Proposal/PredictionBox.tsx
```diff
@@ -188,6 +188,20 @@ export default class PredictionBox extends React.Component<IProps, IState> {
       [css.hidden]: stakeAmount <= currentAccountGens
     });
 
+    const passButton = (
+      <button className={isPredictingPass ? css.pendingPrediction : undefined} onClick={disableStakePass ? "" : this.showModal.bind(this, 1)}>
```
tibetsprague: seems odd to use undefined here instead of "", but if it works it works

PR #164, opened 2018-08-09
use `NewProposal` event to get proposer

dev-matan-tsuberi:


dev-matan-tsuberi:
[Use `NewContributionProposal` event to get the `proposer` instead of using the proposal's params](https://trello.com/c/2tj23LdU/458-use-newcontributionproposal-event-to-get-the-proposer-instead-of-using-the-proposals-params)

src/actions/arcActions.ts
```diff
@@ -338,7 +340,7 @@ async function getProposalDetails(daoInstance: Arc.DAO, votingMachineInstance: A
     reputationChange: Util.fromWei(contributionProposal.reputationChange),
     periodLength: Number(contributionProposal.periodLength),
     preBoostedVotePeriodLimit,
-    proposer: proposalDetails.proposer,
```
tibetsprague: onProposalExecuted also updates the proposer, so this needs to happen there too

dev-matan-tsuberi: https://github.com/daostack/alchemy/pull/164/commits/23795c8413e09bfc63ca4f9a36355bc069811917

src/actions/arcActions.ts
```diff
@@ -820,8 +820,8 @@ export function onProposalExecuted(avatarAddress: string, proposalId: string, ex
       proposal.reputationWhenExecuted = reputationWhenExecuted;
       proposal.winningVote = decision;
 
-      const gpProposalDetails = await votingMachineInstance.getProposal(proposalId);
-      proposal.proposer = gpProposalDetails.proposer; // Have to do this because redeem sets proposer to 0, to prevent future redemptions for proposer
+      // Have to do this because redeem sets proposer to 0, to prevent future redemptions for proposer
+      proposal.proposer = (await votingMachineInstance.NewProposal({_proposalId: proposalId}, {fromBlock: 0, toBlock: 'latest'}).get(undefined, -1))[0].args._proposer;
```
tibetsprague: hmm, wait, actually this is tricky. We actually want to know that the proposer is 0 here, so we don't show redemptions for the proposer. Or wait, i guess that was fixed on the Arc side? Did you test this?

dev-matan-tsuberi: alchemy needs the `proposal.proposer` to always be non-zero which is guarenteed to be the case when fetching from the event, `getRedemptions(..)` gets the proposer from the blockchain so it'll be zero after a redeem. I don't understand the issue.

tibetsprague: yeah i hear that, but then seems like you can remove a few places where we check for whether proposer is 0x000...

PR #167, opened 2018-08-09
Make sure that the proposal executed event doesn’t overwrite data fro…

tibetsprague:
…m votes

This fixes the bug where a proposal passed by decision but shows up as passed by timeout, and doesnt show the votes correctly.
This is only an issue in the real time updates which happen asynchronously. In the cache script everything happens one event at a time so it works fine.
The fix is to only update the proposal values that change in each event, though perhaps the better future fix will be to handle each event one at a time in order.

https://trello.com/c/sF11k1Nj


jellegerbrandy:
Closing this as it seems obsolete. Plz reopen if still relevant

PR #168, opened 2018-08-13
Show proposal creators/proposers in the history

tibetsprague:
Whoa look, and we can clean up duplicate code in ProposalContainer!

https://trello.com/c/6YDMezEq


dev-matan-tsuberi:
Aren't they already shown?

PR #169, opened 2018-08-13
Don’t show execute button for already executed proposals

tibetsprague:
https://trello.com/c/XwojFhw7

### DIFF
--- a/src/components/Proposal/ProposalContainer.tsx
+++ b/src/components/Proposal/ProposalContainer.tsx
@@ -189,26 +189,6 @@ class ProposalContainer extends React.Component<IProps, IState> {
         [css.disabled]: !redeemable && !executable
       });
 
-      const redeemButton = (
-        <button
-          style={{whiteSpace: 'nowrap'}}
-          disabled={!redeemable && !executable}
-          className={redeemRewards}
-          onClick={this.handleClickRedeem.bind(this)}
-        >
-          {
-            isRedeemPending ?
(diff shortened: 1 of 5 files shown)


src/components/Proposal/ProposalContainer.tsx
```diff
@@ -271,6 +251,28 @@ class ProposalContainer extends React.Component<IProps, IState> {
           {isRedeemPending ? <strong><i>Warning: Redeeming for this proposal is already in progress</i></strong> : ''}
         </div>;
 
+      const redeemButton = (redeemable || executable ?
```
dev-matan-tsuberi: should be `currentRedemptions || beneficiaryRedemptions || executable` because there may be nonredeemable rewards and we still want to show the button.

tibetsprague: i think that was what was causing the issue though with showing an execute button when we shouldnt have. ok ill test some more

tibetsprague: @dev-matan-tsuberi i think it should be `currentRedemptions || beneficiaryHasRewards || executable` because we don't want to show the button if the beneficiary has GP rewards, only CR rewards (beneficiary rewards)

dev-matan-tsuberi: :+1:

tibetsprague:
@dev-matan-tsuberi ok updated, check it out

PR #171, opened 2018-08-14
Use new arc to get correct rewards to redeem at all times

tibetsprague:


src/actions/arcActions.ts
```diff
@@ -339,20 +332,18 @@ async function getProposalDetails(daoInstance: Arc.DAO, votingMachineInstance: A
     redemptions: [],
     reputationChange: Util.fromWei(contributionProposal.reputationChange),
     periodLength: Number(contributionProposal.periodLength),
-    preBoostedVotePeriodLimit,
+    preBoostedVotePeriodLimit: Number(votingMachineParams[1]),
```
dev-matan-tsuberi: can get from `proposalDetails`

tibetsprague: no, this is a param on the voting machine, not on the proposal

PR #176, opened 2018-08-20
Dao script

dev-matan-tsuberi:


dev-matan-tsuberi:
[Populate the DAO script](https://trello.com/c/SJ98eY7N/504-populate-the-dao-script)

package.json
```diff
@@ -144,12 +148,14 @@
     "sass-resources-loader": "^1.3.0",
     "source-map-loader": "^0.2.1",
     "ts-jest": "^22.4.1",
+    "ts-node": "^5.0.1",
```
tibetsprague: why is this now a devDependency? its already a regular dependency

scripts/example.json
```diff
@@ -0,0 +1,45 @@
+{
```
tibetsprague: i think the name should be something more descriptive than example.json. exampleDaoData.json? also maybe this should go in the test directory?

scripts/populateDAO.ts
```diff
@@ -0,0 +1,335 @@
+import * as yargs from 'yargs';
```
tibetsprague: lets alphabetize our includes

scripts/populateDAO.ts
```diff
@@ -0,0 +1,335 @@
...
+import chalk from 'chalk';
+import * as moment from 'moment';
+import * as fs from 'fs';
+import * as Arc from '@daostack/arc.js';
+import { VoteOptions, DAO } from '@daostack/arc.js';
+import Util from '../src/lib/util';
+import axios from "axios";
+import { ContributionRewardWrapper, GenesisProtocolWrapper } from '@daostack/arc.js';
```
tibetsprague: also put the arc.js stuff together

scripts/populateDAO.ts
```diff
@@ -0,0 +1,335 @@
+import * as yargs from 'yargs';
+import chalk from 'chalk';
+import * as moment from 'moment';
+import * as fs from 'fs';
+import * as Arc from '@daostack/arc.js';
+import { VoteOptions, DAO } from '@daostack/arc.js';
+import Util from '../src/lib/util';
```
tibetsprague: and our code after third party vendors

scripts/populateDAO.ts
```diff
@@ -0,0 +1,335 @@
...
+const Web3 = require("web3");
+import promisify = require("es6-promisify");
+// tslint:disable-next-line:no-var-requires
+const NonceTrackerSubprovider = require("web3-provider-engine/subproviders/nonce-tracker");
+
+// tslint:disable-next-line:no-var-requires
+require('dotenv').config();
+
+enum Level {
+    Success = 'success',
```
tibetsprague: looks like you may have used tabs in this file? lets stick to two spaces everywhere

scripts/populateDAO.ts
```diff
@@ -0,0 +1,335 @@
...
+interface Opts extends CreateDAOOpts {
+    steps: Step[];
+    api: string;
+    network: string;
+    logfile: string;
+    mnemonic: string;
+}
+
+async function main(options: Opts) {
+    const mapping: {[id: number]: string} = {};
```
tibetsprague: mapping what? better name please

scripts/populateDAO.ts
```diff
@@ -0,0 +1,335 @@
...
+                alias: 'l',
+                describe: 'path to a log file to save results',
+                type: 'string'
+            }
+        )
+        .option(
+            'api', {
+                alias: 'a',
+                describe: 'api root url',
+                default: 'https://daostack-alchemy.herokuapp.com',
```
tibetsprague: hmm, i think we should not default to production server, lets default to daostack-alchemy-server-stage

tibetsprague: though really we just need to always use the right server for the network used, im going to pull that ticket in

package.json
dev-matan-tsuberi: does it need to be a regular dependency? its should not be bundled with the app.

tibetsprague: we use it to run scripts on Heroku

dev-matan-tsuberi: doesn't it `npm install` it and runs the command inside the dir?

PR #177, opened 2018-08-20
fix "beneficiary will receive..." bug

dev-matan-tsuberi:


dev-matan-tsuberi:
["The beneficiary of this proposal will receive" text appears in redeeming preview on failed proposals.](https://trello.com/c/11YcqTMI/524-the-beneficiary-of-this-proposal-will-receive-text-appears-in-redeeming-preview-on-failed-proposals)

dev-matan-tsuberi:
Actually I think we should revert your fix. This error appears because we use `beneficiaryRedemptions` instead of `beneficiaryHasRewards`. When the proposal fails the beneficiary doesn't have contribution rewards but may have other rewards (e.g. voting on the proposal). `beneficiaryHasRewards` will only show contribution rewards, while the rest of the view will show the others.

dev-matan-tsuberi:
So I think the appropriate condition is: https://github.com/daostack/alchemy/pull/177/commits/dd5fb1455a0142349b90a131dc9fcfe2cb00582b

PR #197, opened 2018-11-02
Performance

jellegerbrandy:
This is work in progress, don't merge yet :-)


webpack.dev.config.js
```diff
@@ -8,7 +8,7 @@ const baseConfig = require('./webpack.base.config.js');
 require('dotenv').config();
 
 module.exports = merge(baseConfig, {
-  devtool: 'cheap-module-eval-source-map',
+  devtool: 'eval-source-map',
```
jellegerbrandy: this will give us more detailed source maps to use in dev tools...

package.json
```diff
@@ -87,9 +87,9 @@
     "normalizr": "^3.2.4",
     "object-hash": "^1.3.0",
     "rc-tooltip": "^3.7.0",
-    "react": "^16.2.0",
+    "react": "^16.6.0",
```
jellegerbrandy: upgrade react, bc they claim to have more timer tools in the new version..

package.json
```diff
@@ -41,25 +41,26 @@
...
-    "migrate-ganache": "npm explore @daostack/arc.js -- npm start migrateContracts && npm explore @daostack/arc.js -- npm start createGenesisDao && node ./scripts/mintGen.js",
+    "heroku-postbuild": "npm run build && tsc",
     "lint": "tslint --project .",
-    "test": "jest --coverage",
-    "test:watch": "jest --watch",
+    "migrate-ganache": "npm explore @daostack/arc.js -- npm start migrateContracts && npm explore @daostack/arc.js -- npm start createGenesisDao && node ./scripts/mintGen.js",
+    "populate-dao": "ts-node scripts/populateDAO.ts",
     "start": "webpack-dev-server --config webpack.dev.config.js",
-    "start-staging": "cross-env S3_BUCKET=daostack-alchemy-staging webpack-dev-server --config webpack.dev.config.js",
+    "start-dev-profiling": "cross-env NODE_ENV=dev-profiling API_URL=https://daostack-alchemy.herokuapp.com webpack-dev-server --config webpack.dev.config.js",
```
jellegerbrandy: new command to use to start profiling server

src/reducers/arcReducer.ts
```diff
@@ -298,7 +298,7 @@ const arcReducer = (state = initialState, action: any) => {
 
   switch (action.type) {
     case ActionTypes.ARC_LOAD_CACHED_STATE_FULFILLED: {
-      return payload;
+      return update(state, { $merge: payload});
```
jellegerbrandy: Not sure about this change.  The reason is that the payload (i.e. the S3 initial state) does not have an `accounts` key, and this leads to an error in https://github.com/daostack/alchemy/blob/dev/src/layouts/HeaderContainer.tsx#L42

package.json
```diff
@@ -41,25 +41,26 @@
...
-    "migrate-ganache": "npm explore @daostack/arc.js -- npm start migrateContracts && npm explore @daostack/arc.js -- npm start createGenesisDao && node ./scripts/mintGen.js",
+    "heroku-postbuild": "npm run build && tsc",
     "lint": "tslint --project .",
-    "test": "jest --coverage",
-    "test:watch": "jest --watch",
+    "migrate-ganache": "npm explore @daostack/arc.js -- npm start migrateContracts && npm explore @daostack/arc.js -- npm start createGenesisDao && node ./scripts/mintGen.js",
+    "populate-dao": "ts-node scripts/populateDAO.ts",
     "start": "webpack-dev-server --config webpack.dev.config.js",
-    "start-staging": "cross-env S3_BUCKET=daostack-alchemy-staging webpack-dev-server --config webpack.dev.config.js",
+    "start-dev-profiling": "cross-env NODE_ENV=dev-profiling API_URL=https://daostack-alchemy.herokuapp.com S3_BUCKET=daostack-alchemy-staging webpack-dev-server --config webpack.dev.config.js",
```
jellegerbrandy: this is the new start-dev-profiling startup command. Note the S3_BUCKET set to the "-staging" file, which is also where alchemy.daostack.io gets its initial data from..

dkent600:
@jellegerbrandy Sorry I rather prematurely approved this thinking it was a single trivla change, before noticing there are actually a lot of changes that I have not sufficient knowledge to approve.

src/components/Home/HomeContainer.tsx
```diff
@@ -45,8 +45,8 @@ class HomeContainer extends React.Component<IProps, null> {
           <h1>Alchemy</h1>
           <h2>Budgeting and resource allocation for decentralized organizations</h2>
           <div className={css.topCta}>
-            { process.env.NODE_ENV == 'production'
-                ? <a href='https://alchemy.daostack.io/dao/0xa3f5411cfc9eee0dd108bf0d07433b6dd99037f1'>View Alchemy</a>
+            { (process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'dev-profiling')
```
tibetsprague: why do you want the View Alchemy link to point to production on dev-profiling?

src/components/ViewDao/ViewDaoContainer.tsx
```diff
@@ -229,9 +244,15 @@ class ViewDaoContainer extends React.Component<IProps, IState> {
 
       updateDAOLastBlock(daoAvatarAddress, await Util.getLatestBlock());
       this.setState({ readyToShow: true });
+    }
+
+    updateState()
```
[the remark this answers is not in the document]
tibetsprague: waiting 10 seconds was a choice, a tradeoff where we chose to give the user consistency on first load, rather than speed

src/layouts/HeaderContainer.tsx
```diff
@@ -221,7 +221,7 @@ class HeaderContainer extends React.Component<IProps, null> {
               <div className={css.backgroundBlock}></div>
               <ul>
                 <li><Link to='/'>Home</Link></li>
-                { process.env.NODE_ENV == 'production'
+                { (process.env.NODE_ENV === 'production' || process.env.NODE_ENV === 'dev-profiling')
```
tibetsprague: again confused by this

src/reducers/arcReducer.ts
```diff
@@ -298,7 +298,7 @@ const arcReducer = (state = initialState, action: any) => {
 
   switch (action.type) {
     case ActionTypes.ARC_LOAD_CACHED_STATE_FULFILLED: {
-      return payload;
+      return payload
```
tibetsprague: ?

src/components/Home/HomeContainer.tsx
jellegerbrandy: I don't. See the line below :)

src/components/ViewDao/ViewDaoContainer.tsx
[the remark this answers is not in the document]
jellegerbrandy: I wrote to you on discord to discuss this further...

src/layouts/HeaderContainer.tsx
jellegerbrandy: as above, I made it a relative link...

src/reducers/arcReducer.ts
jellegerbrandy: I changed this line a few times bc of the confusion about the location of the initialstate file (on staging or production...) and forgot to restore the `; `. 

Shoudl the linter have caught this?

src/reducers/arcReducer.ts
tibetsprague: hmm, it should (and does) have an accounts key, when/where are you seeing this issue?

src/components/ViewDao/ViewDaoContainer.tsx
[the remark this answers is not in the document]
tibetsprague: responded, but i don't think your change should make a meaningful difference in load time

src/reducers/arcReducer.ts
tibetsprague: seems like it

src/components/Home/HomeContainer.tsx
tibetsprague: even relative though doesnt make sense because the DAO address is fixed and network dependent, so that still only works on production. what we actually want is for every environment except production (but including staging, which is the main change that needs doing) to point to /daos

src/reducers/arcReducer.ts
jellegerbrandy: yeah, this was becuase of my confusion about the location of the initial state, cf #201, I undid this change in a later commit

src/components/ViewDao/ViewDaoContainer.tsx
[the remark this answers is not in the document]
jellegerbrandy: resolvign this conf, as we cleared in up in pvt :-)

src/components/Home/HomeContainer.tsx
jellegerbrandy: the reasoning is to make de `dev-profiling` behavior as similar to production as possible

[that said, I think it is better to have separate flags/env variables for these settings, i.e. isntead of having this depend on NODE_ENV, we can add a new env variable called MAIN_DAO_ADDRESS or something]

jellegerbrandy: ok, I see what you mean now.. Fixed this.

jellegerbrandy:
most changes are in the client-integration branch, closing this

PR #204, opened 2018-11-09
Identity MVP

tibetsprague:
Still missing social verification


src/components/Account/AccountProfileContainer.tsx
```diff
@@ -0,0 +1,255 @@
...
+
+  public async componentWillMount() {
+    const { accountAddress, dao } = this.props;
+    const web3 = await Arc.Utils.getWeb3();
+    const getBalance = promisify(web3.eth.getBalance);
+    const ethBalance = await getBalance(accountAddress);
+
+    let votingMachineInstance: Arc.GenesisProtocolWrapper;
+    if (dao) {
+      const contributionRewardInstance = await Arc.ContributionRewardFactory.deployed();
```
jellegerbrandy: This is a lot of logic just to get the GEN balance. arc.js does calls to the blockchain for each of these `deployed` and `at` calls. Wouldn't it make more sense to have all these addressed in the redux state, so they are easily re-used?

src/components/Account/AccountProfileContainer.tsx
```diff
@@ -0,0 +1,255 @@
...
+    const msg = ethUtil.bufferToHex(Buffer.from(text, 'utf8'));
+    const fromAddress = this.props.accountAddress;
+
+    let signature = localStorage.getItem("signature-" + fromAddress);
+    if (!signature) {
+      const method = 'personal_sign';
+      const sendAsync = promisify(web3.currentProvider.sendAsync);
+      const params = [msg, fromAddress];
+      const result = await sendAsync({ method, params, fromAddress });
+      signature = result.result;
```
jellegerbrandy: wait what? You are singing the message on the Ethereum node? Why would you do that?

src/components/Account/AccountProfileContainer.tsx
```diff
@@ -0,0 +1,255 @@
...
+    const web3 = await Arc.Utils.getWeb3();
+    const text = "Please sign in to Alchemy";
+    const msg = ethUtil.bufferToHex(Buffer.from(text, 'utf8'));
+    const fromAddress = this.props.accountAddress;
+
+    let signature = localStorage.getItem("signature-" + fromAddress);
+    if (!signature) {
+      const method = 'personal_sign';
+      const sendAsync = promisify(web3.currentProvider.sendAsync);
+      const params = [msg, fromAddress];
```
jellegerbrandy: you should sign *all the parameters you send*, otherwise it provides very little security.

src/components/Account/AccountProfileContainer.tsx
jellegerbrandy: you should use https://web3js.readthedocs.io/en/1.0/web3-eth-personal.html#sign

src/components/Account/AccountProfileContainer.tsx
```diff
@@ -0,0 +1,255 @@
...
+      signature = result.result;
+      localStorage.setItem("signature-" + fromAddress, signature);
+    }
+
+    const recoveredAddress = sigUtil.recoverPersonalSignature({ data: msg, sig: signature });
+
+    if (recoveredAddress == this.props.accountAddress) {
+      await updateProfile(accountAddress, values.name, values.description, signature);
+    } else {
+      console.error("Signing in failed, please ensure you are logged in to MetaMask with an address that you own");
```
jellegerbrandy: What are you "signing in" for here? What does "logged in to metamask with an address that you own" mean?

src/components/Account/AccountProfileContainer.tsx
```diff
@@ -0,0 +1,255 @@
...
+      const sendAsync = promisify(web3.currentProvider.sendAsync);
+      const params = [msg, fromAddress];
+      const result = await sendAsync({ method, params, fromAddress });
+      signature = result.result;
+      localStorage.setItem("signature-" + fromAddress, signature);
+    }
+
+    const recoveredAddress = sigUtil.recoverPersonalSignature({ data: msg, sig: signature });
+
+    if (recoveredAddress == this.props.accountAddress) {
```
jellegerbrandy: how could this ever happen?

bohendo:
Why not use ENS and/or uPor? Let's give users a little sovereignty over their personal data. And on our end, has a legal team reviewed this implementation to ensure it's compliant with GDPR & other data protection regulations?

https://bitsonblocks.net/2017/05/17/gentle-introduction-self-sovereign-identity/

Personally, I'd love to see this private data server disappear rather than grow..

src/components/Account/AccountProfileContainer.tsx
dkent600: Arc.js doesn't go to the blockchain for the `deployed` calls.  It gets the addresses and code from the contract json files.  But it does for `at` calls.

dkent600: Further, Arc.js caches everything associated with getting the hydrated wrapper instance.  So as far as the Arc.js part of this is concerned, it is quite efficient.

src/actions/profilesActions.ts
```diff
@@ -0,0 +1,74 @@
...
+    const accounts = getState().arc.accounts;
+
+    let accountFilters = [];
+    for (const accountKey of Object.keys(accounts)) {
+      const account = accounts[accountKey];
+      accountFilters.push('{"ethereumAccountAddress":"' + account.address + '"}');
+    }
+
+    try {
+      const results = await axios.get(process.env.API_URL + '/api/accounts?filter={"where":{"or":[' + accountFilters.join(",") + ']}}');
```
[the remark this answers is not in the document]
tibetsprague: yeah i guess getting all the addresses for now makes sense, ill do that

src/components/Account/AccountProfileContainer.tsx
tibetsprague: its just using language people are more familiar with, but I agree it doesn't really map the flow we are presenting. I will change it.

src/components/Account/AccountProfileContainer.tsx
tibetsprague: If someone injected a current address into web3 that didn't come from MetaMask, or into the Redux state for that matter

tibetsprague:
@bohendo definitely agreed, this is a short term solution to experiment with Identity. We looked at using 3box for profiles (built on uPort) but it didn't quite seem ready for primetime, and it would be really slow to load all the profile information from OrbitDB for all the proposals given the current design of Alchemy. So we decided to just roll with this as an MVP for now while we let the identity solutions shake out a bit more.

src/components/Account/AccountProfileContainer.tsx
jellegerbrandy: Ok, perhaps i misunderstand. 
It seems to be that you are using the truffle-contract `deployed` and `at` functions, which are doing a network check by polling the blockchain:  https://github.com/trufflesuite/truffle/blob/next/packages/truffle-contract/lib/contract.js#L231,  and for the at() function here: https://github.com/trufflesuite/truffle/blob/next/packages/truffle-contract/lib/contract.js#L231

src/components/Account/AccountProfileContainer.tsx
jellegerbrandy: I don't know, perhaps you are searching for other things than I do.
This sstackexchange answwer also has patterns for older version of web3 or using another lib:
https://ethereum.stackexchange.com/questions/12033/sign-message-with-metamask-and-verify-with-ethereumjs-utils
https://gist.github.com/alexanderattar/29bef134239d5760b8d1fcc310b632be

src/components/Account/AccountProfileContainer.tsx
dkent600: @jellegerbrandy Ah, you are right it is checking for the ability to connect to the network (rather inefficiently I must say since it caches the result and could be just checking for the cached networkId.

In any case, after the first call Arc.js caches the result and ensuing calls will not access the net.

src/components/Account/AccountProfileContainer.tsx
tibetsprague: personal.sign is not supported by parity for example

src/components/Account/AccountProfileContainer.tsx
```diff
@@ -0,0 +1,274 @@
...
+    const web3 = await Arc.Utils.getWeb3();
+    const timestamp = new Date().getTime().toString();
+    const text = "Please sign this message to confirm your request to update your profile to name '" + values.name + "' and description '" + values.description + "'. There's no gas cost to you. Timestamp:" + timestamp;
+    const msg = ethUtil.bufferToHex(Buffer.from(text, 'utf8'));
+    const fromAddress = this.props.accountAddress;
+
+    const method = 'personal_sign';
+    const sendAsync = promisify(web3.currentProvider.sendAsync);
+    const params = [msg, fromAddress];
+    const result = await sendAsync({ method, params, fromAddress });
```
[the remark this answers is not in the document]
tibetsprague: this is happening locally

src/components/Account/AccountProfileContainer.tsx
```diff
@@ -0,0 +1,274 @@
...
+    showNotification(NotificationStatus.Success, `Copied to clipboard!`);
+    e.preventDefault();
+  }
+
+  public async handleSubmit(values: FormValues, { props, setSubmitting, setErrors }: any ) {
+    const { accountAddress, showNotification, updateProfile } = this.props;
+
+    const web3 = await Arc.Utils.getWeb3();
+    const timestamp = new Date().getTime().toString();
+    const text = "Please sign this message to confirm your request to update your profile to name '" + values.name + "' and description '" + values.description + "'. There's no gas cost to you. Timestamp:" + timestamp;
```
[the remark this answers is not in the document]
tibetsprague: actually no this is the message that a human reads and decides whether to sign

src/components/Account/AccountProfileContainer.tsx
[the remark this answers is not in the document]
tibetsprague: and like I said personal sign is not widely supported yet

PR #206, opened 2018-11-28
Decentralize proposals

bohendo:
This pull request is the product of a [proposal](https://gitlab.com/bohendo/alchemy/blob/master/proposals/decentralize-proposal-data.md) recently approved & funded by Genesis Alpha 

When a new proposal is submitted, save proposal data to IPFS rather than a private data server. A derivative of an IPFS-hash is stored as a proposal's `contributionDescriptionHash` on chain & can be used to retrieve the proposal data from IPFS as needed.

A script for fetching legacy data is included which provides backwards compatibility for all the data that has not been saved on IPFS.

Adopting these changes means the alchemy-server no longer needs to be run behind the client.


src/lib/ipfs.ts
```diff
@@ -0,0 +1,104 @@
...
+ipfs.hexToHash = (hex: string): string => {
+  if (hex.substring(0,2) === '0x') hex = hex.substring(2);
+  return bs58.encode(Buffer.from(ipfsHashPrefix + hex, 'hex'));
+}
+
+ipfs.add = (data: Object) => {
+  return promisify(ipfsApi.addJSON.bind(ipfsApi))(data);
+}
+
+const cache: any = {};
```
bohendo: Browsers cache API calls for us so caching like this isn't necessary client-side..
But this module also runs in node as part of the bg_cacher and in this environment it is worthwhile

src/lib/ipfs.ts
```diff
@@ -0,0 +1,104 @@
...
+    setTimeout(function() {
+      reject(new Error("timeout"))
+    }, ms)
+    promise.then(resolve, reject)
+  })
+}
+
+const ipfsHashPrefix = '1220';
+// Stores a static snapshot of the proposal data from before the ipfs migration
+const legacyIpfsHash = 'QmNaTgnLUAWneQwifeeLZHc5N1k2wqFKSfWPRjTS9ZjTxZ';
```
bohendo: IPFS nodes don't talk to each other very reliably, might need to add this file to your IPFS node manually.. 😕

bohendo:
Room for improvement: [IPFS file propagation is slow](https://gitlab.com/bohendo/alchemy/issues/12)

src/actions/arcActions.ts
```diff
@@ -674,19 +665,37 @@ export function createProposal(daoAvatarAddress: string, title: string, descript
...
+          Util.toWei(externalTokenReward),
+          "0",
+          "1"
+        ],
+        externalTokenAddress,
+        beneficiaryAddress
+      ]
+
+      const gasLimit = await contributionReward.estimate.proposeContributionReward(arg[0], arg[1], arg[2], arg[3], arg[4], arg[5])
+      const txReceipt = await contributionReward.proposeContributionReward(arg[0], arg[1], arg[2], arg[3], arg[4], arg[5], { gasLimit })
```
bohendo: We need to send this tx w/out the help of arc.js because of an [unfortunate feature](https://github.com/daostack/arc.js/issues/82)

dkent600: Alchemy has support from arc.js for tracking the tx lifecycle, so I am not sure we can bypass that.  I think best way to address this is probably to change the arc.js behavior.

src/actions/arcActions.ts
```diff
@@ -267,30 +276,17 @@ async function getProposalDetails(daoInstance: Arc.DAO, votingMachineInstance: A
 
   const submittedTime = Number(proposalDetails.submittedTime);
 
-  // Use title and description from the server
-  let description = "";
+  // Use title and url from the server
```
tibetsprague: server = ipfs

src/actions/arcActions.ts
```diff
@@ -186,11 +191,15 @@ export async function getDAOData(avatarAddress: string, currentAccountAddress: s
...
   let serverProposals: { [key: string]: any } = {};
   try {
-    const results = await axios.get(process.env.API_URL + '/api/proposals?filter={"where":{"daoAvatarAddress":"' + avatarAddress + '"}}');
-    serverProposals = _.keyBy(results.data, "arcId");
+    const promises = proposals.map(async (proposal) => {
+      const ipfsHash = ipfs.hexToHash(proposal.contributionDescriptionHash)
+      const result = await ipfs.get(ipfsHash)
+      return result
+    })
+    serverProposals = await Promise.all(promises)
```
tibetsprague: I think we should change the use of the word "server" everywhere to something else, maybe metaProposals? better ideas?

src/actions/arcActions.ts
```diff
@@ -301,7 +297,7 @@ async function getProposalDetails(daoInstance: Arc.DAO, votingMachineInstance: A
     boostedVotePeriodLimit: Number(proposalDetails.currentBoostedVotePeriodLimit),
     contributionDescriptionHash: contributionProposal.contributionDescriptionHash,
     daoAvatarAddress: avatarAddress,
-    description,
+    description: url,
```
tibetsprague: hmm, was there a final discussion around whether we will always be storing a URL in IPFS. Seems like one the main advantages of IPFS is that we could store the full description there and make sure it cant be changed after the fact

src/actions/arcActions.ts
```diff
@@ -632,37 +628,32 @@ export function createDAO(daoName: string, tokenName: string, tokenSymbol: strin
...
 
-      // Save the proposal title, description and submitted time on the server
+      const description = {
+        title,
+        url,
+        daoAvatarAddress,
+        submittedTime,
+      }
+
+      // Save the proposal title, url and submitted time on the server
```
tibetsprague: server = ipfs

src/actions/arcActions.ts
```diff
@@ -674,19 +665,37 @@ export function createProposal(daoAvatarAddress: string, title: string, descript
...
-        externalToken: dao.externalTokenAddress ? dao.externalTokenAddress : "",
-        externalTokenReward: dao.externalTokenAddress && externalTokenReward ? Util.toWei(externalTokenReward) : 0,
-        nativeTokenReward: Util.toWei(nativeTokenReward),
-        numberOfPeriods: 1,
-        periodLength: 0,
-        reputationChange: Util.toWei(reputationReward),
-      } as any);
+      const dao = getState().arc.daos[daoAvatarAddress];
+      const contributionRewardInstance = await Arc.ContributionRewardFactory.deployed();
+      const wallet = await getWallet()
```
tibetsprague: semicolons

src/actions/arcActions.ts
```diff
@@ -674,19 +665,37 @@ export function createProposal(daoAvatarAddress: string, title: string, descript
...
+      const contributionReward = new ethers.Contract(contributionRewardInstance.address, contributionRewardArtifacts.abi, wallet)
```
tibetsprague: adding ethers just for this seems unneccesary, you can get the truffle contract through contributionRewardInstance.contract and use that

src/actions/arcActions.ts
```diff
@@ -674,19 +665,37 @@ export function createProposal(daoAvatarAddress: string, title: string, descript
...
+          "1"
+        ],
+        externalTokenAddress,
+        beneficiaryAddress
+      ]
+
+      const gasLimit = await contributionReward.estimate.proposeContributionReward(arg[0], arg[1], arg[2], arg[3], arg[4], arg[5])
+      const txReceipt = await contributionReward.proposeContributionReward(arg[0], arg[1], arg[2], arg[3], arg[4], arg[5], { gasLimit })
+
+      console.log(JSON.stringify(txReceipt,null,2))
```
tibetsprague: remove console.log. this should have been caught by the linter

src/actions/arcActions.ts
tibetsprague: why do we need this?

src/actions/arcActions.ts
jellegerbrandy: Agree with Tibet here. Or use web3 (< 1.0 comes with arc.js)

src/actions/arcActions.ts
```diff
@@ -186,11 +191,15 @@ export async function getDAOData(avatarAddress: string, currentAccountAddress: s
...
   const proposals = [...votableProposals, ...executedProposals];
 
-  // Get all proposals' details like title and description from the server
+  // Get all proposals' details like title and url from the server
   let serverProposals: { [key: string]: any } = {};
   try {
-    const results = await axios.get(process.env.API_URL + '/api/proposals?filter={"where":{"daoAvatarAddress":"' + avatarAddress + '"}}');
-    serverProposals = _.keyBy(results.data, "arcId");
+    const promises = proposals.map(async (proposal) => {
+      const ipfsHash = ipfs.hexToHash(proposal.contributionDescriptionHash)
```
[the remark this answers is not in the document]
jellegerbrandy: plz wait with that, as we have many different branches in parallel now and if we start liting on one branch we may run into trouble merging them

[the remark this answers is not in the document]
jellegerbrandy: also I'd vote for configurating the linter to *never* use semicolons :-)

src/actions/arcActions.ts
jellegerbrandy: "proposals"

src/actions/arcActions.ts
```diff
@@ -10,7 +11,12 @@ import * as Redux from "redux";
 import { ThunkAction } from "redux-thunk";
 import { Web3 } from "web3";
 
+const contributionRewardArtifacts = require('@daostack/arc.js/migrated_contracts/ContributionReward.json');
```
jellegerbrandy: This looks a bit fragile. if you just need the abi can't you get it from the arc.js instance (the "contributionrewardinstancewrapper" or whatever it is called)?

src/actions/arcActions.ts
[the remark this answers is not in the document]
jellegerbrandy: More important, though: why do we need this conversino "hextohash"?

src/actions/arcActions.ts
```diff
@@ -267,30 +277,17 @@ async function getProposalDetails(daoInstance: Arc.DAO, votingMachineInstance: A
...
-        daoAvatarAddress: avatarAddress,
-        descriptionHash,
-        description,
-        submittedAt: submittedTime,
-        title
-      });
-    }
+    let ipfsHash = ipfs.hexToHash(descriptionHash)
+    let serverProposal = await ipfs.get(ipfsHash)
+    url = serverProposal.url;
```
jellegerbrandy: This can and will go wrong - hash not found, incorrect document found under ipfsHash . We need some error handling here.

src/actions/arcActions.ts
```diff
@@ -632,37 +629,32 @@ export function createDAO(daoName: string, tokenName: string, tokenSymbol: strin
...
-      const contributionRewardInstance = await Arc.ContributionRewardFactory.deployed();
-
-      const descriptionHash = Arc.Utils.SHA3(description);
       const submittedTime = Math.round((new Date()).getTime() / 1000);
 
-      // Save the proposal title, description and submitted time on the server
+      const description = {
+        title,
+        url,
+        daoAvatarAddress,
```
jellegerbrandy: why do we need this?

src/actions/arcActions.ts
jellegerbrandy: Ok, so @dkent600 , are you going to change this in arc.js?

src/lib/ipfs.ts
jellegerbrandy: The bg_cacher will be phased out soon. 
Apart from browser cache, we are saving this info also in the redux state. So I'd remove the cachaing machinery here.

src/lib/ipfs.ts
jellegerbrandy: What is "the" proposal here?

src/actions/arcActions.ts
bohendo: I absolutely want to keep the option to attach URLs to proposals, the commenting feature of google docs & flexibility of pointing to other platforms is too useful to do away with.

But I agree, an additional "description" field on the IPFS object would be good to finalize some chunk of text as a description & including both fields would give us best of both. Maybe we could throw a proposal's abstract/summary into it's description on IPFS and have a longer-form description with comments attached to a google doc via the url.

src/actions/arcActions.ts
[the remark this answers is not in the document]
bohendo: The IPFS file-id is a multihash encoded as base58 while the data stored on the blockchain is stored in hex. This function handles the conversion between raw hex in blockchain notion & the multihash required for interacting w ipfs nodes.

& If we want to keep the convention of adding unnecessary characters to the end of each line but don't want to use semicolons, how about we compromise & end each line with happy comments? `var isNiceCompromise = true || false // :)`

src/actions/arcActions.ts
bohendo: Proposals are built from both off-chain details and on-chain parameters so we'd probably want to reserve the `proposals` name for the complete chunk of proposal data.
"proposalDetails" is good but conflicts w existing vars so might be a pain to incorporate.
"offChainProposalDetails" seems most descriptive although a bit wordy.

src/actions/arcActions.ts
bohendo: I think logging this is useful, is there another logging framework I should use other than console.log?

src/actions/arcActions.ts
bohendo: Good suggestion, I like this idea better

src/actions/arcActions.ts
bohendo: If we use web3 or the truffle contract, then we don't. I'll remove it.

src/actions/arcActions.ts
bohendo: I'll look into using the truffle contract instance, if that works well then I can remove this line entirely.

src/actions/arcActions.ts
bohendo: Good catch, will look into handling edge cases better here

src/actions/arcActions.ts
bohendo: Following the convention set by the alchemy-server which stores: arcId, daoAvatarAddress, descriptionHash, description, submittedAt, and an id. To help this new data store fit into the existing system, I added the same fields with slight modifications eg renaming description to url.

src/lib/ipfs.ts
bohendo: "the proposal data" here is all the proposal data that's currently stored on the private data server aka the output of `curl https://daostack-alchemy.herokuapp.com/api/proposals`

src/actions/arcActions.ts
dkent600: @bohendo @jellegerbrandy  This change has already been released

src/actions/arcActions.ts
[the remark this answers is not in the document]
dkent600: Regarding linting, I would vote for being consistent across DAOstack projects.  Here is the one from arc.js which I think has been working nicely for a long time:

https://github.com/daostack/arc.js/blob/master/tslint.json

src/actions/arcActions.ts
dkent600: arc.js has a [logging subsystem](https://daostack.github.io/arc.js/api/classes/LoggingService/)

src/actions/arcActions.ts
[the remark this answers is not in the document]
jellegerbrandy: @bohendo: What do you mean "the data stored on the blockchain is stored in hex"? It is just a string (isn't it? If not, it should be :-))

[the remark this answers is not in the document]
bohendo: @jellegerbrandy Ultimately, all data is stored as binary whether it's stored in Ethereum storage or being passed around as an IPFS hash

Popular ethereum tools like web3.js will, by default, interpret this chunk of binary data (stored on the blockchain as a bytes32 I believe) as a hex string.

Popular IPFS tools like the default HTTP interface, expect to receive a chunk of binary data in the form of an IPFS hash aka base58 encoded w a hash-specifying prefix.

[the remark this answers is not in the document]
jellegerbrandy: well yes of course it is binary in the end, my question was a retorical and not very well phrased, sorry about that.

My point is that there is no need to convert from "raw hex in blockchain" to "mutliahsh required for interacting with ipfs nodes". You can just use the string that IPFS gives you, and pass that without changing it to the web3 library (for writing), read the same string back with web3 and give that string, unchanged, to your ipfs interface and get the file. 

The fact that this string happends to be a base58-encoded multihash is just not relevant at this level of the code. (I know, I wrote very similar code myself about 12 months ago)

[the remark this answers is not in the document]
bohendo: Strings in solidity are UTF-8 encoded and the contract reserves 32 bytes for the descriptionHash.

An IPFS hash represented as UTF-8 consumes 46 bytes of data. This string *could* be saved to the blockchain as-is but not within 32 bytes of storage.

Or, if we base58 decode it and drop the multihash prefix, we get 32 bytes of data which fits cleanly in this storage slot.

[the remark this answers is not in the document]
jellegerbrandy: Ok, I see now where you are coming from. I had not realized that the contracts used as `bytes32` field for the ipfs hash.

I think this is a mistake and should be changed. There is no guarantee that IPFS hashes fit in 32 bytes (indeed, as you notice,  they do not). It is **not** Ok to drop the multihash prefix - this is meaningful information that encodes the hashing schema and length. IFS can and WILL  change the default hashing schema in the future: https://github.com/ipfs/ipfs/issues/337

I'll create an issue in the arc repo to change the type of the hash field, if you agree with my reasoning.

src/actions/arcActions.ts
jellegerbrandy: there's this doc which can also be the locus for this discussion: https://docs.google.com/document/d/1mnhU_L5opxMGuCcdtgOb3D40KX4jquROnx2HvP1qG50/edit#

src/actions/arcActions.ts
jellegerbrandy: +1 for wordy vars :-)

src/actions/arcActions.ts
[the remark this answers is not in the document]
jellegerbrandy: https://github.com/daostack/arc/issues/577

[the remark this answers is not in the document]
bohendo: I'm glad you created this issue so we can make sure we're thinking about a better long-term solution. I definitely agree with your reasoning.

I'm not thrilled about dropping the prefix either but it seems like, given the current the contracts, this is a viable short term solution. Until the contracts change, I'm going to assume that this hacky solution works.

Important to note: we're already handling legacy data. When IPFS upgrades & breaks these old prefix-less hashes, we can incorporate them into the legacy data the same way we incorporated the data that was never saved to IPFS. So, this solution might not be future proof but it would be trivial to transition it to a future proof solution once one's available.

bohendo:
Apologies for letting this issue get stale, I submitted it immediately before diving into some other work and haven't had the time to get it cleaned up & pushed to completion.

I'll be prioritizing it in the next couple days though & will do what I can to get it merged by the end of next week.

src/actions/arcActions.ts
bohendo: @dkent600 Correct me if I'm wrong but this new release isn't useful here is it?

The [PR that fixes this](https://github.com/daostack/arc.js/pull/382) also upgrades the contract versions to v58 which we don't want yet.

So I don't think there's a version of arc.js available that has this bug fixes and is also using the current contracts (v50 I believe).

src/actions/arcActions.ts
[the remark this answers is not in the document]
bohendo: So until we migrate to new contracts, we have no choice but to use this not-so-future-proof way of storing the IPFS hash right?

jellegerbrandy:
Guys, @bohendo especially: I am closing this branch because in the Alchemy Earth, this will be handled by the client library, and was added in https://github.com/daostack/client/pull/54

PR #207, opened 2018-11-28
use a later version of flatmap-stream as the old package is removed

jellegerbrandy:
https://www.npmjs.com/package/flatmap-stream/v/11.1.3


jellegerbrandy:
Ok this did nto fix it, event-stream needs to be updated as well :-(

jellegerbrandy:
Cannot create an issue on the `event-stream` repo, as it is "archived"

PR #234, opened 2019-01-10
Adding comms info to README

EzraWeller:
The marketing department would like to add a bit of general comms info to the READMEs of some of our repos. Let us know if this stuff is a no-go for whatever reason.


jellegerbrandy:
sorry about merging this only now, Ezra

Issue #243, opened 2019-01-29
Trying to save a user profile crashes alchemy-server


tibetsprague:
Error
```
 /alchemy-server/common/models/account.js:11
alchemy-server_1      |       const timestamp = parseInt(ctx.data.timestamp);
alchemy-server_1      |                                           ^
alchemy-server_1      | 
alchemy-server_1      | TypeError: Cannot read property 'timestamp' of undefined
alchemy-server_1      |     at /alchemy-server/common/models/account.js:11:43
alchemy-server_1      |     at notifySingleObserver (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:160:22)
alchemy-server_1      |     at /alchemy-server/node_modules/async/dist/async.js:3110:16
alchemy-server_1      |     at replenish (/alchemy-server/node_modules/async/dist/async.js:1011:17)
alchemy-server_1      |     at /alchemy-server/node_modules/async/dist/async.js:1016:9
alchemy-server_1      |     at eachLimit$1 (/alchemy-server/node_modules/async/dist/async.js:3196:24)
alchemy-server_1      |     at Object.<anonymous> (/alchemy-server/node_modules/async/dist/async.js:1046:16)
alchemy-server_1      |     at doNotify (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:157:11)
alchemy-server_1      |     at doNotify (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:155:49)
alchemy-server_1      |     at doNotify (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:155:49)
alchemy-server_1      |     at doNotify (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:155:49)
alchemy-server_1      |     at Function.ObserverMixin._notifyBaseObservers (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:178:5)
alchemy-server_1      |     at Function.ObserverMixin.notifyObserversOf (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:153:8)
alchemy-server_1      |     at Function.ObserverMixin._notifyBaseObservers (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:176:15)
alchemy-server_1      |     at Function.ObserverMixin.notifyObserversOf (/alchemy-server/node_modules/loopback-datasource-juggler/lib/observer.js:153:8)
[...]

PR #246, opened 2019-01-29
Use master branch of alchemy-server

tibetsprague:
Fixes bug saving user profile: https://github.com/daostack/alchemy/issues/243


jellegerbrandy:
closign this as the change was included in an earlier PR

PR #273, opened 2019-02-02
Fix bug that prevented page header from appearing on profile page

tibetsprague:
@jellegerbrandy lets discuss this, not sure why you made the change that broke this, adding 
`  if (props.match) {` to AppContainer

### DIFF
--- a/src/components/Account/AccountProfileContainer.tsx
+++ b/src/components/Account/AccountProfileContainer.tsx
@@ -163,128 +163,127 @@ class AccountProfileContainer extends React.Component<IProps, IState> {
       <div>
         { dao ? <div className={css.daoHeader}><DaoHeader address={dao.address} /></div> : ""}
 
-      <div className={css.profileContainer} data-test-id="profile-container">
-        <h3>{ editing ? (accountProfile && accountProfile.name ? "Edit Profile" : "Set Profile") : "View Profile"}</h3>
-        { editing && (!accountProfile || !accountProfile.name) ? <div>In order to evoke a sense of trust and reduce risk of scams, we invite you to create a user profile which will be associated with your current Ethereum address.<br/><br/></div> : ""}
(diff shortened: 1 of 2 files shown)


jellegerbrandy:
The problemthat I so brutally "fixed" was that the header shows lots of ifno specific to the chosen DAO, and this breaks if no DAO is selected. For example, on the `daos` page here: http://127.0.0.1:3000/daos. This is still broken in this PR (and indeed the tests fail)

I had not noticed that the hader also is disappeared on the profile page. The reason is that the profile page uses a different addressing scheme then the other pages, i.e. we have:

http://127.0.0.1:3000/dao/0x46d6cdc1dc33a3bf63bb2e654e5622173365ed6a

http://127.0.0.1:3000/dao/0x46d6cdc1dc33a3bf63bb2e654e5622173365ed6a/members/

http://127.0.0.1:3000/dao/0x46d6cdc1dc33a3bf63bb2e654e5622173365ed6a/history/
http://127.0.0.1:3000/dao/0x46d6cdc1dc33a3bf63bb2e654e5622173365ed6a/proposals/create

http://127.0.0.1:3000/profile/0x90f8bf6a479f320ead074411a4b0e7944ea8c9c1?daoAvatarAddress=0x46d6cdc1dc33a3bf63bb2e654e5622173365ed6a

So perhaps the right way to fix this is to have more consistent routing, and cahnge the latter to 

`/dao/0xavataraddress/profile/0xprofileaddress`

What do you think?

tibetsprague:
@jellegerbrandy but people can view the profile without being inside a specific DAO. its optional to have DAO selected. but also nice for people who are inside of a DAO to be able to easily go back to it from their profile page. this is also a question for @A-Zak to decide how the UX should work

jellegerbrandy:
@A-Zak  we're waiting for you input here, thanks :-)

PR #286, opened 2019-02-11
[WIP] Alchemy Earth Frontend

jellegerbrandy:


jellegerbrandy:
Hi guys, I merged the latest branch here. It looks good, but we've got two failing tests. It would be great to fix these asap so we can merge this to `dev` adn work from a common code base again.

The two tests that are failing are:

- `test/integratinn/profile.ts` test: the "save" button is not clickable (no scrollbars, and the header is broken)
- `test/itnegration/proposal.ts`, the "create a proposal" test is brkoen because the "create-proposal" button has disappeared.

PR #293, opened 2019-02-14
[WIP] threshold calculation

tibetsprague:


jellegerbrandy:
@tibetsprague i'm closing this as it is old and we should get this info from the subgraph

Issue #294, opened 2019-02-14
define what to do on network/account changes


jellegerbrandy:
Alchemy gets its current state from an external provider (metamask for now, portis maybe as well later).

the user can change this while using the app. Alchemy used to reload the entire app when he accountchanges, which seems a bit overkill. Here's a proposal:

(1) quick iteration one: we reload just as alchemy does

(2) ideal behavior:

- when network changes: reload entire app. This makes ok bc it will typically never happen (only for devs), and it is good bec it is a kind of warning "you are loading anew app"

- when account address changes: only update the components that are affected. To follow the pattern we ahve been using, this would imply implementing an observable `arc.currentAccount` that sends a new account through the pipe when it changes.

tibetsprague:
Alchemy used to only update some parts when account changed but there kept being edge cases that we missed so thats why we decided to just reload... we can try again but doesnt seem like high priority right now to make this work. people won't change their accounts very often either, only devs

jellegerbrandy:
related: #326 and #338

A-Zak:
#227 also related :)

jellegerbrandy:
I'm closing this as we have something working now, even if it is not ideal :-)

PR #327, opened 2019-02-26
detect network on load, and use relevant providers in arc

jellegerbrandy:
addresses #294


src/arc.ts
```diff
@@ -1,29 +1,54 @@
 import { Arc } from "@daostack/client";
+import Util from "lib/util";
 import { Observable } from "rxjs";
 
 const Web3 = require("web3");
 
+const INFURA_KEY = "16bDz7U53RbXysQiYOyc";
```
tibetsprague: This shouldn't go in the codebase, it should be read from an environment variable (you can put it in a .env file for development)

src/arc.ts
```diff
@@ -1,29 +1,54 @@
...
+
+const providers = {
+  dev: {
+    graphqlHttpProvider: "http://127.0.0.1:8000/subgraphs/name/daostack",
```
tibetsprague: I have been putting these global config variables in env vars too, so we can set them on Heroku. See webpack.dev.config.js in the DefinePlugin section

src/arc.ts
```diff
@@ -77,28 +105,54 @@ export function getWeb3Provider() {
...
+  // if not, we ????
+  if (web3Provider && web3Provider.isMetaMask) {
+    console.log(web3Provider);
+    const networkName = Util.networkName(web3Provider.networkVersion);
+    if (networkName === "ganache") {
+      console.log(`Connected to ${networkName} - this is great`);
+    } else if (networkName === "kovan") {
+      console.log(`Connected to ${networkName} - this is not working yet but will soon`);
+    } else {
+      console.warn(`YOU ARE NOT CONNECTED TO GANACHE (but to ${networkName}) - please switch your metamask connection to Kovan or Private network`);
```
tibetsprague: We can show real notification bubbles for this too using showNotification

src/arc.ts
```diff
@@ -77,28 +105,54 @@ export function getWeb3Provider() {
...
-    });
+    // if we are connected with metamask, we find the right settings
+    let arcSettings: any;
+    if (web3Provider && web3Provider.isMetaMask) {
+      const networkName = Util.networkName(web3Provider.networkVersion);
+      if (networkName === "ganache") {
+        console.log(`Connected to ${networkName} - this is great`);
+        arcSettings = providers.dev;
+        console.log(arcSettings);
+      } else if (networkName === "kovan") {
```
tibetsprague: sounds like we might use Rinkeby instead of Koran for our testate

src/arc.ts
```diff
@@ -77,28 +105,54 @@ export function getWeb3Provider() {
...
+        console.log(`Connected to ${networkName} - this is great`);
+        arcSettings = providers.staging;
+        console.log(arcSettings);
+      } else {
+        const msg = `YOU ARE NOT CONNECTED TO GANACHE (but to ${networkName}) - please switch your metamask connection to Kovan or Private network`;
```
tibetsprague: Again, visible notifications would be really helpful for users

src/arc.ts
dkent600: I personally prefer to use a file containing configuration information, rather than environment variables.  I don't like that the app has to know anything so closely tied to the OS.  A config file can be structured to have env-specific sections, like "dev", "staging", "production" or "kovan", "ganache", "live", etc.  Such a config file can be maintained in the repo if desired.

tibetsprague: but we dont want to share our INFURA key publicly with the world, thats the whole point of using environment variables.

jellegerbrandy: I'd agree in general, but there is notthing secret about the infura key, it is sent in plain text with each request.

src/arc.ts
jellegerbrandy: yeah, but the question is _when_ to show those...

src/arc.ts
jellegerbrandy: Okay. Always a difficult call to decide where to put these kind of things, but environment variables seem a good place. Mind postponing the decision a bit still, also in the light of a possible change of deployment platform?

PR #333, opened 2019-02-28
Correctly subscribe to ETH balances everywhere

tibetsprague:
Also double extra be sure that all balances are BNs


src/components/Account/AccountBalances.tsx
```diff
@@ -82,7 +82,7 @@ export default (props: { dao: IDAOState, address: Address}) => {
         } else if (state.error) {
           return <div>{state.error}</div>;
         } else {
-          return <AccountBalances dao={props.dao} member={state.data[0]} ethBalance={state.data[1]} genBalance={state.data[2]} genAllowance={state.data[3] ? new BN(state.data[3].amount) : new BN(0)} />;
+          return <AccountBalances dao={props.dao} member={state.data[0]} ethBalance={new BN(state.data[1])} genBalance={state.data[2]} genAllowance={state.data[3] ? new BN(state.data[3].amount) : new BN(0)} />;
```
jellegerbrandy: So with the new client update, casting these numbers to BN should not be needed anymore, right?

tibetsprague: yeah

PR #334, opened 2019-03-01
Docker containers

jellegerbrandy:


.travis.yml
```diff
@@ -29,10 +29,9 @@ script:
   - npm run service-status
 
   # setup the environment
-  - echo 'npm run setup-env' && echo -en 'travis_fold:start:script.4\\r'
-  - npm run setup-env
   # move the migration.json file froms etup-env into the docker container
-  - cp node_modules/@daostack/subgraph/migration.json config/
+  # TODO: we really need this?
```
tibetsprague: well its been helpful to find it here and be able to copy it into the client library for running the tests there when doing local dev on that and alchemy at the same time

PR #354, opened 2019-03-07
337 websockets

jellegerbrandy:


src/components/Proposal/ProposalContainer.tsx
```diff
@@ -472,14 +471,13 @@ class ProposalContainer extends React.Component<IProps, IState> {
...
                   proposal={proposal}
-                  stakeProposal={stakeProposal}
                   threshold={threshold}
                   approveStakingGens={approveStakingGens}
                   detailView={detailView}
                 />
               }
           </div>
-        </div>
+        </div >
```
brodeur: Whitespace ^_^

src/components/Shared/PreTransactionModal.tsx
```diff
@@ -211,7 +198,7 @@ export default class PreTransactionModal extends React.Component<IProps, IState>
             </div>
             <div className={css.proposalInfo}>
               <div className={css.proposalTitle}>
-                <strong>{proposal.title || "[no title]"}</strong>
+                <strong>{proposal.title || proposal.id}</strong>
```
brodeur: Isn't proposal.id a hash? Is that better to show than "No title"? Could also be: "Title unavailable"

src/components/Proposal/PredictionBox.tsx
```diff
@@ -224,7 +230,7 @@ export default class PredictionBox extends React.Component<IProps, IState> {
           </div>
 
           <div className={css.enablePredictions}>
-            <button onClick={this.showApprovalModal.bind(this)}>Enable Predicting</button>
+            <button onClick={this.showApprovalModal }>Enable Predicting</button>
```
brodeur: Whitespace

src/components/Shared/PreTransactionModal.tsx
A-Zak: Can we please combine the two but only using part of the id?
Like this - [No title 0xc839...8eee]

jellegerbrandy: Ok

Issue #325, opened 2019-02-26
Better handle not connected users


jellegerbrandy:
When we have no metamask connection we should not show the account avatar on the top right.

tibetsprague:
@A-Zak what shall we show while we wait for the account to load, or if MetaMask is not enabled? Nothing? Maybe a button to connect to MetaMask?

jellegerbrandy:
A button like "please log in"?

jellegerbrandy:
Ok I implemented this in #350  - it needs a design @A-Zak  and styling @brodeur

jellegerbrandy:
And with "implement this" I mean there is now a text that says "please log in". It should probably be a link to metamask or something?

A-Zak:
I think please log in is good :)

jellegerbrandy:
I choose to write "please log in"  because i thought that it was obvious that that was not good :-) (what are you "logging in" to?)  And we cannot prompt metamask to popup, because one of the possible cases is that metamask is not installed.

What we need is some clear explanation that we can use in all the parts of the app where this is relevant, and soem way to show that explanation. Here are some situations where we shoudl inform the user to "log in" 

- the "please log in" 
- on the rewards page 
- when initiating a transaction
- "you cannot vote right now"
- [maybe others]

the text is probably smoething of the form "If you do not have metamsk, please intall it, create an account, and get yourself some GEN/REP (???); if instead you do have metamask, please open it and select your account". We can (I think) detect is metamask is installed in the browser, so this text can be dynamic, depending on the situation. 

I'm not sure if this should be a separate page or a popup. We need specs @A-Zak  :heart:

A-Zak:
Ok, now I get what you meant, I'll put it on my list then

PR #419, opened 2019-03-13
Activate Please log in button in header

tibetsprague:
Still needs styling from @brodeur


jellegerbrandy:
This is actually more complex, and I do not thing "enable metamask" is the correct semantics for that button. We've got an ongoing discussion here: https://github.com/daostack/alchemy/issues/325 (see the last comments)

tibetsprague:
Well yes, thats true, but aren't we waiting on more design details from Alex? In the meantime this allows us to use Alchemy with MetaMask and privacy mode turned on

tibetsprague:
@A-Zak could use more of your input here too. we do need a button like this to allow people to work with MetaMask's privacy mode on, which is now the default

jellegerbrandy:
I borrowed some of yor logic for PR #485 , closing this one...

PR #431, opened 2019-03-14
do not use MM user account if not connected to the right network

jellegerbrandy:


tibetsprague:
there are no changes here :)

Issue #317, opened 2019-02-25
Get transaction notifications working again (now without arc.js)


tibetsprague:

jellegerbrandy:
hey @tibetsprague I understand you started on this but do not know where... (also, i have opinions on this :-))

tibetsprague:
@jellegerbrandy I started pondering and poking around but really didn't make significant enough process to share. Would love to hear your thoughts :)

dkent600:
@tibetsprague  The arc.js TransactionService docs page is a good reference for how it has been meant to be used:

https://daostack.github.io/arc.js/Transactions/

jellegerbrandy:
Ok, well, while the arc.js system is very impressive and versatile, I'd like to do something more lightweight for alchemy. And that means starting with a description of what we actually want to see there. I suppose two things:

1. transaction state updates (i.e. "sent", "mined", "confirmed", "twice confirmed"...)
2. [maybe] have a list of pending transactions that is persistent over reload

Now, for 1. the client library has provides an observable of a stream of udpates, i.e. if `vote()` is the action,  you can do something like:

```
vote().subscribe((txUpdate) => {
   if (txUpdate.state === 'sent' ) { notify("your transaction has been sent, waitin'for it to be mnied") }
   if (txUpdate.state === 'mined'} {
           notify("your transaction has been mined! It was confirmed ${txUpdate.confirmations} times"}
           // and we also ahve the txUpdate.receipt and the txUpdate.result to do stuff with
    }

}}
```

jellegerbrandy:
@tibetsprague , I started implementing this in #439

jellegerbrandy:
We had a short discussion on discord about how many confirmations to show. 

We could sow:  1 confirmation = three notifications: "transaction sent", "transaction mined", "transaction confirmed".

@A-Zak  what do you say?

PR #439, opened 2019-03-18
Cache management and other fixes

### LINKED ISSUE #400: Proposal Cards are not updated with latest state
To reproduce (there are different ways):

- Boost a proposal. It will move to boosted, but still show the "boost" button

### LINKED ISSUE #438: connection state warnings are alerts, should become notifications

jellegerbrandy:
- fix bug in calculating total rewards
- update client to 0.0.41, which as more aggrssive fetch policy - this fixes #400 
- use notifications for connection status instead of alerts - this fixes #438 
- restore linter triple-equals rule
- show notifications on proposal execute action [a first step towards implemeting #317

### DIFF
--- a/src/reducers/arcReducer.ts
+++ b/src/reducers/arcReducer.ts
@@ -4,7 +4,7 @@ import * as moment from "moment";
 import { CreateProposalAction, RedeemAction, StakeAction, VoteAction } from "actions/arcActions";
 import { AsyncActionSequence } from "actions/async";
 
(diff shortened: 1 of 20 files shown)


src/reducers/arcReducer.ts
```diff
@@ -170,7 +170,7 @@ export interface IProposalState {
   state: ProposalStates;
   submittedTime: number;
   title: string;
-  transactionState: TransactionStates;
+  ITransactionState: TransactionStates;
```
tibetsprague: this looks weird and maybe wrong

tibetsprague: but we dont use this anymore...

PR #452, opened 2019-03-19
Remove hover states

brodeur:


src/components/Account/AccountPopupContainer.tsx
```diff
@@ -68,9 +72,16 @@ class AccountPopupContainer extends React.Component<IProps, null> {
   }
 
   public render() {
-    const { accountAddress, dao, profile, reputation } = this.props;
+    const { accountAddress, avatarSize, dao, detailTransferDetailView, profile, reputation } = this.props;
+
+    let className = " ";
+
+    if (detailTransferDetailView) {
```
tibetsprague: you should use the classNames function like we do below for this

jellegerbrandy:
@brodeur this PR is 20 days old, what is the status of it?

tibetsprague:
@jellegerbrandy we decided to hold off on this for now, but might still want it later

jellegerbrandy:
Ok @tibetsprague I'm closing this so we do not have to see it each time, we can always reopen when we need it....

PR #491, opened 2019-03-26
upgrade client dependency to 0.0.45

jellegerbrandy:
- upgrade client dep to 0.0.45
- implement the logic from https://github.com/daostack/alchemy/issues/493

(some of this PR is already deployed to staging)


src/arc.ts
```diff
@@ -127,7 +127,7 @@ export function checkNetwork() {
     }
 
     // check if Metamask account access is enabled, and if not, ask to enable it
-    enableMetamask();
+    // enableMetamask();
```
tibetsprague: Why is this commented out now?

jellegerbrandy: it is commented out because like this it would pop up a metamask dialog on loading the page, and I was in doubt on wether that is desired behavior yes or no. 

So I made a new issue with specs: https://github.com/daostack/alchemy/issues/493  According to these specs, we should _not_ call "enableMetamask" here. But before makign the definitive decision, I'd like some feedback on that issue, both from you @tibetsprague as wella s from @A-Zak

jellegerbrandy: ímplemented the specs

src/arc.ts
```diff
@@ -88,16 +88,37 @@ export function subscribeToAccountChanges(web3: any, currentAccountAddress?: Add
...
  * @return
  */
-export function checkNetwork() {
+export async function checkNetwork() {
   const web3: any = getArc().web3;
   const web3Provider = web3.currentProvider;
   if (web3Provider && web3Provider.isMetaMask) {
     // we are interacting with Metamask, let's just use window.ethereum to interact with MM
     const ethereum = (<any> window).ethereum;
+    // for (let i = 0; i < 10; i++) {
```
tibetsprague: ??

src/arc.ts
```diff
@@ -120,31 +141,47 @@ export function checkNetwork() {
...
+ * get the current user from the web3 provider (metamask)
+ * this function will throw an Error if there is anything wrong with the connection
+ * (when no provider is available, or the provider is not the expected provider)
+ * which need to be handled by the caller of the function
+ * @return [description]
+ */
+export async function getCurrentUser(): Promise<Address> {
+  await checkNetwork();
+  return (<any> window).ethereum.selectedAddress;
+  // const arc = getArc();
```
tibetsprague: ?

src/arc.ts
jellegerbrandy: ok ok, so I am messy programmer :-)

PR #541, opened 2019-04-03
[WIP] 383 new dao statistics

tibetsprague:
shows the number of reputation holders and open proposals in the DAO. Still need to get the amount of GEN staked on open proposals working, which im not sure if we should do or the client should do?


jellegerbrandy:
the client (actually subgraph) should calculate the totals

Issue #510, opened 2019-03-28
String preparation for staking amounts


brodeur:
Never show less than .000...

0-1
< .001 - .999

0-100
00.00 - 99.99

100-1000
.1k - 1k

1000 - 100,000
1k, 100k

100,000 - 100,000,000
.1M - 100M

tibetsprague:
Meaning if its less than .000 we just show .000? what about the really tiny stake amounts that the DAO makes on new proposals?

jellegerbrandy:
Guys, we CANNOT show 0 if the value is different from 0. A proposal might pass or move to boosted because of such a value, and the user will be utterly confused.

jellegerbrandy:
This holds for all differences: We cannot show "5.00% for and 5.00% against" and have the propsal fail because against was really "5.0000001".

jellegerbrandy:
(I realize I am only noting a problem and not proposing a solution :-/)

tibetsprague:
we could show 1e-10 for very small amounts, but tiny differences in percentage sound hard, maybe its enough though to just show which side is winning even if the numbers look the same. or we could use a > symbol like >5.00%

tibetsprague:
or 5.00+% 
 
and put whole number in the tooltip

jellegerbrandy:
To recapt what we said on the call yesterday:

- We  do not care so much about precision when showing balances and payouts and such 
- the only thing that a different of 0,000001% is relevant is when comparing values (votes or stakes)
- in the comparison widget, we aan also see which vote is witting (thumbs up, thumbs down, red line whatever). So whowing "5.00% for, 5.00% against, against wins" is ok
- we want a consistent rep throught the app (say two decimals + a mouseover with the whole precision)
- we have nto really decide what to do with small or big numbers (use the `e` notation, or the `100K` notatin.

A-Zak:
I prefer 100k notation, as noted in the first comment...

A-Zak:
I think zeros show up as '~0.001'.... if it's an absolute zero, please display them as '0'. :)

<img width="635" alt="Screenshot 2019-04-10 at 1 14 34" src="https://user-images.githubusercontent.com/302774/55841532-2f37e800-5b38-11e9-949c-a9e086fcc1ac.png">

A-Zak:
@tibetsprague I've tried this out today, and it doesn't look so good... /:
[...]

PR #546, opened 2019-04-04
Create function for consistently formatting tokens throughout the app

tibetsprague:
Fixes https://github.com/daostack/alchemy/issues/510


tibetsprague:
hmm that's a good point about rounding, it is not rounding up right now and javascript has terrible functions for rounding. Ill ponder

jellegerbrandy:
@tibetsprague i fixed the buidl error here. Not sure about the rounding down, though. Shall we merge tis and make an issue for it?

PR #563, opened 2019-04-08
Fix network warning

tibetsprague:
This is kind of a hack for now because the only function i could get working consistently is web3.eth.net.getNetworkType which returns a string and doesnt know if we are connected to ganache. When we upgrade web3 we should be able to use web3.eth.getChainId


jellegerbrandy:
Basically, instead of aksing the Metamask directly which network it is connecting to, in this PR we do request to the ethereum node and ask the node which network it is. 

I don't like this very much because it will make all these actions slower

Let me try it out before merging :-)

Issue #603, opened 2019-04-14
Some proposals are missing in the queue O_O


A-Zak:
Some of the proposals are missing from the queue. The count seems to be correct (currently 10), but only shows 6 proposals. 

For example this one: 0xf038967ac6e37121f54d5d4168df9d717282d63fac3f01f4c7e4b216268dfe6a

jellegerbrandy:
For reference, this is the (ALMOST) the query: 

https://subgraph.daostack.io/subgraphs/name/v9/graphql?query=%7B%20%20proposals%20(where%3A%20%7B%20stage%3A%20%22Queued%22%20%7D)%20%7B%0A%20%20%0A%20title%0A%20%20expiresInQueueAt%0A%7D%0A%7D

It seems the total number is right, but we are indeed not rendering all proposals.

jellegerbrandy:
(note that we are not showing ALL queued proposals under "regular proposals" - we are showing only those that are not `expiredInQueue`)

A-Zak:
Right, but this proposal is still way before it's `expiredInQueue` (May 5th)...

@tibetsprague can you take a look... also, if you [navigate to directly](https://alchemy.daostack.io/dao/0x294f999356ed03347c7a23bcbcf8d33fa41dc830/proposal/0xf038967ac6e37121f54d5d4168df9d717282d63fac3f01f4c7e4b216268dfe6a), it seems like we do get the proposal title in the breadcrumbs, but not the proposal container...
Seems something is crashing while rendering it...

A-Zak:
Maybe a clue - I'm pretty sure this proposal has been boosted and then unboosted...
hope this helps...

tibetsprague:
PR out and ready to go pending a review

PR #619, opened 2019-04-16
Don’t hide some unboosted proposals

tibetsprague:
Basically if a proposal was boosted then unboosted we were hiding it when it reached what was previously the end of its pre boosting time period.

Fixes https://github.com/daostack/alchemy/issues/603

### DIFF
--- a/src/reducers/arcReducer.ts
+++ b/src/reducers/arcReducer.ts
@@ -186,12 +186,21 @@ export const initialState: IArcState = {
 };
 
 export const closingTime = (proposal: IProposalStateFromDaoStackClient) => {
-  // TODO: how does quiet ending period play into this?
-  const start = proposal.boostedAt || proposal.preBoostedAt || proposal.createdAt;
-  const duration = proposal.boostedAt ? proposal.boostedVotePeriodLimit :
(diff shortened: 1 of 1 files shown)


src/reducers/arcReducer.ts
```diff
@@ -186,12 +186,21 @@ export const initialState: IArcState = {
...
-  const duration = proposal.boostedAt ? proposal.boostedVotePeriodLimit :
-                     proposal.preBoostedAt ? proposal.preBoostedVotePeriodLimit :
-                      proposal.queuedVotePeriodLimit;
-  return moment((proposal.executedAt || start + duration) * 1000);
+  switch (proposal.stage) {
+    case IProposalStage.ExpiredInQueue:
+    case IProposalStage.Queued:
+      return moment((proposal.createdAt + proposal.queuedVotePeriodLimit) * 1000);
+    case IProposalStage.PreBoosted:
+      console.log(proposal.preBoostedAt, proposal.preBoostedVotePeriodLimit, (proposal.preBoostedAt + proposal.preBoostedVotePeriodLimit) * 1000, moment((proposal.preBoostedAt + proposal.preBoostedVotePeriodLimit) * 1000));
```
jellegerbrandy: hey man a console.log!

PR #632, opened 2019-04-17
Add Zendesk snipper to Alchemy

tibetsprague:
doesnt seem to work locally, but maybe it will on prod?


jellegerbrandy:
This is blocked by the fact that we need to confiugre zendesk, right? Is there an issue for that, @A-Zak , so we can track it?

Issue #599, opened 2019-04-14
Scheme registrar


A-Zak:
Todo
======
- [x] Client support for creating a scheme registrar support
- [x] New proposal
- [x] Proposal Card
- [x] Proposal Details

Scheme Registrar specs
===================

Scheme registrar queue has three proposal types: Add scheme, remove scheme and edit scheme.
The three proposal types are accessible in the same create proposal form, via the tabs on the bottom part of the form.

Scheme registrar proposals
======================
* Remove scheme (tab title)
    * zeplin - https://zpl.io/2GyzQL4
    * Parameters
        * Scheme to remove (select)
            * The user can select one of the existing schemes
            * the list should display the scheme names or address if no name
    * Proposal summary
        * Icon - Trash icon
        * Text - Remove scheme
        * Extended details (proposal details):
            * Scheme name or address + link to etherscan + icon for is varified
* Add scheme (tab title)
    * Zeplin - https://zpl.io/2p4oOEQ
    * Parameters
        * Scheme to add (select / text)
            * The user can select one from the supported schemes (scheme registrar, generic action & contribution reward)
            * The user can select other, and supply the address of the new scheme (zeplin - https://zpl.io/2GyzAA4)
        * Parameter hash (text)
            * The parameter hash to use with the scheme
        * Scheme permissions (checkboxes)
            * The user can select which permissions to give the scheme
    * Proposal summary
        * Icon - plus icon
        * Text - Add scheme
        * Extended details (proposal details):
            * Scheme name
            * Scheme address + link to etherscan + icon for is verified
            * Param hash (can we read params? Doesn't have to be now)
            * The scheme permissions 
* Edit scheme (tab title)
    * Edit scheme is very similar to add scheme (uses the same method). The differences are:
    * Parameters
        * Scheme to add (select)
            * The user can only select one of the schemes in the dao
    * Proposal summary
        * Icon - pencil icon
        * Text - Edit scheme

A-Zak:
Some UI bugs:
![image](https://user-images.githubusercontent.com/302774/57622691-d2947680-7596-11e9-8e65-c623063685f3.png)
![image](https://user-images.githubusercontent.com/302774/57633609-78061500-75ac-11e9-8091-63a49b759c3b.png)

tibetsprague:
@A-Zak parameter hash is required

leviadam:
@tibetsprague 
Why is it required?
If it is empty we can just have zero, some schemes do not need params.

[...]

Issue #384, opened 2019-03-12
All schemes view


A-Zak:
- [x] scheme card
- [x] multiple schemes layout
- - - - - - 
- [ ] fix visual bugs

A-Zak:
Added some notes on these screenshots:
![image](https://user-images.githubusercontent.com/302774/57620659-1769de80-7592-11e9-9fd2-2a54761553ef.png)

![image](https://user-images.githubusercontent.com/302774/57621112-4e8cbf80-7593-11e9-9df5-f0f91f5b9e9e.png)

![image](https://user-images.githubusercontent.com/302774/57621255-9e6b8680-7593-11e9-9550-3a257de6cb03.png)


Also, some regressions from the production version:
=========
![image](https://user-images.githubusercontent.com/302774/57621467-30738f00-7594-11e9-9460-59cf18997be8.png)

![image](https://user-images.githubusercontent.com/302774/57620849-a2e36f80-7592-11e9-95b0-1a500058df3d.png)

tibetsprague:
@A-Zak global constraint registrar and upgrade scheme are actual schemes on the DAO. we could hide them since we dont support proposals for them...

brodeur:
The only part I haven't addressed here is related to Voting Disabled regression -- what should it look like? I can't find it in the designs for the detailView.

brodeur:
@A-Zak

PR #655, opened 2019-04-29
[WIP] Multiple schemes

jellegerbrandy:


A-Zak:
I've added some visual bugs on the two related issues: #599 & #384 

Also creating a few general issues:
#695, #696, #697, #698

src/index.tsx
```diff
@@ -18,6 +19,7 @@ async function renderApp() {
     environment: process.env.NODE_ENV
   });
 
+  await initializeArc();
```
jellegerbrandy: is this a good idea? If this fails or takes a long time, we have no good way of showing user feedback...

Issue #691, opened 2019-05-12
gen to boost on failed (time out) proposal ? 


orenyodfat:
https://alchemy.daostack.io/dao/0xecd657fe5c8ea9bc62af84c06e2a2f48e57079fa/proposal/0x6d8ebe983a6f8063feffd4a97286914db40839ba5f9087aae3eabac72f616709

![Screen Shot 2019-05-12 at 13 29 41](https://user-images.githubusercontent.com/5994398/57581031-11a0ca00-74ba-11e9-8aae-a86b16566feb.png)

orenyodfat:
In this case the user cannot claim back its stakes (false)

A-Zak:
Is this issue about the label "11.4 GEN to boost"?

> In this case the user cannot claim back its stakes (false)

@orenyodfat what do you mean by this?

orenyodfat:
> Is this issue about the label "11.4 GEN to boost"?

yes, and also a boost button appears. 

> Also
> @orenyodfat what do you mean by this comment?
> 
> > In this case the user cannot claim back its stakes (false)

in this case the user who downstake as no button to redeem back its stake.

orenyodfat:
does these UIs issues are not covered by the manual UI tests spec ? is there such a spec ?

jellegerbrandy:
This is a straighforward ui bug, let's fix it tomorrow.

tibetsprague:
I'm actually still not clear what the issues are here

orenyodfat:
1. no need to present predictions banner + boost button + gen to boost on already timed out proposal which is part of the history,
2. user which downstakes on timeout proposal should be able to get back their stakes somehow.. though there is no ui for that ..(it does not appears on the redemption tab)

PR #702, opened 2019-05-13
Don’t show stake buttons or boost amount in proposal details for queu…

tibetsprague:
…ed but expired proposals


tibetsprague:
Fixes https://github.com/daostack/alchemy/issues/691

A-Zak:
This fixes the staking / prediction stuff. But it still doesn't allow me to redeem the 5 gen I staked against this proposal.

@jellegerbrandy, should we merge this and open another issue, or should we fix it in this pr?

image

orenyodfat:
#691 contain both ui + redeem

PR #714, opened 2019-05-16
Cleaned up buy GEN menu

brodeur:


A-Zak:
Two comments:
- [ ] 1) it seems we have two lists of "place to by GEN at" (one for the side menu, and one for the staking popup). We should consolidate it to have just one source of truth.
Screenshot 2019-05-16 at 15 03 03
Screenshot 2019-05-16 at 15 02 49


- [ ] 2) let's align the icons and labels a bit better... 
alignment guide

A-Zak:
Also, please put BitFinex above EthFinex... :)
They are after all the parent company :)

leviadam:
@brodeur I think ethfinex is still not aligned in the menu.

leviadam:
Also, the two menus are different, there is no one source of truth as @A-Zak called it.

brodeur:
Alright, I think I got it to where we want it after talking with Alex today. Let me know if further changes are needed

leviadam:
Actually @brodeur the icons are not centered.
In addition, the menu is not read from one source of truth.

leviadam:
OK. We can probably delay the single source of truth. Better merge this.

A-Zak:
@jellegerbrandy Does it matter what we do with this conflict in the package-lock important? or will it work either way? :)

A-Zak:
If not, can you please merge? :) 🙏

Issue #600, opened 2019-04-14
Generic Action (dxDAO)


A-Zak:
Todo
======
- [ ] New proposal
- [ ] Proposal Card
- [ ] Proposal Details

Generic scheme specs
==================

The generic scheme queue (for the dxDAO) will feature 4 proposal categories, with a total of 7 types of proposals (details below).

When clicking the create new proposal button it should show the list of categories. Changing between the proposal types within the category is done within the proposal creation form (via the tabs on the bottom part of the form).

The design for proposal summary (on proposal cards and proposal details page) will vary mostly on icons and text (details below).

The design for creating these proposals is just like the scheme registrar with differences in titles tabs and fields (zeplin - https://zpl.io/a7WJRQ8).

Upgrade the DutchX (category)
=========================
* Update mastercopy (tab title)
    * (wait for 30days, after the cooldown an additional tx is required to make it effective. in the future this might be possible from within alchemy, but not for now).
    * Parameters
        * New Contract address
    * Proposal summary
        * Icon - edit pencil
        * Text - Update Mastercopy 
        * Extended details (proposal details):
            * New master copy address -> link to etherscan + icon for is verified
* Change ETH:USD oracle (tab title)
    * Same as above except:
    * Proposal summary
        * Text - Change ETH:USD oracle
* Change DutchX owner (tab title)
    * Parameters
        * New owner address
    * Proposal summary
        * Icon - edit pencil
        * Text - Change DutchX owner
        * Extended details (proposal details):
             * New owner address -> link to etherscan

Update whitelisted tokens
=====================
* Add tokens
    * Parameters
        * Token address(s)
            * Should start as one field, and a plus button to add another field
    * Proposal summary
        * Icon - plus icon
        * Text - Whitelist 5 token(s) - only show 's' if more then one token
        * Extended details (proposal details):
            * List of tokens to be added with links to etherscan
* Remove tokens
    * Same as above except:
    * Proposal summary
        * Icon - minus icon
        * Text - Delist 3 tokens(s)

Update thresholds
===============
* Update new token pair threshold
    * Parameters
        * New threshold
            * Value in USD (contract expects the value * 10^18)
    * Proposal summary
        * Icon - pencil icon
        * Text - Set new token pair threshold to $7.4k
* Update new auction threshold
    * Same as above except:
    * Proposal summary
        * Text - Set new auction threshold to $3.2k

jellegerbrandy:
[...]

PR #718, opened 2019-05-17
WIP: Dutchx

tibetsprague:
Related issue: #6000

Tabs work, forms mostly in place, doesnt submit correctly, still need to work on whitelisting tokens


src/genericSchemeRegistry/dxDao.json
```diff
@@ -0,0 +1,135 @@
...
+  "actions": [
+    {
+      "id": "updateMasterCopy",
+      "label": "Update MasterCopy",
+      "notes": "https://github.com/gnosis/dx-contracts/blob/master/contracts/base/DxUpgrade.sol#L17",
+      "abi": {
+        "constant": false,
+        "inputs": [
+          {
+            "label": "New Contract Address",
```
jellegerbrandy: sorry we cannot do this. The ABi is a strict format, we need to save the input labels apart from that...

jellegerbrandy: something like `action.fields = { 
   _masterCopy: { label: "the label" }}
}`

src/components/CreateProposal/CreateProposalContainer.tsx
```diff
@@ -50,8 +51,8 @@ class CreateProposalContainer extends React.Component<IProps, null> {
           : schemeName === "SchemeRegistrar" ?
             <CreateSchemeRegistrarProposal daoAvatarAddress={daoAvatarAddress} handleClose={this.goBack.bind(this)} />
           : schemeName === "GenericScheme" ?
-            <CreateGenericSchemeProposal daoAvatarAddress={daoAvatarAddress} handleClose={this.goBack.bind(this)} />
-          : ""
+            <CreateDutchXProposal daoAvatarAddress={daoAvatarAddress} handleClose={this.goBack.bind(this)} />
```
jellegerbrandy: Given that we are tryuing to implement a generic solution here, perhaps we should rename these components and remove any references to DutchX...

A-Zak:
A few comments:
• the arrow should be the same size as the tab :)
image
• placeholders should be a little less abi-ish - [a bunch of small changes](https://docs.google.com/spreadsheets/d/1FKjKj18MjhlOK5GapMguyOfZZ6wtWp9WcKTZu_YtK6w/edit#gid=0)
• Typo in "New auction threshold" tab title... should be "auction", not "action"
• When adding / removing multiple tokens we only need the [+] & [-] buttons once (at the bottom of the list)...
image

-----

Just to make sure, that the title of the popup will be the title of the scheme... and when it'll be called DutchX, it will also show up in the popup title...
image

tibetsprague:
@A-Zak for the title when you say the title of "the scheme",  do you mean "DutchExchange"? because thats the name of the scheme. Or do you want it to be the name of the DAO?

tibetsprague:
@A-Zak also i dont totally agree with the plus and minus for tokens only being at the end, its kinda handy to be able to remove one in the middle. maybe overkill, but we already have it :)

tibetsprague:
also, i really dont think it looks good to have a field's label and placeholder be the same text. its just duplicate. the placeholder should be an example of what a correct contract address looksl ike, for example, or nothing, we may not need placeholders

tibetsprague:
actually @A-Zak i think it just dont know what you mean by New Title in that spreadsheet. i assumed you meant the new label for the field. but also you dont mention the Token Address field

A-Zak:
@tibetsprague 

--------

> @A-Zak for the title when you say the title of "the scheme", do you mean "DutchExchange"? because that's the name of the scheme. Or do you want it to be the name of the DAO?

I mean the "DutchExhcnage" (I thought it would be called DutchX)...

--------

> @A-Zak also i dont totally agree with the plus and minus for tokens only being at the end, its kinda handy to be able to remove one in the middle. maybe overkill, but we already have it :)

I think that in this case, it'll be more distracting than useful... let's remove them please :)

----------

> also, i really dont think it looks good to have a field's label and placeholder be the same text. its just duplicate. the placeholder should be an example of what a correct contract address looksl ike, for example, or nothing, we may not need placeholders

You're right... let's go with no placeholders in this case... When we show this to gnosis, maybe they'll have an idea for a good placeholder... but for now, let's remove them :)

---------

> actually @A-Zak i think it just dont know what you mean by New Title in that spreadsheet. i assumed you meant the new label for the field. but also you dont mention the Token Address field

Yes, I meant the field's label. the "Token address" seemed fine, so I didn't include it... Sorry for the confusion... ^_^

tibetsprague:
@A-Zak actually we can call it whatever we want. i will go with DutchX for now

tibetsprague:
@A-Zak you want the SchemeCard to say DutchX too? so it should never say GenericScheme if we know the specific scheme its pointing to?

tibetsprague:
@A-Zak check out the latest, i made it so you can still remove any token in the list but its much cleaner

A-Zak:
> @A-Zak you want the SchemeCard to say DutchX too? so it should never say GenericScheme if we know the specific scheme its pointing to?

yes, exactly :)

A-Zak:
@tibetsprague 

> @A-Zak check out the latest, i made it so you can still remove any token in the list but its much cleaner

This looks great... we just need to start with the first field open :)

tibetsprague:
ok done. @A-Zak let me know if you are ok with our attempts to fix the tab selected arrow. its now centered. its much tricker to try and get it to be the height of the tab, but if you want that ill get back to work on it :)

dkent600:
@tibetsprague 

> Related issue: #6000

I think you meant 600.

Can we refer to this as follows:

> Resolves: #600 

Then the ticket will be automatically closed.

Also, can we have PR descriptions that describe what functionality changes are being proposed?

PR #719, opened 2019-05-17
improve timeout and error handling when initializing arc

jellegerbrandy:


src/layouts/AppContainer.tsx
```diff
@@ -81,48 +81,52 @@ class AppContainer extends React.Component<IProps, IState> {
...
-      if (currentAddress)  {
-        console.log(`using address from web3 connection: ${currentAddress}`);
-        this.props.cookies.set("currentAddress", currentAddress, { path: "/"});
-        this.props.setCurrentAccount(currentAddress);
-      } else {
-        const currentAddressFromCookie = this.props.cookies.get("currentAddress");
-        if (currentAddressFromCookie) {
-          console.log(`using address from cookie: ${currentAddressFromCookie}`);
-          this.props.setCurrentAccount(currentAddressFromCookie);
+    initializeArc()
```
tibetsprague: hmm, I changed how this works on the multiple-schemes branch. we now initializeArc way sooner in the app load process, so this will  be some merge conflicts...

jellegerbrandy: Ok. This is more about error handlign. And the automatic timeout so that  if something hangs in initalizeArc, we will give up after one sec...

Issue #529, opened 2019-04-02
create a smaller bundle


jellegerbrandy:
* the current bundle size is 11.4MB; this value (and the historical values) can be seen in https://travis-ci.org/daostack/alchemy/jobs/514552706#L4200
*  having less dependencies is low-hanging fruit: https://github.com/daostack/alchemy/issues/313
* we certainly should have a very light landing page and load the Big Bundle in the background
* `npm run analyze` analyzes the bundle and shows a nice pciture

jellegerbrandy:
there are many possible improvements here:

![image](https://user-images.githubusercontent.com/1306173/55396160-868bf600-5543-11e9-952b-d2078a771aa1.png)

tibetsprague:
https://hackernoon.com/optimising-your-application-bundle-size-with-webpack-e85b00bab579
https://hackernoon.com/reduce-webpack-bundle-size-for-production-880bb6b2c72f
https://github.com/GoogleChromeLabs/webpack-libs-optimizations
https://github.com/GoogleChromeLabs/webpack-libs-optimizations

tibetsprague:
- Only load the exports we need in each import statement
- Ignore stuff we don't need from vendor libraries, like locales from moment
- separate vendor chunk
- code splitting by routes?
- GZIP compression
- Minimal source maps, figure out best options for each environment
...

jellegerbrandy:
about the first point: we should enable the `no-unused-imports` in `tslint.json`

PR #728, opened 2019-05-21
Bundle size

leviadam:
Address #529.

Decrease bundle size from 14.88mb to 6.52mb parsed, and to 1.45mb zipped


leviadam:
Note that in the live alchemy, we somehow get 1.9mb, not sure how.
So this might not even be an improvement.

leviadam:
Update:

Traced it. It is not heroku that is doing the magic, it is cloudflare.
So either way we will gain.

PR #736, opened 2019-05-24
Added total reputation per user to reputation view

brodeur:


src/components/Account/ReputationView.scss
```diff
@@ -0,0 +1,4 @@
+.reputationAmount {
+	display: inline-block;
```
tibetsprague: I think for some reason all your .scss files are using tabs instead of 2 spaces, can you check that?

jellegerbrandy:
this stuff is 2 weeks old, what's up with that? @brodeur ?

tibetsprague:
waiting for aaron to fix the test, and he was having trouble getting tests running on his machine...

PR #742, opened 2019-05-28
Show all proposal types in the DAO history page

tibetsprague:


src/components/ViewDao/DaoHistoryContainer.tsx
```diff
@@ -59,16 +59,20 @@ export default (props: {currentAccountAddress: Address} & RouteComponentProps<an
...
   return <Subscribe observable={observable}>{
-    (state: IObservableState<[IProposalState[], IProposalState[], IDAOState]>): any => {
+    (state: IObservableState<[IProposalState[], IProposalState[], IProposalState[], IProposalState[], IProposalState[], IProposalState[], IDAOState]>): any => {
       if (state.isLoading) {
         return (<div className={css.loading}><img src="/assets/images/Icon/Loading-black.svg"/></div>);
       } else if (state.error) {
         return <div>{ state.error.message }</div>;
       } else  {
-        return <DaoHistoryContainer proposals={state.data[0].concat(state.data[1])} dao={state.data[2]} currentAccountAddress={currentAccountAddress}/>;
+        return <DaoHistoryContainer proposals={state.data[0].concat(state.data[1]).concat(state.data[2]).concat(state.data[3]).concat(state.data[4]).concat(state.data[5])} dao={state.data[6]} currentAccountAddress={currentAccountAddress}/>;
```
jellegerbrandy: why do we need to concat here? Would it not make sense to have a single `proposals` query here, instead of getting a separate list for each type? I.e. does state.data[1] not already return all proposals?

src/components/ViewDao/DaoHistoryContainer.tsx
```diff
@@ -20,7 +20,7 @@ class DaoHistoryContainer extends React.Component<IProps, null> {
     const { proposals, dao, currentAccountAddress } = this.props;
 
     const proposalsHTML = proposals.map((proposal: IProposalState) => {
-      return (<ProposalHistoryRow key={"proposal_" + proposal.id} proposalId={proposal.id} dao={dao} currentAccountAddress={currentAccountAddress}/>);
+      return (<ProposalHistoryRow key={"proposal_" + proposal.type + "_" + proposal.id} proposalId={proposal.id} dao={dao} currentAccountAddress={currentAccountAddress}/>);
```
jellegerbrandy: why add the proposal.type to the key? Each proposal shoudl only turn up once in this list, so `proposal.id` should be enough for the key...

src/components/ViewDao/DaoHistoryContainer.tsx
tibetsprague: because of how you setup the search query in the client library https://github.com/daostack/client/issues/230

src/components/ViewDao/DaoHistoryContainer.tsx
tibetsprague: yeah we don't need that

src/components/ViewDao/DaoHistoryContainer.tsx
jellegerbrandy: ah, alright, sorry about that. That is fixed in the new client release (but that one is broken in other ways, don't use it yet)

jellegerbrandy: can you mark these lines with a `TODO` so we don't forget to simplify this when then new client is finally done?

jellegerbrandy: https://github.com/daostack/alchemy/issues/752

PR #763, opened 2019-05-29
Fix account profile saving again

tibetsprague:
We weren’t checking if user had enabled MetaMask. And needed to fix the from param to the .send call

FIxes https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1020/silent


jellegerbrandy:
Ok, so reading the title here, this needs a test :-)

Issue #760, opened 2019-05-29
write tests for redeeming


jellegerbrandy:
This is very important, especially given the alrge amount of error reports we have about this.

jellegerbrandy:
part of a greater testing project, specced in target process

Issue #758, opened 2019-05-29
Cannot redeem proposal if the token address is invalid


pat-daostack:
Trying to redeem this proposal:

https://alchemy.daostack.io/dao/0x294f999356ed03347c7a23bcbcf8d33fa41dc830/proposal/0x1a13d7f098a4379ab1fe96645bb80be1302a90e50521240ebb71c97149862a64

Receiving: "Reward: transaction failed :-(" error message

Console log:

inpage.js:1 MetaMask - RPC Error: gas required exceeds allowance or always failing transaction {code: -32000, message: "gas required exceeds allowance or always failing transaction"}code: -32000message: "gas required exceeds allowance or always failing transaction"__proto__: constructor: ƒ Object()hasOwnProperty: ƒ hasOwnProperty()isPrototypeOf: ƒ isPrototypeOf()propertyIsEnumerable: ƒ propertyIsEnumerable()toLocaleString: ƒ toLocaleString()toString: ƒ toString()valueOf: ƒ valueOf()__defineGetter__: ƒ __defineGetter__()__defineSetter__: ƒ __defineSetter__()__lookupGetter__: ƒ __lookupGetter__()__lookupSetter__: ƒ __lookupSetter__()get __proto__: ƒ __proto__()set __proto__: ƒ __proto__()
(anonymous) @ inpage.js:1
(anonymous) @ inpage.js:1
(anonymous) @ inpage.js:1
u @ inpage.js:1
(anonymous) @ inpage.js:1
We @ inpage.js:1
(anonymous) @ inpage.js:1
value @ inpage.js:1
(anonymous) @ inpage.js:1
n @ inpage.js:1
o @ inpage.js:1
(anonymous) @ inpage.js:1
(anonymous) @ inpage.js:1
n @ inpage.js:1
o @ inpage.js:1
(anonymous) @ inpage.js:1
(anonymous) @ inpage.js:1
(anonymous) @ inpage.js:1
(anonymous) @ inpage.js:1
u @ inpage.js:1
a @ inpage.js:1
(anonymous) @ inpage.js:1
(anonymous) @ inpage.js:1
(anonymous) @ inpage.js:1
r @ bundle-ea484ec6.js:77
setTimeout (async)
(anonymous) @ bundle-ea484ec6.js:77
(anonymous) @ inpage.js:1
write @ inpage.js:1
b @ inpage.js:1
(anonymous) @ inpage.js:1
g.write @ inpage.js:1
y @ inpage.js:1
[...]

Issue #732, opened 2019-05-23
redeem for beneficiary button - wrong beneficiary 


orenyodfat:
expected : 
`redeem for beneficiary` should call redeemer.redeem with  the beneficiary  as the 3rd param. 

found : 
`redeem for beneficiary` call redeemer.redeem with  the account which connect to meta mask  as the beneficiary.

jellegerbrandy:
If this is a bug, it seems to me more of a UX bug :-)

What is intended by "redeem for beneficiary" is that by clicking on that button, the rewards of the contributionreward will be transfered for the benficiary of that ContributionReward.

It is just a label on the button: if there are contributionreards to be transfered, the label will be "redeem for beneificiary", after that it will become simply "redeem".

This behavior corresponds exactly to the semantics of the `redeem` function in `Redeemer.sol`.

I agree though that the label is a bit confusing. This is partly because the term "beneficiary" is ambiguous: it can mean the beneficiary of the protocol rewards, or the beneficiary of the proposal....

jellegerbrandy:
@A-Zak  shall we remove "for beneficiary" from that button, as it is confusing?

Issue #712, opened 2019-05-15
Redemptions page show a large "pending rewards" number I can't seem to redeem


parrachia:
![image](https://user-images.githubusercontent.com/9451597/57808236-4ec6bf80-7739-11e9-82b1-ab14dafe1ce4.png)
When I hover the mouse over the redeem buttons, it shows no gen reward and when I tried to execute them the tx went successfully but nothing changed on Alchy.
Last two attempts etherscan links: [(1)](https://etherscan.io/tx/0x036641534a0fadef6fa68be19677158e677e87b4eec422913e496ec0c112f160), [(2)](https://etherscan.io/tx/0x3a990f04278a81f065aec172db25db541f3888bd90cd5bde9cab1d67428689a6)

parrachia:
@A-Zak asked me to post [my address](https://etherscan.io/address/0x144c4e5027b69f7798b2b162d924bcae5c149f15#tokentxns) and what do I think my rewards should be and why: 
I'm ok with what the UI shows in the image above (about 1k). Because it is proporcional to what I staked and, well, it's what the dapp is showing to me.

jellegerbrandy:
I believe this is because the redemptions list shows proposals that already ahve been redeemed.  In that case, clicking the redeem button does nothing. This should be fixed by #778

parrachia:
Sorry for the delay, it's still happening.

PR #778, opened 2019-06-03
fixing up redemptions container

jellegerbrandy:
redemptions container had a large numer of bugs

FIXED in this PR:

- write `+0.00%` instead of `0.00%` if the value is not zero 
- `+0.00%` is still super uninformative, so I added the complate absolute value in the tooltip
- remove the isredeempending checks which are not working in any cases
- also take into account cases in which rewards were already distributed
- show only proposals that are actually executable (work around an upstream bug which is here https://github.com/daostack/subgraph/issues/252)
- #731 [cannot redeem because proposals already redemed, fixed by line above]
- #712 [wrong totals]  - is the same problem as 731, I think
- show all different GP rewards in redemptions tip
- fix logic of "you vs the beneficiary will receive" in redemptions tip
- remove the redemptionsstring (next to the redeem button) as it is was showing the wrong results and it is unclear what is should show and it would take me 2 hours to fix the convoluted logic
- active redeem button on non-contributionreward proposals (where is was not shown at all)
- fix logic of showing redeem button on proposaldetaisl container
- **fix bug where we were actually never claiming our own rewards, but only the ones for the beneficiary**
- #588 [reward display not updated] [Seems fixed now, cannot reproduce]
- the "redeem for beneificiary" button was showing also on failed proposals. I added ane xtra check, so it does not, but the underlying bug is https://github.com/daostack/subgraph/issues/263
- factor out logic of contributionreward's claimable rewards, and fix some of it

RELATED:
- #760 [write tests]
- #658 [redemption tip]
- #732 (needs input from @A-Zak )
 #703 [redeem your stakes on timed-out proposals] 
- #643 [duplicate of 703]
- #758 [try to distribute from invalid token address, needs input on how to handle it]


src/components/ViewDao/DaoRedemptionsContainer.tsx
```diff
@@ -110,22 +114,29 @@ export default (props: { dao: IDAOState, currentAccountAddress: Address } & Rout
     proposals(where: {
       accountsWithUnclaimedRewards_contains: ["${props.currentAccountAddress}"]
       dao: "${props.dao.address}"
-      contributionReward_not: null
+      stage_in: ["Boosted", "QuietEndingPeriod", "Executed"]
```
tibetsprague: why would we want boosted proposals here?

jellegerbrandy: oops, good point.

Just the "executed" proposals, then

jellegerbrandy: when we upgrade the the new version of the subgraph, this line will be unecessary : https://github.com/daostack/subgraph/issues/252

A-Zak:
A few regression we have with all these changes...
• When creating a proposal I get this error in console:
image

I'm guessing that fixing this will also fix:
• an error displayed on proposal card
image

• Crash on redemptions page (with potential rewards)
address: 0x6d1A4be65723355a9f5b12C446780a5000849461
https://daostack-alchemy-stagin-pr-778.herokuapp.com/dao/0x3cdb39c1ca57c19a316dd2f5acf900705220c713/redemptions/
image

jellegerbrandy:
* fixed the crashing redemptions page

jellegerbrandy:
Ok, I believe the first (2) bugs were fixed by @tibetsprague 's commit, I just fixed the last one (crashing redemptions page)

src/components/Proposal/ActionButton.tsx
```diff
@@ -1,39 +1,37 @@
...
 import { showNotification } from "reducers/notifications";
 import { IProfileState } from "reducers/profilesReducer";
 import { Observable, of} from "rxjs";
-import { isRedeemPending } from "selectors/operations";
-import RedemptionsString from "./RedemptionsString";
-import RedemptionsTip from "./RedemptionsTip";
-
+import { map, mergeMap } from "rxjs/operators";
 import * as css from "./ActionButton.scss";
+/* import RedemptionsString from "./RedemptionsString"; */
```
tibetsprague: remove commented out stuff

src/components/Proposal/ActionButton.tsx
```diff
@@ -144,43 +139,43 @@ class ActionButton extends React.Component<IProps, IState> {
...
             <button className={css.executeButton} onClick={this.handleClickExecute.bind(this)}>
               <img src="/assets/images/Icon/execute.svg"/>
               { /* space after <span> is there on purpose */ }
               <span> Execute</span>
             </button>
           : redeemable ?
             <div>
-              { !detailView ?
-                  <RedemptionsString currentAccountAddress={currentAccountAddress} dao={dao} proposal={proposal} rewards={rewardsForCurrentUser} />
+              {/* !detailView ?
```
tibetsprague: we dont want this? or its just broken for now?

src/components/Proposal/Predictions/PredictionButtons.tsx
```diff
@@ -258,7 +257,7 @@ class PredictionBox extends React.Component<IProps, IState> {
               }
             </span>
             : <span className={css.disabledPredictions}>
-               Predictions are disabled
+                { /* Predictions are disabled */ }
```
tibetsprague: ?

src/components/ViewDao/DaoRedemptionsContainer.tsx
```diff
@@ -36,34 +36,42 @@ class DaoRedemptionsContainer extends React.Component<IProps, null> {
     // TODO: the reward object from the subgraph only gives rewards for voting and staking and dao bounty,
     // the original code also considers ethREward and externalTokenRewards
     // let ethReward = 0
+
+    // TODO: move all the next logic to the client library
+    // calculate the total rewards from the genesisprotocol
     const genReward = new BN("0");
+    const ethReward = new BN("0");
     const reputationReward = new BN(0);
     // , externalTokenReward = 0;
```
tibetsprague: shouldnt we add in external token rewards now too?

src/lib/util.ts
```diff
@@ -199,3 +199,35 @@ export function linkToEtherScan(address: Address) {
...
   return `https://${prefix}etherscan.io/address/${address}`;
 }
+
+export function getClaimableRewards(reward: IRewardState) {
+  if (!reward) {
+    return {};
+  }
+
+  const result: {[key: string]: BN} = {};
+  if (reward.reputationForProposer.gt(new BN(0)) && new BN(reward.reputationForProposerRedeemedAt).isZero()) {
```
tibetsprague: why are you converting the RedeemedAt fields to BigNumbers? arent they dates?

src/components/Proposal/ActionButton.tsx
jellegerbrandy: quoting from the PR description: "remove the redemptionsstring (next to the redeem button) as it is was showing the wrong results and it is unclear what is should show"

I thiknk we are making a mistake showing similar info in different formats in different places in slightly different ways, and we should rethink that approach.

src/components/Proposal/Predictions/PredictionButtons.tsx
jellegerbrandy: that text is superconfusing to me, every time I see it. I created an issue and will put it back for now

src/components/ViewDao/DaoRedemptionsContainer.tsx
jellegerbrandy: perhaps, I don't know, there could be 100 different tokens, so it's a whole new ballgame.

src/lib/util.ts
jellegerbrandy: because of a client bug. Now fixed.

PR #790, opened 2019-06-07
[WIP] 1041 upstakes to preboost

tibetsprague:
Show GEN to preboost amounts and preboost button correctly
Only when over the threshold, not at the threshold

Use > symbols instead of + symbols to indicate token amounts that are more than the number shown

### DIFF
--- a/src/components/Proposal/ActionButton.tsx
+++ b/src/components/Proposal/ActionButton.tsx
@@ -153,7 +153,7 @@ class ActionButton extends React.Component<IProps, IState> {
           /> : ""
         }
 
-        { proposal.stage === IProposalStage.Queued && proposal.upstakeNeededToPreBoost.lte(new BN(0)) ?
+        { proposal.stage === IProposalStage.Queued && proposal.upstakeNeededToPreBoost.lt(new BN(0)) ?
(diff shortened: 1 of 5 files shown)


A-Zak:
@tibetsprague why did you change the '+0'  to a '>0' ?
It works for the "X gen to boost" sentence, but it doesn't make sense when just showing a token value...

tibetsprague:
@A-Zak it does make sense, the number of >0, +0 doesn't make sense to me. 0+ makes more sense, but why not use the same symbol everywhere? that would be clearer

jellegerbrandy:
(why 1041?)

A-Zak:
@jellegerbrandy 
> (why 1041?)

ticket number 1041 in TP
https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/4645906706962480881&appConfig=eyJhY2lkIjoiMEFGRDcyNjIyOTgzRjFFMkMwRjIwMDBCNzA0QjRGMzQifQ==&searchPopup=bug/1041

tibetsprague:
i still think > makes more sense throughout, but ok, ill do what you want :)

PR #803, opened 2019-06-12
Mobile v0

brodeur:
Alright! Here's my first attempt at mobile. A lot of this is without design guidance, so please take a look with your eyes to make sure things look good and feel good. Would love any feedback for improvement.


src/components/Account/Account.scss
```diff
@@ -399,4 +399,75 @@ input {
...
+    margin-top: 15px;
+  }
+
+  .socialProof {
+    left: 70px;
+    opacity: .6;
+    width: 77%;
+  }
+
+  :global .membersPage {
```
tibetsprague: why do you need the global here?

src/components/Proposal/ProposalDetails.scss
```diff
@@ -170,27 +170,35 @@ button.disabled {
...
-		font-weight: bold;
-	}
+    opacity: .6;
+    color: rgba(104, 155, 214, 1.000);
+    font-weight: bold;
+  }
 }
 
 .voteButtonsBottom {
-	// text-align: left;
```
tibetsprague: weird these are here, lets remove comments

PR #807, opened 2019-06-13
Redemptions menu item should display feedback when user has unclaimed redemptions

dkent600:
Redemptions menu item should display feedback when user has unclaimed redemptions

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "build-travis": "cross-env NODE_ENV=production node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js --config webpack.prod.config.js",
     "heroku-postbuild": "npm run build",
     "lint": "tslint --project .",
+    "lint.andFix": "tslint --project . --fix",
     "service-status": "node scripts/serviceStatus.js",
     "start": "webpack-dev-server --config webpack.dev.config.js",
     "start-dev-profiling": "cross-env NODE_ENV=dev-profiling API_URL=https://daostack-alchemy.herokuapp.com S3_BUCKET=daostack-alchemy-staging webpack-dev-server --config webpack.dev.config.js",
--- a/src/components/Account/AccountProfileContainer.tsx
+++ b/src/components/Account/AccountProfileContainer.tsx
@@ -91,7 +91,7 @@ class AccountProfileContainer extends React.Component<IProps, null> {
     e.preventDefault();
   }
 
-  public async handleSubmit(values: FormValues, { props, setSubmitting, setErrors }: any ) {
+  public async handleSubmit(values: FormValues, { props, setSubmitting, setErrors }: any) {
     const { accountAddress, currentAccountAddress, showNotification, updateProfile } = this.props;
 
     if (!(await checkMetaMaskAndWarn(this.props.showNotification.bind(this)))) { return; }
@@ -140,14 +140,14 @@ class AccountProfileContainer extends React.Component<IProps, null> {
     return (
       <div className={css.profileWrapper}>
         <BreadcrumbsItem to={`/profile/${accountAddress}`}>
-          { editing ? (accountProfile && accountProfile.name ? "Edit Profile" : "Set Profile") : "View Profile"}
+          {editing ? (accountProfile && accountProfile.name ? "Edit Profile" : "Set Profile") : "View Profile"}
         </BreadcrumbsItem>
 
-        { dao ? <DaoSidebar address={dao.address} /> : ""}
+        {dao ? <DaoSidebar {...this.props} dao={dao} /> : ""}
 
         <div className={css.profileContainer} data-test-id="profile-container">
-          { editing && (!accountProfile || !accountProfile.name) ? <div>In order to evoke a sense of trust and reduce risk of scams, we invite you to create a user profile which will be associated with your current Ethereum address.<br/><br/></div> : ""}
-          { typeof(accountProfile) === "undefined" ? "Loading..." :
+          {editing && (!accountProfile || !accountProfile.name) ? <div>In order to evoke a sense of trust and reduce risk of scams, we invite you to create a user profile which will be associated with your current Ethereum address.<br /><br /></div> : ""}
+          {typeof (accountProfile) === "undefined" ? "Loading..." :
             <Formik
               enableReinitialize={true}
               initialValues={{
@@ -189,7 +189,7 @@ class AccountProfileContainer extends React.Component<IProps, null> {
                         <label htmlFor="nameInput">
                           Name:&nbsp;
                         </label>
-                        { editing ?
+                        {editing ?
                           <div>
                             <Field
                               autoFocus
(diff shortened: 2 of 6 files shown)


dkent600:
@jellegerbrandy Could be nice if the proposal count could come from the client, instead of having to query for all the proposals having redemptions.

src/components/Account/AccountProfileContainer.tsx
```diff
@@ -252,15 +252,15 @@ class AccountProfileContainer extends React.Component<IProps, null> {
...
                       <div className={css.tokens}>
                         {accountInfo
-                           ? <div><strong>Rep. Score</strong><br/><ReputationView reputation={accountInfo.reputation} totalReputation={dao.reputationTotalSupply} daoName={dao.name}/> </div>
-                           : ""}
-                         <div><strong>GEN:</strong><br/><span>{formatTokens(genBalance)}</span></div>
--                        <div><strong>ETH:</strong><br/><span>{formatTokens(ethBalance)}</span></div>
+                          ? <div><strong>Rep. Score</strong><br /><ReputationView reputation={accountInfo.reputation} totalReputation={dao.reputationTotalSupply} daoName={dao.name} /> </div>
+                          : ""}
+                        <div><strong>GEN:</strong><br /><span>{formatTokens(genBalance)}</span></div>
+                        -                        <div><strong>ETH:</strong><br /><span>{formatTokens(ethBalance)}</span></div>
```
tibetsprague: some weird whitespace here

src/components/ViewDao/DaoSidebar.tsx
```diff
@@ -29,50 +32,54 @@ class DaoSidebarComponent extends React.Component<IProps, null> {
...
-              : <p>Anyone can make a proposal to the DAO! Click the button on the top right.</p>
+                : <p>Anyone can make a proposal to the DAO! Click the button on the top right.</p>
             }
           </div>
           <div className={css.navigation}>
             <span className={css.navHeading}><b>Menu</b></span>
             <ul>
               <li>
                 <Link to={"/dao/" + dao.address}>
+                  <span className={css.menu_dot} />
```
tibetsprague: we use camelCase classes

src/components/ViewDao/DaoSidebar.tsx
```diff
@@ -29,50 +32,54 @@ class DaoSidebarComponent extends React.Component<IProps, null> {
...
+                <NavLink activeClassName={css.selected} to={"/dao/" + dao.address + "/history/"}>
+                  <span className={css.menu_dot} />
+                  <span className={css.notification + " " + css.historyNotification}></span>
+                  <img src="/assets/images/Icon/menu/history.svg" />
+                  History
                   </NavLink>
               </li>
               <li>
                 <NavLink activeClassName={css.selected} to={"/dao/" + dao.address + "/redemptions/"}>
+                  <span className={css.menu_dot + (proposalCount ? " " + css.red : "")} />
```
tibetsprague: lets use the classNames library to make this cleaner, we use it all across the app to handle logic for classes on elements

src/components/ViewDao/ViewDaoContainer.tsx
```diff
@@ -85,13 +85,13 @@ class ViewDaoContainer extends React.Component<IProps, IState> {
     return (
       <div className={css.outer}>
         <BreadcrumbsItem to={"/dao/" + dao.address}>{dao.name}</BreadcrumbsItem>
-        <DaoSidebar address={dao.address} />
+        <DaoSidebar {...this.props} dao={dao} />
```
tibetsprague: i doubt you want to send all the ViewContainer's props to the DaoSidebar

tibetsprague: you just need the dao right?

src/components/Account/AccountProfileContainer.tsx
dkent600: It seems that my editor adds those.  I'm not sure it is from the linting rules or what.  But those spaces are not a problem.

src/components/ViewDao/DaoSidebar.tsx
dkent600: fixed

src/components/ViewDao/DaoSidebar.tsx
dkent600: I'm not sure what you're asking for here, since I am using className

src/components/ViewDao/ViewDaoContainer.tsx
dkent600: also need `currentAccountAddress`.  Should I break that out instead of sending `props`?

src/components/Account/AccountProfileContainer.tsx
dkent600: This appears to explain reasoning behind adding the space:  https://stackoverflow.com/questions/462741/space-before-closing-slash

src/components/ViewDao/DaoSidebar.tsx
tibetsprague: search the code for classNames (it's plural), its a library that makes it much easier to use logic to determine which classes should be on an element. we use it in almost every component

src/components/ViewDao/ViewDaoContainer.tsx
tibetsprague: I guess either way is fine, though in this case with just one prop needed its nicer to spell out what's happening

src/components/ViewDao/DaoSidebar.tsx
dkent600: Gotcha.  It seems that in that module there is a bunch of code unrelated to mine which is also not using classNames.  I went ahead and refactored it all to use classNames.

src/components/ViewDao/ViewDaoContainer.tsx
dkent600: I agree, done.

package.json
```diff
@@ -52,6 +52,7 @@
     "build-travis": "cross-env NODE_ENV=production node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js --config webpack.prod.config.js",
     "heroku-postbuild": "npm run build",
     "lint": "tslint --project .",
+    "lint.andFix": "tslint --project . --fix",
```
jellegerbrandy: not sure if this `npm run lint.andFix` is easier to remember then `npm run lint -- --fix` :-)

src/components/Account/AccountProfileContainer.tsx
jellegerbrandy: the whitespace linting rule itself is not a problem either way, but you PR becomes hard to read if you use different linting rules (and "fix" them) then we do, which _is_ important to us :-) So if you can get your linter in agreement;  or do not commit these type of fixes, life would be better :-)

package.json
dkent600: learning the "-- --param" syntax is far from the first thing I learned/memorized about running NPM NPM scripts.

dkent600:
@jellegerbrandy 

> I do not want to be ungrateful, but this PR changes so many lines it is almost impossible to see what you have actually changed.. 

It is easy to turn off whitespace diffs here in github.  Click on "Files Changed" tab, select "Hide whitespace changes"

A-Zak:
@dkent600
It’s better if everyone use the same linting rules, also it’s bothering everyone’s ocd...

If you really believe it’s better, try to convince everyone to agree to make this change...

Until they agree, please try to fix your linter.

PR #813, opened 2019-06-18
Combine unsupported schemes

dkent600:
Combine all unsupported schemes into a single card on the allschemes page.  Show scheme name when available, the address if not.  Either way, should be hot, directing the user to etherscan.

We'll show the sceheme registration date when it becomes possible in the subgraph and client.


src/components/ViewDao/UnknownSchemeCard.scss
```diff
@@ -0,0 +1,88 @@
+.wrapper {
+	background: white;
```
tibetsprague: weird indentation here, tabs?

dkent600: It seems that the tabs predate my changes:  They came from SchemeCard.scss.

dkent600: So I went ahead and removed the tabs from this file, but will not do the same for other files.

PR #818, opened 2019-06-19
don't show the EnablePredicting button

dkent600:
When the user isn't connected to the network, don't show the EnablePredicting button, show the staking yes/no buttons.  Don't show the "insufficient GEN" tooltip.  When the user clicks, connect them to the network.

(Also:  cleaned up the layout of the "Enable Predicting" button.)


A-Zak:
jhgjhf

PR #824, opened 2019-06-20
New proposal status feedback

dkent600:
Provide feedback on proposal details page, next to the action button, showing the status of the proposal:

- expired
- passed
- failed
- boosted
- pending
- regular


dkent600:
@brodeur I had a lot of merge conflicts with the mobile scss.  This PR adds a new element that would also need to look good on mobile.  Maybe we can take a look?

leviadam:
I think it is better to merge and later fix the mobile view.
If you can add a ticket for that @brodeur @A-Zak

PR #837, opened 2019-06-26
Make sure that current account address is always in lower case throug…

tibetsprague:
…hout the app

This fixes issue I was seeing with profile not loading correctly, and potentially unknown bugs


tibetsprague:
@jellegerbrandy do you see any potential issues with this?

PR #842, opened 2019-06-26
use local storage instead of cookies

dkent600:
Resubmitting.  I've requested @tibetsprague as reviewer, though he already approved the original PR, to which this is identical.

Resolves: https://daostack.tpondemand.com/entity/1147-use-local-storage-instead-of-cookies


dkent600:
@jellegerbrandy 

> This just changes a standard interface into a non-standard one, we are better of without.

It is actually more than just a simple one-to-one API wrapper.  But I'll go ahead and remove it anyway.

dkent600:
@jellegerbrandy 

> Can we please remove the "wrapper"

Done

PR #857, opened 2019-07-02
Load more buttons

jellegerbrandy:
this adds: 

* load more buttons to the history and members page
* members are sorted by reputation desceding now
* history is sorted by expiry


tibetsprague:
replaced by https://github.com/daostack/alchemy/pull/862

PR #862, opened 2019-07-03
Infinite scrolling

tibetsprague:
Add partial loading of items with scrolling to load more items for the DAO History, Reputation Holders, and Scheme Queued Proposal lists.

One known issue is that if you scroll to the bottom of the scheme proposals list and then add a new proposal that proposal will never load because it thinks its done loading


jellegerbrandy:
just a heads up that we are working on a more workable system for creating our test set (so we do not have to depend on oren/ben), and we'll probably move the proposal-generation code there...

Perhaps we can test this for now with "mini-pages" in development (show 3 items per page or something) and try it out with the rinkebydata using `npm run start-staging`, and postpone the proposals generation until then.

jellegerbrandy:
@tibetsprague  so the functionality regarding infinite scrolling there for the members and history page, in the form of a working "load more" button, https://github.com/daostack/alchemy/pull/857

The only thing that needs to happen is to replace the button with the correct DOM event.

It makes more sense to continue there; and forget about having a decent test environment until we have sorted out the test-env part https://github.com/daostack/test_env)

leviadam:
Thanks @tibetsprague and @jellegerbrandy for the great work.
There is one thing that might be worth fixing, there is no scroll bar in the reputation holders for example.

tibetsprague:
@jellegerbrandy i dont understand how this fixes the issue i brought up about new proposals not appearing immediately. did you try this (and the tests) in a DAO with more then 2 open proposals?

jellegerbrandy:
@tibetsprague can you be more precise about the problem that you are seeing?

PR #872, opened 2019-07-09
Add support for Gnosis Safe extension

dkent600:
Add support for Gnosis Safe extension

### DIFF
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -41,11 +41,11 @@ const settings = {
  *    it will use `alert()` if no such function is provided
  * @return the web3 connection, if everything is fine
  */
-export async function checkMetaMaskAndWarn(showNotification?: any): Promise<boolean> {
+export async function checkWeb3ProviderAndWarn(showNotification?: any): Promise<boolean> {
   try {
-    const metamask = await checkWeb3Provider();
-    await metamask.enable();
-    return metamask;
+    const web3Provider = await checkWeb3Provider();
+    await web3Provider.enable();
+    return web3Provider;
   } catch (err) {
     const msg =  err.message;
     if (msg.match(/enable metamask/i) && process.env.NODE_ENV === "development") {
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -43,7 +43,7 @@ const settings = {
  */
 export async function checkMetaMaskAndWarn(showNotification?: any): Promise<boolean> {
   try {
-    const metamask = await checkMetaMask();
+    const metamask = await checkWeb3Provider();
     await metamask.enable();
     return metamask;
   } catch (err) {
@@ -66,7 +66,7 @@ export async function checkMetaMaskAndWarn(showNotification?: any): Promise<bool
  * throws an Error if something is wrong, returns the web3 connection if that is ok
  * @return
  */
-export async function checkMetaMask() {
+export async function checkWeb3Provider() {
   let expectedNetworkName;
   switch (process.env.NODE_ENV) {
     case "development": {
@@ -89,7 +89,7 @@ export async function checkMetaMask() {
   let web3Provider = getMetaMask();
 
   if (!web3Provider) {
-    const msg = `Please install or enable metamask`;
+    const msg = `Please install or enable Metamask or Gnosis Safe`;
     throw Error(msg);
   }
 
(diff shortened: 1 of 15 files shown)


src/arc.ts
```diff
@@ -201,7 +190,7 @@ export async function initializeArc(): Promise<Arc> {
   console.log(arcSettings);
   console.log(`alchemy-server (process.env.API_URL): ${process.env.API_URL}`);
   if (arcSettings.web3Provider.isMetaMask) {
-    console.log("Using Metamask Web3 provider");
+    console.log(`Using ${arcSettings.web3Provider.isSafe ? "Gnosis Safe" : "Metamask"} Web3 provider`);
```
jellegerbrandy: so the provider of Gnosis Safe returns `isMetamask` as true?

src/lib/util.ts
```diff
@@ -50,8 +52,32 @@ export default class Util {
   public static getWeb3() {
     return getArc().web3;
   }
-  public static async getNetworkId() {
-    return await getArc().web3.eth.net.getId();
+
+  /**
+   * return network id, independent of the presence of Arc
```
jellegerbrandy: this is a good thing :-)

src/lib/util.ts
```diff
@@ -50,8 +52,32 @@ export default class Util {
...
+      arc = getArc();
+    } catch {}
+
+    if (arc) {
+      web3 = arc.web3;
+    } else if ((<any> window).web3) {
+      web3 = (<any> window).web3;
+    } else if (web3Provider) {
+      web3 = new Web3(web3Provider);
+    }
```
jellegerbrandy: So this code will try to find the web3provider on the window object by metamask or gnosis safe.

The `getMetamask`  function which is used to find the web3provider to pass to Arc does the same thing (it looks at the `window.ethereum` object). It is probably better to use the same function also here, so we are sure that we pass the same web3 provider to Arc as we are using here to check the id.

src/lib/util.ts
```diff
@@ -50,8 +52,32 @@ export default class Util {
...
+
+    if (!web3) {
+      throw new Error("getNetworkId: unable to find web3");
+    }
+
+    return (await (web3.eth.net ? web3.eth.net.getId() : promisify(web3.version.getNetwork)())).toString();
```
jellegerbrandy: O don't think promsify is needed here, as the web3 functions should return promises (if you do not provide a callback)

src/arc.ts
dkent600: Yes, it returns both flags set to true

src/lib/util.ts
dkent600: pre-1.0 we need promisify

src/lib/util.ts
dkent600: That code isn't actually looking for web3Provider here.  web3provider  is optionally passed-in.  The code is looking for window.web3.

src/lib/util.ts
dkent600: That is, one has to provide a callback or else in some cases (I forget now which) these functions will crash because the don't support async operations.

dkent600: https://ethereum.stackexchange.com/questions/47366/the-metamask-web3-object-does-not-support-synchronous-methods-like-personal-newa

src/lib/util.ts
jellegerbrandy: what I mean is that what we are interested in here is to get info about the network that we are connected to (which is what the web3 provider does).

The `getMetamask` function, that is used for the same purpose - find the injected web3 instance, get information about the provider, and pass that on to Arc.

The logic of the two methods is slightly different (get the window.ethereum object vs the window.web3 object), and it should be the same.

src/lib/util.ts
jellegerbrandy: true, this is the injected old web3... bleah.

src/lib/util.ts
dkent600: ... and the provider that is passed-in is the same one that we'll send to Arc

dkent600: OK I simplified the use of getMetaMask a bit, and added the code from above to firm up getNetworkId

jellegerbrandy:
Ok, here are some issues that i am working to resolve:



* [ ] [Chrome with MM disabled and Gnosis Safe enabled, known account] submit button on profile page throws an error (this is bc it tries to sign with the default readonly connection, I think, and not with gnosis safe) 
* [ ] [chrome in anynimous mode, MM disabled, Safe enabled] conolse logs error "please install or enable MM", but this is not shown to user

jellegerbrandy:
So message signing (which we need for alchemy server) seems to be a real problem: https://github.com/gnosis/safe/issues/134#issuecomment-476403989

jellegerbrandy:
It appears that Gnosis safe does not support personal_sign: https://github.com/gnosis/safe-browser-extension/issues/81

Issue #858, opened 2019-07-02
The Great Alchemy Consolidation Issue


jellegerbrandy:
# Refactoring

## URL schemes

| Currently    | Let's change it to              |
|--------------|--------------|
|`/dao/${daoADdresss}/proposals/${schemeId}` |/dao/${daoADdresss}/scheme/${schemeId}

- [x] this is done in #920 

## Global state handling

- [x]  clean up the redux state (e.g. remove dead (unused parts, like `arc` and `ui`, from the state, remove dead functions in `reducers`  and `selectors`, etc)

- [x] [perhaps...] consider not using redux at all, but the (easier, lighter) https://reactjs.org/docs/context.html

## Clean up comments

* [x] remove all "remind me of stuff" code and create issues. working on this in #924

## Dead code

- [x] remove code  that is obviously dead (like `operations.ts`)
- [ ] Use some test coverage tool to see which code is (and is not) covered by our tests
- [x] #450 (mintgen script)

## Linting

- [x] Ticket is [here](https://github.com/daostack/alchemy/issues/915).  We are still working with a large set of disabled rules in `tslint.json`. Some useful ones that should really be activated are `no-console`, `no-var-requires`, `no-shadowed-variable`. Others that are just plain common sense are `max-line-length`, `no-unused-expression`, 

## Dependencies

- [x] Review app depdencies and see what we can do without (#313)
- [x] upgrade dependencies where possible #917
- [ ] try to avoid having including different versions of the same package

## DutchX-specific code

-  [x] should be removed (apart from the DuthX.json file, from whcih all should be generated). This specifically concerns the `ProposalSummaryDutchX.ts` file, which shoudl disappear.

## Directory structure

| Currently    | Let's change it to              |
|--------------|--------------|
| /DaoList/ | Daos
| /CreateProposal/ | /Proposal/Create/
| /ViewDao/ | /Dao
| /Proposal/Predictions | /Proposal/Staking
| - | `Scheme/` a new directory for the scheme stuff (like the `SchemeProposalsContainer`)

## Component naming

We use `View` and `Container` suffixes more or less at random, let's just drop those:

| Currently    | Let's change it to              |
|--------------|--------------|
|  ReputationView | Reputation
| ViewDao | Dao
| DaoHistoryContainer | History
| DaoRedemptionsContainer | Redemptions
| DaoMembersContainer | Members
| SchemeProposalsContainer |  Scheme | 
| AllSchemesContainer | DAOHome |

## Modals

[...]

PR #875, opened 2019-07-09
Fix bug with decimal external token rewards being turned into integers

tibetsprague:
Also do a nice refactor by removing the Util class and making all the util functions individual exports


tibetsprague:
fixes Util part of https://github.com/daostack/alchemy/issues/858

PR #878, opened 2019-07-10
1154 new proposal card

brodeur:
https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=feature/1154/silent


jellegerbrandy:
@brodeur can you fix the test? thanks :-)

src/components/Proposal/Predictions/PredictionButtons.tsx
```diff
@@ -227,46 +230,95 @@ class PredictionBox extends React.Component<IProps, IState> {
...
+              </span>
+            }
+          </div>
         </div>
-      </div>
-    );
+      );
+    } else {
+      return (
+        <div className={wrapperClass}>
```
tibetsprague: this looks very duplicated from above, do we really need to duplicate all this code? what actually changes for the "context menu"?

src/components/Proposal/ProposalCard.scss
```diff
@@ -43,7 +43,7 @@
   display: inline-block;
   position: relative;
   box-shadow: 0 0 15px 0 rgba(133,168,208,.15);
-  overflow: hidden;
+  // overflow: hidden;
```
tibetsprague: remove if we dont want

src/components/Proposal/Voting/VoteButtons.tsx
```diff
@@ -120,60 +124,139 @@ class VoteButtons extends React.Component<IContainerProps, IState> {
...
+              </span>
+              : " "
+            }
+          </div>
         </div>
+      );
+    } else {
+      return (
+        <div className={wrapperClass} >
+          {this.state.showPreVoteModal ?
```
tibetsprague: again id like to figure out what doesnt need to be duplicated, like this probably doesnt

src/components/Proposal/Predictions/PredictionButtons.tsx
brodeur: The proposal card for mobile still uses the other way of voting.. Not sure a more elegant way to do it. Perhaps a mobile design for the new cards will allow us to do away with the duplication? @A-Zak

src/components/Proposal/Voting/VoteButtons.tsx
brodeur: Happy to look at it together... perhaps they should be two separate components?

PR #900, opened 2019-07-12
add descriptions to each proposal type

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=userstory/1248/silent

and:

https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=userstory/1249/silent

**Note** this uses the package [Interweave](https://github.com/milesj/interweave) to safely render the anchor tags in the GenericScheme descriptions.


package.json
```diff
@@ -86,11 +86,12 @@
     "heroku-client": "^3.0.7",
     "history": "4.7.2",
     "immutability-helper": "2.6.4",
+    "interweave": "^11.1.0",
     "moment": "^2.23.0",
     "normalizr": "^3.3.0",
     "query-string": "^6.2.0",
     "rc-tooltip": "^3.7.3",
-    "react": "^16.7.0",
+    "react": "^16.8.6",
```
tibetsprague: hmm, this seems potentially dangerous, did you make sure to look at any breaking changes in the new version? i guess we'll just have to test!

dkent600: I didn't do that inentionally, it just happened.  Perhaps we should lock the version so it unless upgrades intentionally?  Frankly I'd rather back that out, so I think I will, and I will lock it.

dkent600: @tibetsprague 

Storybook wants a newer version:

+-- @storybook/addon-info@5.1.9
| +-- @storybook/addons@5.1.9
| | `-- @storybook/api@5.1.9
| |   `-- react@16.8.6
| +-- @storybook/components@5.1.9
| | `-- react@16.8.6
| `-- react@16.8.6
+-- @storybook/react@5.1.9
| `-- @storybook/core@5.1.9
|   `-- @storybook/ui@5.1.9
|     `-- react@16.8.6
`-- react@16.7.0

jellegerbrandy: I've added an item to the Great Refactoring Issue that we should look into  de-duplicating package dependencies. But in this PR, let's not upgrade react itself yet...

jellegerbrandy: This line  specified a minimal version (that's what the "^" mean). We should, in this PR, not update that minimal version, and also not lock it to a specific version (by removing the circonflex) , which leads to its own problems (these requirement conflicst which  will inflate our bundle size).

I'll reset this to how it originallly was - playing with react versions has no place in this PR.

 The react version that is actually used is in `package-lock.json`. So tibet, if you are are worried about that, you should look there, not in the package.json file.

jellegerbrandy: Just be be more precise, in commit fb78e25c78853b0e4a24bafef13dd1c7b5872eb9

- we are bundling react version 16.7.0 with our app
- storybook uses 16.8.5
- storybook is not part of our bundle, so these "double versions" do not do any harm, afaics

PR #917, opened 2019-07-15
Dependencies

leviadam:
I have updated most of the dependecies in `package.json`.

Tried to update `css-loader` and `immutability-helper` and ran into some errors.
Also tried to update the whole react/redux packages, and got errors, I think this requires much more attention.

This can either be merged, or you can push more updates commits on top.


leviadam:
Added WIP since the test are failing.

PR #919, opened 2019-07-16
Make sure no votes appear in the vote graph

tibetsprague:


tibetsprague:
@dkent600 with accent-1 no color was appearing at all, so no votes were visible

PR #927, opened 2019-07-16
WIP lint for typescript via eslint

dkent600:
**Notes**

- this removes some packages for which I see no use.
- using lots of "recommended" rules presets
- some rules are going to be just warnings (these are defined at the top of the rules list).  The rest I plan to convert to errors as they are fixed
- There are a few old rules that don't seem to be available:

```
object-literal-sort-keys
no-string-literal
jsx-self-close
jsx-alignment
jsx-wrap-multiline
jsx-no-lambda
jsx-no-multiline-js
jsx-no-string-ref
```

The jsx ones are from tslint-react.  the other two are from tslint.  (could be the latter are in eslint, but I can't find them)

- I came across this project that has a "recommended" set of rules for tsx in react web apps, but I didn't pull it in.  It had several dependencies that put me off.  But we could still consider it:  https://www.npmjs.com/package/eslint-config-react-app


.eslintrc
```diff
@@ -1,14 +1,49 @@
...
+        "version": "detect", // React version. "detect" automatically picks the version you have installed.
+      }
+    },
     "rules": {
-        "semi": 2,
-        "eol-last": 2,
-        "eqeqeq": "error",
-        "no-console": 0
+        "max-len": ["warn", 120, 2],
+        "indent": "off",
```
jellegerbrandy: what does that (not) do? We'd like consistent indents...

.eslintrc
```diff
@@ -1,14 +1,49 @@
...
+        "@typescript-eslint/indent": ["error", 2],
+        "@typescript-eslint/prefer-for-of": "warn",
```
jellegerbrandy: why "warn"? Do we not want to enforce these rules?

.eslintrc
dkent600: that disables the eslint rule so that the following @typescript rule can work unimpeaded.

.eslintrc
dkent600: will gradually turn these all into "error"

dkent600:
Useful links:

[tslint rules](https://palantir.github.io/tslint/rules/)
[@typescript-eslint rules](https://github.com/typescript-eslint/typescript-eslint/tree/master/packages/eslint-plugin) (scroll down)
[eslint rules](https://eslint.org/docs/2.0.0/rules/) (scroll down)

dkent600:
Currently no errors, 1616 warnings.

I propose merging this while leaving the PR open for later merging fixes to the warnings

PR #931, opened 2019-07-17
lint for typescript via eslint -- II

dkent600:
- Revert eslint to the version required by @typescript-eslint
- pull in - [eslint-plugin-import](https://www.npmjs.com/package/eslint-plugin-import) to autofix import sorting
- fixed all the import sorting errors
- fixed a bunch of other errors
- honed .eslintrc

There remain 0 errors, 839 warnings


dkent600:
@jellegerbrandy 

> Not sure in all cases about the underscore solution - often it seems cleaner to remove the argument altogether.

I did consider that but took the conservative route.

PR #942, opened 2019-07-18
Refactor components

tibetsprague:
- CreateProposal components to Proposal/Create/
- Prediction components renamed to Stake/Staking
- Use XXXPage for components that represent a distinct URL the user can browse to
- Get rid of use of “Container” except for two components that “contain” many pages: AppContainer and DaoContainer
- ViewDao -> Dao
- DaoList -> Daos
- ReputationView -> Reputation


leviadam:
I like the new naming, thanks @tibetsprague.
What do you think @jellegerbrandy ?

PR #947, opened 2019-07-19
remove unused operations+notifications code, fix lint warnings

dkent600:


dkent600:
@tibetsprague 

> hmm, cant we just remove all the operations code? I thought we weren't using it at all anymore

First, there are two modules named "operations.ts". To which are you referring?

The first, src\reducers\operations.ts, contains operationsReducer which is used in src\reducers\index.ts.  The `operations` reducer appears to be in use in the other operations module, src\selectors\operations.ts.

This other operations module contains two functions that are used: `isVotePending` and `isStakePending`, in src\components\Proposal\Predictions\PredictionButtons.tsx

tibetsprague:
Sounds like I should take on this task since I understand it much more :)

dkent600:
@tibetsprague are you saying that isVotePending should be refactored to use something else besides operations? If so, what? ... that the operations _is_ in use, as I have noted, but that in fact the idea is that we want to stop using it and use something else? (difference between "no longer using" and "want to stop using in favor of something else"?)

PR #952, opened 2019-07-20
tsPrune results

dkent600:
a few small things revealed by txPrune.  Some of them involve simply removing unneeded `export` .

webpack tree-shaking is giving no additional satisfaction

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -28,12 +28,6 @@ export function copyToClipboard(value: any) {
   document.body.removeChild(el);
 }
 
-export function trace<T>(x: T, ...args: any[]): T {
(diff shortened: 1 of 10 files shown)


src/lib/util.ts
```diff
@@ -28,12 +28,6 @@ export function copyToClipboard(value: any) {
   document.body.removeChild(el);
 }
 
-export function trace<T>(x: T, ...args: any[]): T {
-  // tslint:disable-next-line:no-console
-  console.debug("trace", ...args, x);
```
tibetsprague: I use trace sometimes!

dkent600: ok I'll restore that function in another PR

PR #958, opened 2019-07-22
WIP: remove references to DutchX

dkent600:
Get rid of references to DutchX including ProposalSummaryDutchX .

Resolves: #959 

Includes:
- Cleanup the JSON a bit
- Handle non-DutchX DAOs that use GenericScheme.  Does not rely on the DAO name to identify the DAO.
- Handle DAOS that can have more than one GenericScheme
- Avoids redundancy of form specs when multiple DAOs are using the same GenericScheme
- no longer does `new Web3()` to encode/decode (now uses existing)
- fixes what I assume is a bug, which is in production, where the dxDAO Generic Scheme is being called "Generic Scheme" instead of "DutchX"

**Notes** 

- It works with the Generic Scheme in main and rinkeby.  I haven't tested with other DAOs nor DAOs with multiple generic schemes.
- the json now requires the DAO avatar address.  This is because the DAO name may not be unique, and it may eventually be able to change.  Identifying the DAO enables, for a given target contract, that DAOs can define their own desired subset of actions to present in the UI  (So for example: some might want just voteInDAO, others might want just stakeInDAO).
- the json is split in two: the DAOs spec and the Forms spec.
- a side-effect is that I made the app, when in Development mode and MM is available, use whatever network MM is pointing at instead of insisting on localhost.  This helped me test the code.


src/components/Proposal/ProposalSummary/ProposalSummaryKnownGenericScheme.tsx
```diff
@@ -22,47 +23,143 @@ export default class ProposalSummary extends React.Component<IProps> {
...
-  public render() {
+  public render(): any {
     const { proposal, detailView, transactionModal, genericSchemeInfo } = this.props;
-    if (genericSchemeInfo.specs.name === "DutchX") {
-      return <ProposalSummaryDutchX {...this.props} />;
-    }
+
+    return this.renderFormSummary(genericSchemeInfo, proposal, detailView, transactionModal);
+    
+    // const proposalSummaryClass = classNames({
```
dkent600: @jellegerbrandy I don't understand what this code is for, whether it is ever executed, so I commented it out.  Suggestions?

src/genericSchemeRegistry/schemes/DaoGenericSchemesForms.json
```diff
@@ -0,0 +1,31 @@
+{
+  "daos": [
+    {
+      "name": "DutchX",
```
jellegerbrandy: Ok, I am not a fan of this change.

So the purpose of tis form generation stuff is that is (1) to make it as easy as possible to add scheme interfaces, ideally without writing code (2) make it modular: it should be easy to add and remove particular interfaces, and (in the far (?) future) bundle them separately 

The "one-file-per-genericScheme" format is definitly simpler: to add a new interface, you can just take an existint json file as a model, adapt it, add it, and you are done. The current format makes this much harder.

It is also more modular, in a sort of trivial way: to add or remove an interface, just add or remove the json file.

The only point where this new scheme is superior (afaics) is that it provides a mechanism for re-using code. But I personally do not think this is worth the price.

src/components/Proposal/ProposalSummary/ProposalSummaryKnownGenericScheme.tsx
dkent600: Yeah, but I believe that fact will have been detected before ever reaching this component.

src/lib/util.ts
```diff
@@ -15,11 +15,17 @@ const Web3 = require("web3");
 const tokens = require("data/tokens.json");
 const exchangesList = require("data/exchangesList.json");
 
-export function getExchangesList() {
+export interface ITokenInfo {
```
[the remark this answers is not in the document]
dkent600: Lint

src/genericSchemeRegistry/schemes/DaoGenericSchemesForms.json
dkent600: > The "one-file-per-genericScheme" format is definitly simpler: to add a new interface, you can just take an existint json file as a model, adapt it, add it, and you are done. The current format makes this much harder.

In the separate-file pattern, when adding a new DAO (we should be talking about DAOs here)  one would have to also add a new `import` statement and a reference to it the import in the typescript.  The only way around this would be some dynamic importing, but then you move the problem to the webpack config to know how to bundle the json files.

In contrast, with I am proposing, adding a new DAO involves only a copy/paste/adapt of a simple section of JSON.  Period.  Even without noting the additional work regarding imports (in the para above), this would still be no more complicated than having separate files.

dkent600: > and (in the far (?) future) bundle them separately

I think in any future case, whatever query we run should ultimately (in the future beyond the current hard-coded local json, which is a hack) be running against a single data store, ideally ethereum or ipfs or whatever.  A single json file follows the pattern of a single data store and is easier to code against (no fluctuating imports, see the comment above).

The internal architecture of how the data is arranged within the store should not be the concern of the application typescript. The import pattern violates this separation of concerns.

src/components/Proposal/ProposalSummary/ProposalSummaryKnownGenericScheme.tsx
jellegerbrandy: I am not following you.  When the proposal regards a function call that is not recognized, we should tell the user that:  "Unknown function call, it looks like this: ..".  That is what this commeted code is doing.

src/genericSchemeRegistry/schemes/DaoGenericSchemesForms.json
jellegerbrandy: > We should be talking about DAOs here

Why? I think we should be talking aobut Schemes here, not DAOs.

> have to also add a new `import` statement

Dynamic loading is the way, and the JSON files would not be part of the bundle (see my remark about 1000s of custom components), but loaded separately.

> simple section of JSON

well, that is exactly my point,  that it is a section, so different DAOs with different schemes would ahve to "edit" the **same** file. 

> I think that in the future...

So clearly we do not have the same ideas about how this software would evolve and become more modular. In my view it would not be a single application, but an application framework with pluggable parts (of which these JSON files are an embrionic form)  - and so having a single file is quite the opposite of where I see this going. 

But it is way too early anyway to say something definitive about that.

The bottom line is: we have an existing, working, non-broken, implementation, this PR makes that implementation more complex,  so I'd prefer to not merge this...

dkent600: >We should be talking about DAOs here
>> Why? I think we should be talking aobut Schemes here, not DAOs.

For example, if dxDAO wanted the dao-to-dao, they can't as the JSON is currently structured (am I wrong?).  This scenario involves two different schemes in one DAO.  I think that we want that a DAO can customize what schemes it wants to use and how it wants to use them (the set of actions/forms).

dkent600: > so different DAOs with different schemes would ahve to "edit" the same file

Where, when and how are they going to do that?

dkent600: >  single application, but an application framework with pluggable parts 

Indeed soon, and I would note that "pluggable parts", while it has merits in some scenarios, is not an automated process from the DAO's perspective.  With specific respect to GenericScheme I envision a process that _can_ be fully automated, where resorting to a plugin would be a last resort when generic functionality will not suffice.

dkent600: Correcting myself:  It appears I _was_ wrong about:

> For example, if dxDAO wanted the dao-to-dao, they can't as the JSON is currently structured (am I wrong?).

Why?  Simply by virtue of the GenericScheme being registered twice, each with a different target contract address, the Schemes list page will ,for each, find the JSON containing the target contract address.

dkent600:
I think there are several issues with the current architeture that are resolved in this PR.  But there is not currently a desire to address those shortcoming, so am closing the PR.

Issues with the current architecture:

* Without DAO-specific configurations, every DAO that wants to use a particular target contract will be unable to customize it to their own needs, either by renaming it or including only a subset of actions (they will be forced to have all actions enabled even if they only want voting).

*  Requires duplication of JSON when there are multiple instances of a target contract. To enable DAOs to use other instances of GenesisProtocol, for example, would require duplication of GenesisProtocol.json, modifying only the target contract addresses.

* adding new target contracts requires modifying the typescript code which should not be reliant on a hard-coded set of target contracts.

* the typescript uses module imports and JSON to obtain the data, which then ties into webpack. The UI Component typescript should not be aware of the implementation of the data store, much less know about the data's internal architecture. (These issues are only partly solved in this PR.)

PR #961, opened 2019-07-22
Added ReputationFromTokenScheme

tibetsprague:
Waiting now for https://github.com/daostack/subgraph/issues/294 and https://github.com/daostack/client/issues/278 but @brodeur could style


src/components/Scheme/FixedReputationAllocationScheme.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+        <Sticky enabled top={50} innerZ={10000}>
+          <h2 className={schemeCss.schemeName}>
+            {schemeName(scheme, scheme.address)}
+          </h2>
+        </Sticky>
+
+        <div>
+          <h3>100 Rep to redeem </h3>
+          <b>Which account would you like to receive the reputation?</b>
```
dkent600: I wonder if we should have a more detailed helpful description for the user of what is going on here, why they are being presented with this UI?

tibetsprague: that would be up to @A-Zak

jellegerbrandy:
Just dropping some information for testing this here:

We need to test this with version rc.23 of the reptuationfromtokens contract. This contract is [here on rinkeby](https://rinkeby.etherscan.io/address/0x953fbf5fc33d2cc8eb3752976f0f8e92f1db5f67)

PR #971, opened 2019-07-25
Validate network status

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=userstory/797/silent


src/components/Proposal/Voting/VoteButtons.tsx
```diff
@@ -86,24 +86,22 @@ class VoteButtons extends React.Component<IContainerProps, IState> {
...
-            proposal.stage === IProposalStage.ExpiredInQueue || (proposal.stage === IProposalStage.Boosted && expired) || (proposal.stage === IProposalStage.QuietEndingPeriod && expired) ?
-              "Can't vote on expired proposals" :
-              proposal.stage === IProposalStage.Executed ?
-                "Can't vote on executed proposals" :
-                isVoting ?
-                  "Warning: Voting for this proposal is already in progress" :
-                  `Vote ${vote === IProposalOutcome.Pass ? "for" : "against"}`
+      currentVote ?
+        "Can't change your vote" :
+        currentAccountState.reputation.eq(new BN(0)) ?
```
tibetsprague: wait so there will always be a currentAccountState now?

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -10,26 +11,50 @@ import { Link, Route, RouteComponentProps, Switch } from "react-router-dom";
...
+      this.props.history.push(`/dao/${daoAvatarAddress}/scheme/${schemeId}/proposals/create`);
+    }
+  }
 
-  public render() {
+  public render(): any {
     const { currentAccountAddress, match } = this.props;
     const schemeId = match.params.schemeId;
     const daoAvatarAddress = match.params.daoAvatarAddress;
+    const _handleNewProposal = (e: any): void => { this.handleNewProposal(daoAvatarAddress, schemeId); e.preventDefault(); /* e.stopPropagation(); */};
```
tibetsprague: can we break this onto multiple lines

tibetsprague: also why do we have this? cant you put the preventDefault above?

src/components/Proposal/Voting/VoteButtons.tsx
dkent600: good catch

src/components/Scheme/SchemeContainer.tsx
dkent600: I modified to put it all together in one function.

dkent600:
Got some tests that need fixing...

src/components/Proposal/ActionButton.tsx
```diff
@@ -90,12 +91,12 @@ class ActionButton extends React.Component<IProps, IState> {
       daoEthBalance,
       detailView,
       proposalState,
-      redeemProposal,
       rewardsForCurrentUser,
     } = this.props;
 
     const executable = proposalEnded(proposalState) && !proposalState.executedAt;
     const expired = closingTime(proposalState).isSameOrBefore(moment());
+    const _handleRedeemProposal = this.handleRedeemProposal.bind(this);
```
jellegerbrandy: The pattern we use in other places (and I think is slightly better) is to do the binding in the consturctor. I.e. the constructor would contain the line:

`this.handleRedeemProposal = this.handleRedeemProposal.bind(this)`

And then you can use the bound `this.handleRedeemProposal` anywhere (not just in render)

src/components/Proposal/ActionButton.tsx
```diff
@@ -207,19 +224,19 @@ interface IMyProps {
 
 const ConnectedActionButton = connect(mapStateToProps, mapDispatchToProps)(ActionButton);
 
-export default (props: IMyProps) => {
+export default (props: IMyProps): any => {
```
jellegerbrandy: i propose we drop the explicit return types rules :-)

src/components/Proposal/ActionButton.tsx
```diff
@@ -207,19 +224,19 @@ interface IMyProps {
...
 
   const proposalState = props.proposalState;
   let observable: Observable<IRewardState>;
   if (props.currentAccountAddress) {
     observable = proposalState.proposal.rewards({ where: {beneficiary: props.currentAccountAddress}})
       .pipe(map((rewards: Reward[]): Reward => rewards.length === 1 && rewards[0] || null))
-      .pipe(mergeMap(((reward) => reward ? reward.state() : of(null))));
+      .pipe(mergeMap(((reward): Observable<IRewardState> => reward ? reward.state() : of(null))));
```
jellegerbrandy: This annotation is wrong, though - it returns `Obserable<IRewardState|null>`. (Again, I vote for turning explicity return type off :-)

dkent600: Thanks for catching that!  I think this is an example of reasons why I strongly support requiring about return-type annotations:  They document the code, which lessens the (re)learning curve for anyone working with the code, they enrichen intellisense feedback which facilitates coding, and they make it clear that `null` is a valid possibility that the code needs to account for, and by such accounting, they reduce risk.

Would be interested in why you would want to ditch this rule.

src/components/Proposal/ActionButton.tsx
dkent600: I'm reflexively suspicious of that pattern, but anyway went ahead and made it so.

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -8,28 +9,51 @@ import * as React from "react";
...
+};
 
-  public render() {
+class SchemeContainer extends React.Component<IProps & RouteComponentProps<any>, IExternalProps> {
+
+  public render(): any {
     const { currentAccountAddress, match } = this.props;
     const schemeId = match.params.schemeId;
     const daoAvatarAddress = match.params.daoAvatarAddress;
+    const handleNewProposal = async (e: any): Promise<void> => {
```
tibetsprague: i would normally put this outside of the render function, feels weird to put it inside, but not a big deal

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -63,9 +87,11 @@ export default class SchemeContainer extends React.Component<IProps & RouteCompo
           <div className={css.schemeMenu}>
             <Link className={proposalsTabClass} to={`/dao/${daoAvatarAddress}/scheme/${scheme.id}/proposals/`}>Proposals</Link>
             <Link className={infoTabClass} to={`/dao/${daoAvatarAddress}/scheme/${scheme.id}/info/`}>Info</Link>
-            <Link className={css.createProposal} to={`/dao/${daoAvatarAddress}/scheme/${scheme.id}/proposals/create`} data-test-id="createProposal">
-              + New proposal
-            </Link>
+            <a className={css.createProposal}
+              data-test-id="createProposal"
+              href="#"
```
tibetsprague: this will add a # to the URL, thats the issue you are seeing. better to do href="javascript:void(0)"

src/components/Scheme/SchemeProposalsPage.tsx
```diff
@@ -94,7 +105,11 @@ class SchemeProposals extends React.Component<IProps, IState> {
               <Link to={"/dao/" + dao.address}>
                 <img className={css.relax} src="/assets/images/lt.svg"/> Back to schemes
               </Link>
-              <Link to={`/dao/${dao.address}/scheme/${scheme.id}/proposals/create/`} data-test-id="createProposal" className={css.blueButton}>+ New Proposal</Link>
+              <a className={css.blueButton}
+                href="#"
```
tibetsprague: same here

src/components/Scheme/SchemeContainer.tsx
dkent600: that is a lot easier as is so will leave it.

src/components/Scheme/SchemeProposalsPage.tsx
dkent600: fixed

src/components/Scheme/SchemeContainer.tsx
dkent600: fixed

src/components/Scheme/SchemeContainer.tsx
jellegerbrandy: so there is a reason to not put it in the render function: the render function is called very often, and so as a rule should be as 'light" as possible. (Not that a function def like this gives significant overhead, but as a general ruleof thumb, what can be done outside of the render function should be outside)

dkent600: this is so trivial a "load" I don't see any reason to worry about it.  the alternative is more complicated, and therefore risky, code.  the only place this function is used is from within render, and from that perspective it makes perfect sense to define it within, though yes, out of habit, normally I would likely put the function outside of render, as I tried to do but changed my mind as this version is simpler.

If you are insisting on doing this to approve  this PR, then I will, otherwise no.  @tibetsprague   @jellegerbrandy

jellegerbrandy: It is a line or two of code extra, but it is the standard pattern, and for the good reasons that I mentioned. (The "more risky" argument sounds absurd to me :-))

PR #981, opened 2019-07-29
Refreshing

jellegerbrandy:
- make it so that reload is not needed anymore


dkent600:
Just curious;  Is this really all that is required?  Or just a start on it?

jellegerbrandy:
yeah i think this is it, or at least the main thing: a few lines, many hours of work :-) (there was other more invislbe work on the client lib to prepare, though)

but I need to check the web3 provider cache in the client - so I marked this as WIP.

jellegerbrandy:
This seems to me ready to go, but given that this covers lots of different use cases and user behavior, it would be good to doublecheck this on the review app: https://alchemy-staging-rinkeby-pr-981.herokuapp.com/

PR #1005, opened 2019-08-02
Temporary undo the dao registry filtration

A-Zak:


jellegerbrandy:
because of bad comms the dao was not officially registered and is not visible

PR #1012, opened 2019-08-06
upgrade client (should fix redemption bug)

jellegerbrandy:


dkent600:
I wonder if client should have a test that would've have caught this?

jellegerbrandy:
caught what? the redemption bug? Probably yes, but it was a complex bug to test (it was a redeemer-version confusino, and on production there are many more versions around then in the test environment)

Issue #1013, opened 2019-08-06
Logging out using the MetaMask UI freezes Alchemy's "Please log in!" button.


dkent600:
Logging out using the MetaMask UI freezes Alchemy's "Please log in!" button.

![image](https://user-images.githubusercontent.com/1821666/62536805-ec073080-b81c-11e9-85ce-04461afa10f5.png)

Issue #1011, opened 2019-08-06
"Log out" now also resets the account information


dkent600:
From @jellegerbrandy:

"Log out" now also resets the account information (i.e. usage becomes "anonymous"). Is that what we want?

A-Zak:
yes, this is good

PR #1015, opened 2019-08-06
don't freeze login button after MM account logout

dkent600:
Resolves: #1013 

**Note** Logging out via MM will now have the same behavior as logging out from within the app. 
 See:  https://github.com/daostack/alchemy/issues/1011

### DIFF
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -398,27 +398,31 @@ async function getCurrentAccountFromProvider(): Promise<Address | null> {
  * switch to readonly mode (effectively a logout)
  */
 export async function gotoReadonly(showNotification?: any): Promise<boolean> {
-  // clearing this, initializeArc will be made to use the default web3Provider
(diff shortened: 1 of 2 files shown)


src/arc.ts
```diff
@@ -398,27 +398,31 @@ async function getCurrentAccountFromProvider(): Promise<Address | null> {
...
-    /* hold off on this for now: 
-    if (success && showNotification) {
-      showNotification(NotificationStatus.Success, `Logged out from ${networkName}`);
+  if (selectedProvider) {
+    // clearing this, initializeArc will be made to use the default web3Provider
+    const networkName = await getNetworkName();
+    selectedProvider = undefined;
+    try {
+      success = await initializeArc();
+      /* hold off on this for now:
```
tibetsprague: should we just delete?

dkent600: Done

PR #1024, opened 2019-08-07
add basic unit test

jellegerbrandy:
* define `npm run test:unit`
* write a unit test for formatToken
* make sure that formatToken rounds down (to address https://daostack.tpondemand.com/RestUI/Board.aspx#page=profile&boardPopup=userstory/1258/silent)
* make the App.spec js test work

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,23 +16,24 @@
   "jest": {
     "testURL": "http://127.0.0.1:3000/",
     "transform": {
-      ".(ts|tsx)": "<rootDir>/node_modules/ts-jest/preprocessor.js"
+      ".(ts|tsx)$": "ts-jest"
     },
     "testPathIgnorePatterns": [
       "/node_modules/",
       "/lib/",
       "/test/integration/"
     ],
-    "testRegex": "(/test/.*|\\.(test|spec))\\.(ts|tsx|js)$",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -16,35 +16,40 @@
...
       "tsx",
       "js",
       "jsx",
-      "json"
+      "json",
+      "node"
     ],
     "moduleNameMapper": {
-      "\\.(scss|css|less)$": "identity-obj-proxy",
+      "\\.(scss|css|less|svg|png|jpg)$": "identity-obj-proxy",
```
tibetsprague: what is this for?

src/lib/util.ts
```diff
@@ -6,7 +6,7 @@ import {
   IProposalState,
   IRewardState,
   ISchemeState } from "@daostack/client";
-import { GenericSchemeRegistry } from "genericSchemeRegistry";
+import { GenericSchemeRegistry } from "../genericSchemeRegistry";
```
tibetsprague: isn't this fixed by the change you made in package.json?

package.json
jellegerbrandy: this is for jest to understand what to do with svg  png jpg files (namely: nothing at all)

src/lib/util.ts
jellegerbrandy: ah yeah that's from before the fix

Issue #1020, opened 2019-08-06
Subscribe component calling setState in render


tibetsprague:
e41844836889be3b8d0363ebcf51005047fc9721 causes react warnings, and probably issues by calling setState in render(). this.setupSubscription should be called in an earlier component lifecycle function like componentDidMount

jellegerbrandy:
what kind of warnings? How do you know they are from that commit?

By way of explanation: 

I moved it way from componentwillmount because that hook is deprecated: https://reactjs.org/docs/react-component.html#unsafe_componentwillmount

And I am not calling setState in the render() method in any unrecommended way, I think. What is happening is the the render sets up a new subscrition (if the observable has changed), that subscription will fetch the data and only then reset the state (and re-render()). 

(btw, componentDidMont is not earlier but later in the lifecyle: https://reactjs.org/docs/react-component.html#the-component-lifecycle)

dkent600:
Newbie suggestion:  May be optimal to memoize it?

https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

jellegerbrandy:
what is "it" in your suggestion?

tibetsprague:
"componentDidMount() is invoked immediately after a component is mounted (inserted into the tree). Initialization that requires DOM nodes should go here. If you need to load data from a remote endpoint, this is a good place to instantiate the network request.

**This method is a good place to set up any subscriptions. If you do that, don’t forget to unsubscribe in componentWillUnmount().**"

jellegerbrandy:
fixed in #1025

PR #1025, opened 2019-08-07
Fix incorrect use of setState in render() function of Subscribe

tibetsprague:
https://github.com/daostack/alchemy/issues/1020


tibetsprague:
actually this is broken, im working on a better solution

PR #1038, opened 2019-08-13
[WIP] More tests

tibetsprague:
Testing subscriptions by adding names to each one so we know which one is doing what when

Working on full integration test with time traveling


jellegerbrandy:
@tibetsprague  this is pretty old and seems dead code. I'm closing this for now, please re-open if you think it is still relevant...

Issue #1000, opened 2019-08-02
delete all auto-generated .scss.d.ts files at the start of each build


dkent600:
.scss.d.ts files are not rebuilt if their date is newer than the date of the associated .scss file.  This can break a build after doing an `npm ci`.  

So the proposal is to delete all auto-generated .scss.d.ts files at the start of each build,

jellegerbrandy:
how does it break the build?

dkent600:
It can break the build if the former, un-rebuilt typings are missing types referenced in newer code.

tibetsprague:
or we could go back to adding the files to the repo and not worry about this

jellegerbrandy:
we also do not want to delete them before each build, as this slows down the building considerably.

dkent600:
> his slows down the building considerably

Looking at `npm run start`, I am not detecting any delay at all, much less anything that would be a significant proportion of the total build time.  What are you seeing?

jellegerbrandy:
I thought I saw a significant dely, but perhaps it was an optical illusion.

PR #1042, opened 2019-08-13
remove *.scss.d.ts on package install

dkent600:
Resolves: #1000

- removes *.scss.d.ts on npm package install
- adds script build-dev
- changes lint.andFix to lint-andFix, for script naming consistency


package.json
```diff
@@ -54,10 +54,12 @@
   "scripts": {
     "analyze": "NODE_ENV=production ANALYZE=1 node --max_old_space_size=4096 node_modules/.bin/webpack --config webpack.prod.config.js --progress --colors",
     "build": "webpack --config webpack.prod.config.js",
+    "build-dev": "webpack --config webpack.dev.config.js",
```
jellegerbrandy: why do we/you need this?

package.json
```diff
@@ -54,10 +54,12 @@
...
     "build-travis": "cross-env NODE_ENV=production node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js --config webpack.prod.config.js",
     "heroku-postbuild": "npm run build",
     "lint": "eslint --ext .js,.ts,.tsx ./src ./test",
-    "lint.andFix": "npm run lint -- --fix",
+    "lint-andFix": "npm run lint -- --fix",
+    "postinstall": "rimraf ./src/**/*.scss.d.ts",
```
jellegerbrandy: Will this solve the problem in all cases? I encountered the problem myself, and this was not because I re-installed the package, it was because I checkout out a different branch....

dkent600: Hmm, good question.  In some cases a different branch requires doing an `npm install`, but in some cases not.  Theoretically in that case then one can run `npm run postinstall`.  But if there is a preference to make this happen on every build, which was the original plan, then I can do that.  Let me know.

package.json
dkent600: I occasionally wish to test the build without bringing up the dev server.

Issue #1036, opened 2019-08-12
reputionfortoken redeeming by sending a private key


jellegerbrandy:
For dhack in berlin the scenario is as follows:

- user gets a QR code of a URL
- the url is of the form `link_to_rep4tokenscheme?pk=0x0000000` and shows a form (similar to) [this form right here](https://alchemy.daostack.io/dao/0x0ed985925bb42c6719d10dcd1cc02d8cf596c15b/scheme/0x5f400b33b15208e84ec419ac0789f7accb5b8e20785b6044cbb9eb438cfe3e7b)
- the rep4token scheme page behaves as follows: if there is a pk given together with the request, when the user clicks "send", alchemy will not send the tx to the wallet to sign, but instead will sign it immediately with the private key given in the request. (otherwise, it will just behave as it does now)

PR #1044, opened 2019-08-14
redeem with private key

jellegerbrandy:
This PR implements the functionality from issue https://github.com/daostack/alchemy/issues/1036

### DIFF
--- a/src/actions/arcActions.ts
+++ b/src/actions/arcActions.ts
@@ -150,9 +150,13 @@ export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: strin
       redeemMethod.gasPrice = gasPrice;
       const signedTransaction = await arc.web3.eth.accounts.signTransaction(redeemMethod, privateKey);
       dispatch(showNotification(NotificationStatus.Success, "Sending redeem transaction, please wait for it to be mined"));
+      const txHash = await arc.web3.utils.sha3(signedTransaction.rawTransaction);
+
+      console.log(`transaction hash: ${txHash}`);
       try {
-        await arc.web3.eth.sendSignedTransaction(signedTransaction.rawTransaction);
+        const receipt  = await arc.web3.eth.sendSignedTransaction(signedTransaction.rawTransaction);
         dispatch(showNotification(NotificationStatus.Success, "Transaction was succesful!"));
+        console.log(receipt);
       } catch(err) {
         dispatch(showNotification(NotificationStatus.Failure, `Transaction failed: ${err.message}`));
       }
--- a/src/actions/arcActions.ts
+++ b/src/actions/arcActions.ts
@@ -149,9 +149,6 @@ export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: strin
       const gasPrice = await arc.web3.eth.getGasPrice();
       redeemMethod.gasPrice = gasPrice;
       const signedTransaction = await arc.web3.eth.accounts.signTransaction(redeemMethod, privateKey);
-      console.log(signedTransaction)
-      console.log(signedTransaction.tx)
-      console.log(signedTransaction.tx.hash)
       dispatch(showNotification(NotificationStatus.Success, "Sending redeem transaction, please wait for it to be mined"));
       try {
         await arc.web3.eth.sendSignedTransaction(signedTransaction.rawTransaction);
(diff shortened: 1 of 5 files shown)


src/components/Scheme/ReputationFromToken.tsx
```diff
@@ -58,23 +62,43 @@ class ReputationFromToken extends React.Component<IProps, IState> {
...
+    const pk = queryValues["pk"] as string;
+    if (pk) {
+      console.log(`Found private key ${pk}`);
+      const arc = getArc();
+      try {
+        redeemerAddress = arc.web3.eth.accounts.privateKeyToAccount(pk).address;
+      } catch(err) {
+        throw Error(`Invalide private key: ${pk}`);
+      }
+      // redeemerAddress = "0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1";
```
dkent600: remove?

src/components/Scheme/ReputationFromToken.tsx
```diff
@@ -116,13 +144,15 @@ class ReputationFromToken extends React.Component<IProps, IState> {
     if (alreadyRedeemed) {
```
A-Zak: Already redeemed should be based on the `redeemerAddress`, not the `accountAddress`
Looks like some nasty copy-pasting from line  101, probably worth to refactor out to one method ^_^

jellegerbrandy: good call, this was already a bug in the previous version..

src/actions/arcActions.ts
```diff
@@ -135,15 +135,38 @@ export function redeemProposal(daoAvatarAddress: string, proposalId: string, acc
...
+      const gas = block.gasLimit - 100000;
+      const redeemMethod = contract.methods.redeem(addressToRedeem);
+      redeemMethod.gas = gas;
+      const signedTransaction = await arc.web3.eth.accounts.signTransaction(redeemMethod, privateKey);
+      dispatch(showNotification(NotificationStatus.Success, "Sending redeem transaction, please wait for it to be mined"));
+      console.log("Sending transaction");
+      try {
+        const receipt = await arc.web3.eth.sendSignedTransaction(signedTransaction.rawTransaction);
+        console.log(`Transaction sent: ${receipt}`);
+        dispatch(showNotification(NotificationStatus.Success, "Transaction was succesfull!"));
```
tibetsprague: successful

src/actions/arcActions.ts
```diff
@@ -135,15 +135,38 @@ export function redeemProposal(daoAvatarAddress: string, proposalId: string, acc
...
 
-export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: string) {
+export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: string, redeemerAddress: Address, privateKey: string|undefined) {
   return async (dispatch: Redux.Dispatch<any, any>) => {
-    const observer = operationNotifierObserver(dispatch, "Redeem reputation");
-    const reputationFromTokenScheme = scheme.ReputationFromToken as ReputationFromTokenScheme;
+    const arc = getArc();
+    if (privateKey) {
+      console.log("SIGNING WITH PK");
+      const reputationFromTokenScheme = scheme.ReputationFromToken as ReputationFromTokenScheme;
```
tibetsprague: shouldnt this all be in the client library?

jellegerbrandy: well, yes, you are probably right - it started as a one-liner but then I needed more. Also it needs to be deployed quickly, so I chose the quick solution.

src/actions/arcActions.ts
```diff
@@ -135,15 +135,36 @@ export function redeemProposal(daoAvatarAddress: string, proposalId: string, acc
   };
 }
 
-export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: string) {
+export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: string, redeemerAddress: Address, privateKey: string|undefined) {
```
dkent600: `redeemerAddress` is never referenced

dkent600: i'd like to know why lint didn't catch this

jellegerbrandy: good point

PR #1048, opened 2019-08-15
1397 new notifications

brodeur:
https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=userstory/1397/silent


tibetsprague:
@jellegerbrandy notifications not disappearing is already happening currently, not sure when it broke

jellegerbrandy:
ok then I'll merge this and make a separate issue for the timeout

dkent600:
I don't think this looks so great.  See the top of the close button in the header:

image

dkent600:
And I think errors should probably have a red top border, which is missing:

image

dkent600:
There is no tooltip such as is called-for in the mocl:

image

dkent600:
I really don't get why the colored top-borders are not 100% width in the mock:

image

PR #1057, opened 2019-08-16
React select

brodeur:
FINALLY! Styled dropdowns for token select. This task really showed me some of the limitations of working with REACT ^_^


jellegerbrandy:
Hey I fixed the merge conflicts and tested it, seems to work but I cannot really see _what_ is being styled here...

PR #1082, opened 2019-08-24
ProposalCard render props component to load proposal data in one …

tibetsprague:
…place


src/components/Dao/DaoHistoryPage.tsx
```diff
@@ -30,7 +31,21 @@ class DaoHistoryPage extends React.Component<IProps, null> {
     const { currentAccountAddress } = this.props;
 
     const proposalsHTML = proposals.map((proposal: Proposal) => {
-      return (<ProposalHistoryRow key={"proposal_" + proposal.id} proposal={proposal} dao={dao} currentAccountAddress={currentAccountAddress}/>);
+      return <ProposalData currentAccountAddress={currentAccountAddress} dao={dao} proposalId={proposal.id} key={proposal.id}>
```
tibetsprague: why do you think this is better than putting <ProposalData inside ProposalHistoryRow? i guess i dont have an opinion...

tibetsprague: also it seems like we shouldnt even use ProposalData for ProposalHistoryRow because it adds so many unecessary subscriptions.

PR #1097, opened 2019-08-29
fix number formatting for 999 >= n > 99

dkent600:
Resolves:  https://daostack.tpondemand.com/restui/board.aspx?#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1559/silent

Was failing for numbers between 100 and 999.

See also:  https://github.com/daostack/alchemy/issues/1103


src/lib/util.ts
```diff
@@ -133,9 +133,9 @@ export function formatTokens(amountWei: BN, symbol?: string, decimals = 18): str
     significantDigits = 1000000000;
     units = "B";
   }
-  else if (tokenAmount.ltn(1000)) {
+  else if (tokenAmount.ltn(100000)) {
```
tibetsprague: Doesn't this need to be a BN too?

dkent600: `tokenAmount` _is_ a BN

tibetsprague: Yes but 100000 is not

dkent600: The 1000000 doesn't have to be a BN.  It fits just fine in a `number` and `.ltn()`. takes a number by definition.  The function does have its own smaller limit on the size of the param, but throws an exception if that number is too big, which it isn't.

dkent600: The proof is that the unit tests succeed  :-)

tibetsprague: ok fine, but then why did you change the other one to new BN(...)?

dkent600: because that one exceeded the maximum value that `ltn()` allows for the input value.

PR #1125, opened 2019-09-16
Fix infinite scroll loading of queued scheme proposals

tibetsprague:
There is one more bug to fix related to the time we use to check for when proposals expired, tricky one...

https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1639/silent


jellegerbrandy:
This is still not working perfectly, see here: https://alchemy-staging-rinkeb-pr-1125.herokuapp.com/dao/0x72939947e7a1c4ac94bb840e3304b322237ad1a8/scheme/0x16b81c56297ae81ffa32ed128f17dbf7b108eb9200d2c8110a9ff2e05e095995
image

jellegerbrandy:
BUT: Ben just confirmed to me that the bot also moves the "Queued" to "expiredinQueue" proposals.

So we can simplify our logic considerably, and just query for "Queued" proposals without worrying about the time. 

Can you refactor with that in mind?

tibetsprague:
@jellegerbrandy hmm, what about locally and on Rinkeby? I guess we just don't care about those working correctly?

jellegerbrandy:
Well, it would work "correctly" just the same - we just changed the specs :-)

Seriously, we should hide the expired proposals anyway, so the visible behavior would be the same, it is just that we do not do this with the query, but filter those props afterwards.

jellegerbrandy:
Maybe it better to just finish this PR as-is, so we can deploy it, and do the query refactoring is in a separate PR. (sorry for the confusion)

tibetsprague:
@jellegerbrandy fixed a nasty bug, if you create a setTimeout too far in the future (around 24 days) it breaks and immediately fires the timeout! that was causing proposals to "expire" immediately

tibetsprague:
and making the numbers off

PR #1128, opened 2019-09-18
better return type for render

dkent600:
adds `RenderOutput` as the return type for all `render()` functions.


dkent600:
1. if we ever turn back on the requirement for return types (which I think we should), then these would be about 75 errors that would not be triggered.
2. the only reason I didn't do this a long time ago is that I didn't know that there actual _was_ a type that would work in reactjs.  I just recently discovered it.
3.  there are many reasons why explicit return types are in general desirable.  I've enumerated them more than once in the course of our discussions.  When there are questions about a rule, one can find brief justifications for each rule in the lint rules documentation, such as for [this rule, here](https://github.com/bradzacher/eslint-plugin-typescript/blob/master/docs/rules/explicit-function-return-type.md).  The return type in this case is helpful for documenting the code and for catching cases where we're returning the wrong type from `render`.  It can help guide intellitype to suggest and correct as we type.
4. But I suggest that just because we disabled this rule doesn't mean that we should be prevented from supplying return types.

PR #1129, opened 2019-09-18
Less subscriptions 2

jellegerbrandy:
- having better metrics by explicitly logging the rqeuest and subscriptions gives us:

what | queries | subs | url
|----|----|----|---|
CR page | 35 |  91 | [page](http://127.0.0.1:3000/dao/0x294f999356ed03347c7a23bcbcf8d33fa41dc830/scheme/0x28c5b9efd5bdec2c69c613d2df4b5e1b92e44a2d3c2f5092fb45187570029009)


src/components/Scheme/SchemeContainer.tsx
```diff
@@ -112,9 +114,18 @@ const SubscribedSchemeContainer = withSubscription({
   loadingComponent: <div className={css.loading}><Loading/></div>,
   errorComponent: null,
   checkForUpdate: ["schemeId"],
-  createObservable: (props: IProps) => {
+  createObservable: async (props: IProps) => {
     const arc = getArc();
     const scheme = arc.scheme(props.schemeId);
+
+    // prime the cache: get all members fo this DAO
+    await props.daoState.dao.members({ first: 100, skip: 0 }).pipe(first()).toPromise();
```
tibetsprague: you have to prime the cache with the exact paginated request we will use?

jellegerbrandy: no. I'd like to just get all members in one single query. But I can't bc The Graph ppl have set a maximum request amount of 100.

jellegerbrandy:
Ok so I took a different approach. The default is that no query sends a subscription at all; instead, we defie our subscriptions separately. 

**TLDR:** we can do with < 5 subs for each page once:

- https://github.com/daostack/subgraph/issues/326 
- https://github.com/graphprotocol/support/issues/25
- we accept we do not subscribe to reputation to updates of other members

**Current state is only about subscriptions. The query refactoring is still work in progress**

## homepage (list of DAOs)

2 Subscriptions (to get the list of DAOs). Will become one (we currently have two queries, one to get the GP protocol, one to get the rest)

- this should be enough to handle all updates

- proposal count will not be updated automatically until the subgraph issue about proposal count is resolved

# Each page (except thehome page)

- 1 to get the nr of redemptions (left menu)
- 1 to get current rep of user (for user profile)

# DAO page

http://127.0.0.1:3000/dao/0x294f999356ed03347c7a23bcbcf8d33fa41dc830

11 subscriptions for 3 schemes.

- the 2 common subs
- 1 to get the schemes
- 3 x 3 to get proposal counts. Will become 0 when relevant subgraph issue is resolved

## scheme page

http://127.0.0.1:3000/dao/0x294f999356ed03347c7a23bcbcf8d33fa41dc830/scheme/0x28c5b9efd5bdec2c69c613d2df4b5e1b92e44a2d3c2f5092fb45187570029009

Currently with 22 proposals

Uses 71 subscriptions (and so is still problematic)

the problem is that here that we need to query (and subscribe to) the list of votes, stakes and rewards of the current user - which makes for 22 x 3 = 66 subscriptions. Fixing this is blocked by a bug in the graph:  https://github.com/graphprotocol/support/issues/25; we can then subscribe with a single query


## reputation holders page

Currently does not subscribe to any additional updates. 

## history page

Currently does not subscribe to any additional updates

src/components/Scheme/SchemeContainer.tsx
jellegerbrandy: This is a graphql setting, and we will change this, so we can use a single query.

jellegerbrandy:
closing bc this was merged together with another PR

PR #1136, opened 2019-09-25
refactor hamburger button

dkent600:
Refactors the hamburger button to be part of the header instead of part of the sidebar.  This simplifies the coding and CSS layout out both of them.

Originally was about: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1623/silent

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,6 @@
     - Display proper message when no history
     - avoid duplicate Genesis Alpha cards
     - fix action button not appearing when proposal card countdown completes
-    - fix display glitch in the DAO sidebar on Firefox
 
 
 ### version 0.9.4; 2019-08-13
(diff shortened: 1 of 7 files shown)


dkent600:
@jellegerbrandy I am not able to reproduce this.  Maybe some clues?

dkent600:
@jellegerbrandy So this is just going to be about refactoring the hamburger button.  I've backed out all the attempts I made to make firefox behave like chromium.  Note that the “orange line” is actually a scrollbar. It appears on both chromium and firefox, the only difference being when it appears, as the two engines compute different heights for the sidebar.

PR #1138, opened 2019-09-25
Release fixes

jellegerbrandy:
fixes bugs in current release candidate:

- show addresses of unknown schemes
- do not crash redemptiosn page is callDAta of proposal is not recognized

### DIFF
--- a/src/components/Proposal/ProposalSummary/ProposalSummaryDutchX.tsx
+++ b/src/components/Proposal/ProposalSummary/ProposalSummaryDutchX.tsx
@@ -18,7 +18,16 @@ export default class ProposalSummaryDutchX extends React.Component<IProps, null>
 
   public render(): RenderOutput {
     const { proposal, detailView, genericSchemeInfo, transactionModal } = this.props;
-    const decodedCallData = genericSchemeInfo.decodeCallData(proposal.genericScheme.callData);
+    let decodedCallData: any;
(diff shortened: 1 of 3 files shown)


src/components/Proposal/ProposalSummary/ProposalSummaryDutchX.tsx
```diff
@@ -18,7 +18,16 @@ export default class ProposalSummaryDutchX extends React.Component<IProps, null>
 
   public render(): RenderOutput {
     const { proposal, detailView, genericSchemeInfo, transactionModal } = this.props;
-    const decodedCallData = genericSchemeInfo.decodeCallData(proposal.genericScheme.callData);
+    let decodedCallData: any;
+    try {
+      decodedCallData = genericSchemeInfo.decodeCallData(proposal.genericScheme.callData);
+    } catch(err) {
+      if (err.message.match(/no action matching/gi)) {
+        return <div>Error: {err.message} </div>;
```
dkent600: Would be nicer and more consistent with other errors if this error message were a notification

dkent600: Though on second thought what I suggested above doesn't make sense here, so I think the better improvement might be in the text itself to explain more clearly that the referenced contract is not found.

PR #1142, opened 2019-09-26
Add info for generic scheme info for ENS registry contract

tibetsprague:
Support setOwner, setTTL, setResolver, and setSubnodeOwner (create subdomain) to start

https://docs.ens.domains/contract-api-reference/ens

TODO: setup DAO with test ENS contract locally and on rinkeby


jellegerbrandy:
The test is running locally for me now, but times out in travis, so I don't know what's up.

I also changed the destintation address for the ENS action on rinkeby to a dummy one that is currently deployed.

PR #1144, opened 2019-09-26
Cached provider should always be used when possible

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1648/silent

This completes the second part of the ticket:  "The logic should be the same regardless of how the user chooses to connect / login / make an action (e.g. stake when not connected)…  If we have a cached provider then use it, if no cached provider - show the web3connect dialogue…"

- calls to enableWeb3ProviderAndWarn now call enableWalletProvider instead
- unlike the old enableWeb3ProviderAndWarn, enableWalletProvider always looks to the cache first
- **Further note**  I am proposing here to issue a notification when the user is automatically connected (but not when doing it manually through the Connect or Log In buttons).

### DIFF
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -412,7 +412,7 @@ async function enableWeb3Provider(provider?: any, blockOnWrongNetwork = true): P
  * @param provider Optional web3Provider
  * @return boolean whether Arc is successfully initialized.
  */
-export async function enableWeb3ProviderAndWarn(showNotification?: any, blockOnWrongNetwork = true): Promise<boolean> {
+async function enableWeb3ProviderAndWarn(showNotification?: any, blockOnWrongNetwork = true): Promise<boolean> {
   let success = false;
   let msg: string;
   try {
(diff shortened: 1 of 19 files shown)


src/arc.ts
```diff
@@ -566,6 +576,77 @@ export function getAccountIsEnabled(): boolean {
   return !!getWeb3Provider();
 }
 
+const ACCOUNTSTORAGEKEY = "currentAddress";
```
tibetsprague: nice to use underscores between words in constants

src/components/Proposal/ActionButton.tsx
```diff
@@ -68,12 +68,16 @@ class ActionButton extends React.Component<IProps, IState> {
   }
 
   public async handleClickExecute(_event: any): Promise<void> {
-    if (!(await enableWeb3ProviderAndWarn(this.props.showNotification.bind(this)))) { return; }
+    if (!await loadCachedWeb3Provider(this.props.showNotification) &&
```
tibetsprague: if we always use this pattern then shouldn't we have a function that calls and checks both loadCached and enable?

dkent600: yeah, I keep thinking the same thing.  I'll do it.

src/arc.ts
dkent600: Done

src/components/Proposal/ActionButton.tsx
dkent600: Done

PR #1151, opened 2019-10-03
Feedback on whether there are unread Discus posts on DAO Wall

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=userstory/1649/silent

Show red dot next to "DAO Wall" in sidebar when there are unread messages.

**Important**:  To use the Disqus API I had to create an "[API Application](https://help.disqus.com/en/articles/1717083-how-to-create-an-api-application)" in Disqus.  I have enabled access to any domain for debugging until this PR is merged.  Access is limited to "readonly" in any case.

**Note** API access is throttled without upgrading to a "Pro" subscription, which we already have.  Yet it is reporting that we have a limit of 1000 queries per hour.

Here is the API Application config:  https://disqus.com/api/applications/6076435/

### DIFF
--- a/src/components/Dao/DaoSidebar.tsx
+++ b/src/components/Dao/DaoSidebar.tsx
@@ -259,11 +259,7 @@ const SubscribedDaoSidebar = withSubscription({
   loadingComponent: <div className={css.loading}><Loading/></div>,
   createObservable: (props: IProps) => {
 
-    const lastAccessDate = localStorage.getItem(`daoWallEntryDate_${props.dao.address}`);
-
-    if (!lastAccessDate) {
-      return from(Promise.resolve({ hasNewPosts : false }));
-    }
+    const lastAccessDate = localStorage.getItem(`daoWallEntryDate_${props.dao.address}`) || "0";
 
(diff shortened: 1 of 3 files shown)


dkent600:
> I expected the red dot to disappear or not even show up, given that I am reading the page with the discussion.

The red dot should disappear once you visit the DA Wall page and refresh.

dkent600:
@jellegerbrandy Ok I have been led in circles by the apparent reality that, despite the Disqus documentation to the contrary, using a Unix timestamp to specify the date simply does not work as they have spec'd.  So I have modified our code to use an ISO-formatted date, which does work.

src/components/Dao/DaoSidebar.tsx
```diff
@@ -241,4 +253,32 @@ const SubscribedTokenBalance = withSubscription({
   },
 });
 
-export default connect(mapStateToProps)(DaoSidebar);
+const SubscribedDaoSidebar = withSubscription({
```
tibetsprague: intriguing use of withSubscriptions. seemed weird at first, but I guess it makes some sense

PR #1164, opened 2019-10-11
handle SchemeRegistrar special case

dkent600:


src/components/Scheme/SchemeContainer.tsx
```diff
@@ -53,6 +55,28 @@ const mapDispatchToProps = {
 
 class SchemeContainer extends React.Component<IProps, null> {
 
+  private getSchemeIsActive(scheme: ISchemeState): boolean {
```
tibetsprague: seems like this code should be in the client

jellegerbrandy: I agree, but we can first write it here and later port it to the client

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -53,6 +55,28 @@ const mapDispatchToProps = {
...
+    if (scheme.name === "SchemeRegistrar") {
+      /**
+       * If SchemeRegistrar has only one or the other active, then we'll deal with that case in
```
jellegerbrandy: it would be clearer if the "isSchemeActive"  logic is all in one single place - and change the signature to something like `getSchemeisActive(scheme, action)`

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -53,6 +55,28 @@ const mapDispatchToProps = {
...
+      let name = `${scheme.name[0].toLowerCase()}${scheme.name.slice(1)}`;
+      if (name === "genericScheme") {
+        if (scheme.uGenericSchemeParams) {
+          name = "uGenericScheme";
+        }
+      }
+      /**
+     * assumes the voting machine is GP and its params can be found at scheme.<schemeName>Params.voteParams
+     */
+      return moment((scheme as any)[`${name}Params`].voteParams.activationTime).isSameOrBefore(moment());
```
jellegerbrandy: As you note in the PR description, this will fail if a scheme does not have these parameters set. It would be better (more future-proof, in any case) to handle that case explicitly - we can in that case just assume that the scheme is active....

src/components/Scheme/SchemeContainer.tsx
dkent600: This comment is confusing because that *is* the name of the function.

dkent600: Done

src/components/Scheme/SchemeContainer.tsx
dkent600: Done

PR #1166, opened 2019-10-14
Less subscriptions 4

jellegerbrandy:


src/components/Proposal/ProposalData.tsx
```diff
@@ -120,9 +120,9 @@ export default withSubscription({
       return combineLatest(
         proposal.state(), // state of the current proposal
         // TODO: do these subscription on the schemespage level
-        proposal.votes({where: { voter: currentAccountAddress }}, { subscribe: true }),
-        proposal.stakes({where: { staker: currentAccountAddress }}, { subscribe: true }),
-        proposal.rewards({ where: {beneficiary: currentAccountAddress}},{ subscribe: true } )
+        proposal.votes({where: { voter: currentAccountAddress }}),
+        proposal.stakes({where: { staker: currentAccountAddress }}),
+        proposal.rewards({ where: {beneficiary: currentAccountAddress}})
```
dkent600: What do these do then?  Can you please add a comment describing what descriptions are or are not being created, and if no subscription then where else is the subscription happening?

src/components/Scheme/SchemeProposalsPage.tsx
```diff
@@ -214,10 +213,9 @@ const SubscribedSchemeProposalsPage = withSubscription<IProps, SubscriptionData>
     const schemeId = props.scheme.id;
 
     // this query will fetch al data we need before rendering the page, so we avoid hitting the server
-    // with all separate queries for votes and stakes and stuff...
-
-    const prefetchQuery = gql`
-      query prefetchProposalDataForSchemeProposalsPage {
+    // with all separate queries for votes and stakes and rewards...
```
dkent600: Would be great to have a richer comment here about what data is being fetched and what is being subscribed-to, and what role the "separate queries" play in that story.

src/components/Proposal/ProposalData.tsx
jellegerbrandy: I've created this doc that I hope makes things ab it more clear...https://github.com/daostack/alchemy/blob/01b911ba4ba4be5ec3da0138f7ab84098fdbdf10/docs/fetchingandsubscribingdata.md

src/components/Scheme/SchemeProposalsPage.tsx
jellegerbrandy: Please check if this helps: https://github.com/daostack/alchemy/blob/01b911ba4ba4be5ec3da0138f7ab84098fdbdf10/docs/fetchingandsubscribingdata.md

docs/fetchingandsubscribingdata.md
```diff
@@ -0,0 +1,21 @@
...
+2. If the data is not in the cache, fetch the data from the server (and fill the cache)
+3. Subscribe to cache updates (if the data of the cache changes, then update the data)
+4. [optional] Subscribe to server updates of the data (i.e. open a websocket and start a server-side process that will update the cache when data on the server changes)
+
+Executing these four steps for each React component is quite expensive (just think of showing 100 proposals and executing 100 different queries and subscriptions to get the votes to render a vote component). The cost here lies in steps 2. (sending many queries over the network and wait for a response makes the application slow) and 4 (keeping 100s of subscriptions open for 100s of clients puts a considerable load on the server)
+
+So the approach we are taking in Alchemy is the following.
+
+*to avoid the cost of step 2.:*
+Fetch the data that is used by many subcomponents in a single query, before those components are rendered. When those subcomponents are rendered, they will be able to get their data directly  from the cache, and so will avoid a server request in step 2. An example of that pattern is found in the `DaoHistoryPage.tsx`
```
dkent600: Can we be clear here that the improvement is in two parts:  1) using data from the cache rather than going to the server, and 2) reducing the number of queries used to fetch data from the server into the cache.

docs/fetchingandsubscribingdata.md
```diff
@@ -0,0 +1,21 @@
...
+
+*To avoid the cost of step 4:*
+The line `arcSettings.graphqlSubscribeToQueries = false;` overrides the client default behavior to open a subscription for each query. So, by default, step 4 is skipped. Instead, we insert, in strategic places, subscriptions to relatively big queries, that will update the cache only for the data that we need.
```
dkent600: The line `arcSettings.graphqlSubscribeToQueries = false;` , **which Alchemy does at client initialization**, overrides the client...

Instead, we insert, in strategic places, subscriptions ~to relatively big queries,~ that will update the cache only for the data that we need.   [I don't think that being "relatively big" is relevant to point 4]

docs/fetchingandsubscribingdata.md
```diff
@@ -0,0 +1,21 @@
+# Fetching data with apollo and subscribing to updates
```
dkent600: I think much of this can (soon) go into the client documentation, that it will include code snippets, and it will be sibling to docs about the general use of the API to fetch data/entities.  Alchemy-specific stuff would of course remain in the Alchemy docs.

PR #1176, opened 2019-10-16
Fixthefix

jellegerbrandy:
fix  a bug introduced by previous hotfix

### DIFF
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -222,36 +222,11 @@ export async function initializeArc(provider?: any): Promise<boolean> {
     const readonly = typeof provider === "string";
 
     // get contract information from the subgraph
-    arc = new Arc(arcSettings);
-    const contractInfos = await arc.fetchContractInfos();
-    success = !!contractInfos;
-
-    if (success) {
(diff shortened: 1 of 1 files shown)


src/arc.ts
```diff
@@ -224,6 +224,7 @@ export async function initializeArc(provider?: any): Promise<boolean> {
     // get contract information from the subgraph
     if ((window as any).arc) {
       arc = (window as any).arc;
+      arc.web3 = new Web3(provider);
```
dkent600: The network can change.  Is this sufficient if you haven't also updated:

```
graphqlHttpProvider: arcSettings.graphqlHttpProvider ,
graphqlWsProvider: arcSettings.graphqlWsProvider,
```

jellegerbrandy: Perhaps I misunderstand, but no, the network should not ever change - the network is set on starting the app with an environment variable. E.g. if we are on main (or on rinkeby) we will always stay on main (or on rinkeby).

dkent600: Yeah I just remembered that we just recently coded the app to disallow using an "incorrect" network, so indeed the network effectively cannot change.

PR #1177, opened 2019-10-17
Add ENS Public Resolver to known generic scheme contracts

tibetsprague:
We may still want to do more specific validation for these function calls


tibetsprague:
@jellegerbrandy don't i need the DAO to be setup with a Generic Scheme for testing the PublicResolver to do that?

jellegerbrandy:
Yes, indeed. We have a "nectar DAO"  in the test image, which has 3 genericschemes set up, two of them are "taken"  by the ENS and registrylookup, the third is still aviable for testing.

the available one calls 0xe7410170f87102df0055eb195163a03b7f2bff4a 
http://127.0.0.1:8000/subgraphs/name/daostack/graphql?query=%7B%20daos%20(where%3A%20%7Bname%3A%20%22Nectar%20DAO%22%7D)%20%7B%20id%20schemes%20%7B%20name%20genericSchemeParams%20%7B%20contractToCall%20%7D%20%7D%20%7D%7D

jellegerbrandy:
i added the publicresolver test addresss, so in yr local setup the new scheme is here: http://127.0.0.1:3000/dao/0xfc0f7179f3a8368eb45cf6e928abc253e8712fff/scheme/0xabe41ec4340aff84c4649b3f7d9553833e7799a84ca3aaf503ce99b3c66776ba

PR #1180, opened 2019-10-18
Proposals with tags

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=userstory/1691/silent

**Issues/notes**:
- third-party tag control is from [here](https://github.com/prakhar1989/react-tags)
- for matching suggestions, I have coded it to match non-case-sensitively and show a list of suggestions all in lowercase.
- to work-around the maximum number of proposals, 100, returned by a query, suggested tags and tag counts will be derived from all DAOs, not just the current DAO.  As a result of the workaround, there will be imposed by subgraph a maximum number of unique tags that we can support, 1000, summed across all DAOs.
- I have disabled the ability to move tags with drag/drop because we are not prepared at this time to persist the tab order
- I am not creating a subscription when obtaining the list of existing tags, so the available list of suggestions will not change unless you refresh the page
- ~~when there are a lot of suggestions, a scrollbar appears in the suggestions dropdown, but you can't scroll it, though you can move the selection with up/down arrow keys (this is a problem with the control that I've not yet found a work-around for.  the easiest work-around would be to limit the number of suggestions so that a scrollbar is not needed)~~ [can no longer reproduce]
- ~~there is a bug in the control where if you add a tag by typing it, then paste the same tag, it will add the tag again (creating a dup)~~ [can no longer reproduce]


package.json
```diff
@@ -103,6 +103,8 @@
     "react": "^16.8.0",
     "react-autosuggest": "^9.4.3",
     "react-breadcrumbs-dynamic": "^1.2.0",
+    "react-dnd": "^5.0.0",
+    "react-dnd-html5-backend": "^3.0.2",
```
jellegerbrandy: these deps are not explicitly imported anywere in the code - do we need them?

src/components/Proposal/Create/SchemeForms/TagsSelector.tsx
```diff
@@ -0,0 +1,168 @@
...
+        handleDelete={this.handleDelete()}
+        handleAddition={this.handleAddition()}
+        handleDrag={this.handleDrag()}
+        delimiters={delimiters}
+        autocomplete={1}
+        readOnly={!!readOnly}
+        autofocus={false}
+        allowDragDrop={false} // because we have no way to persist the tab order
+        // eslint-disable-next-line react/jsx-no-bind
+        renderSuggestion = {( tag: ITagEx ) => <div>{tag.text} <span className="count">({tag.count})</span></div>}
```
jellegerbrandy: it is RED, mus it be red?

package.json
dkent600: Yes, they are required by the tag input control.  If these are missing then the app won't compile.

src/components/Proposal/Create/SchemeForms/TagsSelector.tsx
dkent600: It is the spec.  There is an open question about that for @A-Zak

dkent600:
@jellegerbrandy FYI I just added handling for blur

src/components/Proposal/Create/SchemeForms/TagsSelector.tsx
dkent600: done

src/components/Proposal/Create/SchemeForms/TagsSelector.tsx
```diff
@@ -0,0 +1,168 @@
...
+  createObservable: (props: IExternalProps) => {
+    if (!props.daoAvatarAddress) {
+      return from([]);
+    }
+    const arc = getArc();
+    const dao = arc.dao(props.daoAvatarAddress);
+    /**
+     * returns an array of arrays of tag names, one array per proposals
+     */
+    const tags = dao.proposals({}, { subscribe: false }).pipe(
```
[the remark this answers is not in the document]
dkent600: Done

jellegerbrandy:
* the boxesare not of the same height now, which give a sfort of restless feel:

image

not sure if that should be onsidered blocking, @A-Zak what do you say?

* The suggestions are not working for me; I get no suggestions at all when typing exiting keywords (I have seen this work before)

dkent600:
@jellegerbrandy 

> The suggestions are not working for me; I get no suggestions at all when typing exiting keywords (I have seen this work before)

I just tried it and it works fine.  Can you give more details?

jellegerbrandy:
> I just tried it and it works fine. Can you give more details?

Tried again and it is working now for me as well :-)

PR #1188, opened 2019-10-23
nervous account menu: drop down instead of left

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1748/silent

Solution is to drop the menu down rather than from the side.  I think this is better UX anyway, since the trigger for the menu is at the top, not on the right.

### DIFF
--- a/src/layouts/App.scss
+++ b/src/layouts/App.scss
@@ -202,16 +202,18 @@ body {
 
     .wallet {
       position: absolute;
-      top: 43px;
-      width: 0px;
-      opacity: 0;
-      transition: all .25s ease;
+      width: 270px;
+      top: 50px;
+      transform: scaleY(0);    
+      transform-origin: top;
+      transition: transform 0.25s ease;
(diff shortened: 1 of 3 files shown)


src/layouts/App.scss
```diff
@@ -434,13 +436,15 @@ body {
   .accountInfoContainer:hover{
     background-color: rgba(10, 28, 61, 1.000);
 
-    .wallet {
-      width: 270px;
-      opacity: 1;
-      box-shadow: 0 2px 16px 0 rgba(133, 168, 208, 0.27);
+    + .wallet {
+    transform: scaleY(1);
```
tibetsprague: weird indentation

PR #1191, opened 2019-10-24
[WIP] connection not open on send?

jellegerbrandy:
- better error propagation of (some) connection problems

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -40,6 +40,7 @@ export class App extends React.Component<{}, {
       .catch ((err): void => {
         // eslint-disable-next-line no-console
         console.log(err);
+        this.setState(() => {throw err;});
       });
 
     let GOOGLE_ANALYTICS_ID: string;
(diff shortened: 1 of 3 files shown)


src/components/Shared/withSubscription.tsx
```diff
@@ -110,11 +110,12 @@ const withSubscription = <Props extends ISubscriptionProps<ObservableType>, Obse
...
           // eslint-disable-next-line no-console
           console.error(getDisplayName(wrappedComponent), "Error in subscription", error);
-
-          this.setState({
-            isLoading: false,
-            error,
-          });
+          // this trick will effectively propagate the error to the error page (its magic!)
+          this.setState(() => { throw error; });
+          // this.setState({
```
dkent600: remove commented code?

src/layouts/Header.tsx
```diff
@@ -84,6 +84,8 @@ class Header extends React.Component<IProps, IStateProps> {
     enableWalletProvider({
       suppressNotifyOnSuccess: true,
       showNotification: this.props.showNotification,
+    }).catch((err) => {
```
dkent600: `enableWalletProvider` never throws an exception.  Instead it shows a notification and returns false (so we are already feeding back to the user)

src/App.tsx
```diff
@@ -40,6 +40,7 @@ export class App extends React.Component<{}, {
       .catch ((err): void => {
```
dkent600: `initializeArc` doesn't throw exceptions.  It returns false if it fails.  Note the `while` loop.

jellegerbrandy:
this PR is not relevant anymore - the behavior was improved in anoth PR #1213

PR #1196, opened 2019-10-27
hotfix for #1195

### LINKED ISSUE #1195: cannot vote (alchemy think user already voted) on all proposals.
switching mm to rinkeby and back to mainnet + refresh the page solve the issue .

jellegerbrandy:
fixes #1195 
Need to find the real cause - followup is in target process

### DIFF
--- a/src/components/Proposal/ProposalCard.tsx
+++ b/src/components/Proposal/ProposalCard.tsx
@@ -57,8 +57,12 @@ export default class ProposalCard extends React.Component<IProps, null> {
         let currentAccountVote = 0;
 
         let currentVote: Vote;
-        if (votes.length > 0) {
-          currentVote = votes[0];
+
+        // TODO: the next line, is a hotfix for a  which filters the votes, should not be necessary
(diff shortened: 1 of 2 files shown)


src/components/Proposal/ProposalCard.tsx
```diff
@@ -57,8 +57,12 @@ export default class ProposalCard extends React.Component<IProps, null> {
...
 
         let currentVote: Vote;
-        if (votes.length > 0) {
-          currentVote = votes[0];
+
+        // TODO: the next line, is a hotfix for a  which filters the votes, should not be necessary
+        // https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1766
+        const currentAccountVotes = votes.filter((v: Vote) => v.staticState.voter === currentAccountAddress);
+        if (currentAccountVotes.length > 0) {
+          currentVote = currentAccountVotes[0];
```
dkent600: A little misleading this is called `currentAccountVotes` rather than `currentAccountVote`  since there can only be one vote per account.

PR #1198, opened 2019-10-29
[WIP] 3box for identity

tibetsprague:


src/components/Account/AccountPopup.tsx
```diff
@@ -70,15 +69,15 @@ class AccountPopup extends React.Component<IProps, null> {
...
         <div className={css.avatar}>
-          <AccountImage accountAddress={accountAddress} />
+          <AccountImage accountAddress={accountAddress} profile={profile} />
         </div>
         <div className={css.accountInfo}>
           <div className={css.name}><AccountProfileName accountAddress={accountAddress} accountProfile={profile} daoAvatarAddress={daoState.address} /></div>
           {!profile || Object.keys(profile.socialURLs).length === 0 ? "No social profiles" :
             <div>
-              <OAuthLogin editing={false} provider="facebook" accountAddress={accountAddress} profile={profile} />
+{/*              <OAuthLogin editing={false} provider="facebook" accountAddress={accountAddress} profile={profile} />
```
jellegerbrandy: we can remove this

src/components/Shared/UserSearchField.tsx
```diff
@@ -97,7 +97,7 @@ class UserSearchField extends React.Component<IProps, IState> {
     return (
       <span>
         <span className={css.suggestionAvatar}>
-          <AccountImage accountAddress={suggestion.ethereumAccountAddress} />
+          <AccountImage accountAddress={suggestion.ethereumAccountAddress} profile={suggestion} />
```
jellegerbrandy: just a question: where do I see the effect of this? (I see no image in the contributionreward selection of beneficiary)

src/reducers/profilesReducer.ts
```diff
@@ -27,22 +22,22 @@ export function newProfile(ethereumAccountAddress: string): IProfileState {
...
-  }
-  if (dbProfile.githubURL) {
-    reduxProfile.socialURLs.github = dbProfile.githubURL;
-  }
-  if (dbProfile.twitterURL) {
-    reduxProfile.socialURLs.twitter = dbProfile.twitterURL;
-  }
-  return reduxProfile;
-}
+// export function profileDbToRedux(dbProfile: any) {
```
jellegerbrandy: remove..

jellegerbrandy:
I've been trying to play with it, here are my results:

- clicking on the "follow" button (on the dao card)  I get `RPC Error: "message" must be a nonempty string. `when i click on follow
- the data seems to be saved (after a while, indeed), the button is updated to "following" , the console messages are re-assuring, so that is good but the button behavior is a bit flaky:
  - after reload, all buttons that were "following" are reset to "follow" again
  - if I follow a second DAO, the status of the first dao is set to "follow" again
- the first time I click on "follow", and ipfs server is started (!!!). This is worrying me for 2 reasons (a) ipfs can potentially generate a lot of traffic (in the websocket, look for "p2p" under network and then choose "messages") and (b) ipfs is a heavy dependency... would be good to see if that is strictly necessary...

jellegerbrandy:
two more things:
- when saving profile, ni console log: `inpage.js:1 MetaMask - RPC Error: The method 3id_getLink does not exist/is not available {code: -32601, message: "The method 3id_getLink does not exist/is not available"}`
- can we show a notification immediately on submit, something like "saving profile, please be patient..."?

oed:
@jellegerbrandy 
> when saving profile, ni console log: inpage.js:1 MetaMask - RPC Error: The method 3id_getLink does not exist/is not available {code: -32601, message: "The method 3id_getLink does not exist/is not available"}

This is an error on the 3box end. It's safe to ignore, but will not be present in the next version of 3box.

PR #1204, opened 2019-10-31
Show accurate* queued proposals count

tibetsprague:
*This doesn't work on dev because the count from the subgraph doesn't exclude timed out proposals. On prod it should work do to the bot which executes timed out proposals.

https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1734/silent


jellegerbrandy:
I thinkthe query for "queued proposals"  on the scheme page should be rewritten to  fetch _all_ queued proposlals (and not filter by expiry date). In that way, the numbers will match on dev as well.

(So if the bot is not wokring, everything will still remain consistent - the only thing that happens is that expired proposals will remain visible in the queue)

jellegerbrandy:
@tibetsprague you saw my comment above?

tibetsprague:
@jellegerbrandy, the question is should i put effort in to make the cards look ok and show that they are expired and have them show an execute button? i guess ill try to do that quickly

jellegerbrandy:
yes :-)

tibetsprague:
@jellegerbrandy check this one out again

PR #1205, opened 2019-10-31
Block redeem when can't be paid

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1312/silent

I think the UX is a little weird in the case when some redemptions are possible and some aren't.  In that case the Redeem button will not disappear after redeeming.  All that changes is the tooltip over the Redeem button (perhaps only after refreshing the page) and when clicking the Redeem button _again_ you get a failure notification (until the DAO is able to pay out).  I think it would be better if we give the user a warning notification after running Redeem the first time to inform the user that the Redeem did not completely succeed.  @jellegerbrandy 

* shows failure notification when the user tries to redeem but the DAO can't pay out at least part of it
* added ability to determine whether externalToken and GEN rewards can be paid
* cleaned up code to make it more clear  **I recommend you look carefully at the changes.  I commented a lot of code and gave meaningful variable names.  I would like second eyes on whether I have made the right assumptions**
* removed the ugly "-" in the title of the preproposal modal

### DIFF
--- a/src/components/Proposal/ActionButton.tsx
+++ b/src/components/Proposal/ActionButton.tsx
@@ -1,4 +1,4 @@
-import { Address, IDAOState, IProposalStage, IProposalState, IRewardState, Token } from "@daostack/client";
+import { Address, IDAOState, IProposalStage, IProposalState, IRewardState, Token, IProposalOutcome } from "@daostack/client";
 import { executeProposal, redeemProposal } from "actions/arcActions";
 import { enableWalletProvider, getArc } from "arc";
 import * as classNames from "classnames";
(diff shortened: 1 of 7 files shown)


src/components/Proposal/ActionButton.tsx
```diff
@@ -217,4 +243,33 @@ class ActionButton extends React.Component<IProps, IState> {
   }
 }
 
-export default connect(mapStateToProps, mapDispatchToProps)(ActionButton);
+const SubscribedActionButton = withSubscription({
```
dkent600: @jellegerbrandy  @tibetsprague Should the gen and externalToken amounts be computed in `ProposalData`?

dkent600: To answer my own question:  I don't think so because `ActionButton` is used outside of `ProposalData`, in `RedemptionsMenu`

dkent600: In fact I think an argument could be made that the computation of ethBalance should be moved from ProposalData and RedemptionsMenu into the ActionButton.

dkent600:
@jellegerbrandy 

> i was thinking that it is not so nice t [....]

As I mentioned in the PR description, I think we should also warn the user in the case that some rewards can be redeemed, and others cannot.

PR #1217, opened 2019-11-06
3box feed

tibetsprague:


.travis.yml
```diff
@@ -30,7 +30,6 @@ script:
 
   # show the daos that are indexed
   - ./scripts/checkDaos.sh
-  - echo -en 'travis_fold:end:script.4\\r'
```
jellegerbrandy: this should not get removed!

app.json
```diff
@@ -3,9 +3,6 @@
   "description": "",
   "scripts": {},
   "env": {
-    "API_URL": {
```
jellegerbrandy: also this should not be removed

.travis.yml
tibetsprague: that's weird, must have been a weird merge? I didn't do that

app.json
tibetsprague: same here

tibetsprague: wait, actually we can remove this, we dont use API_URL anymore

tibetsprague:
> Seems to me one might not care about one's personal feed and ought to be able to hide the "Excuse me, who are you?" stuff and just see the list of DAOs.

@A-Zak is working on a better home page banner when not logged in

tibetsprague:
@A-Zak can you look at all of @dkent600 comments? perhaps we should have a design review... even though its pretty late in the process here

jellegerbrandy:
I have not been able to break the review app, so that is good :-)

Just a few remarks:

- I cannot find the "follow"  button on the user profile
- The onload of the feed is a bit confusing, as it statts as the "you are not ofllowing anything" and only then loads the feed (which can take some time). Perhaps start with a spinner isntead of the empty feed/
- Agree with @dkent600 's observations - this could use a design review.

jellegerbrandy:
Some usability issues that I have:

- It is difficult to unfollow (and hard to see from the feed _why_ an item turns up). [possible solution: add an "unfollow XXX" button tot he feed items]
- It is difficult to find the feed again after you have navigated away. (Consider also users that are always navigating to a dao page -- they will never even see the feed)
- (as doug noted) once your feed is configured, it is very hard to find how to find information outside of your feed (i.e.the dao list)
- the feed list itself: we probably want to make it clearer that this is an EVENT feed (and not, say, a list of proposal descriptions with lots of duplictes). That means more emphasis on what happend (i.e. "Genesis Alpha has a new proposal" should either be highlighted more, or the amount of informatino on the card should be diminished)

PR #1218, opened 2019-11-07
Add Training Tooltips - Phase I

dkent600:
Addresses: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=userstory/1797/silent

This simply adds tooltips that show up on mouse hover.

**Note** it is lacking the dynamically-computed boosted&preboosted times.

### DIFF
--- a/src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
+++ b/src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
@@ -10,6 +10,7 @@ import { connect } from "react-redux";
 import Select from "react-select";
 import { showNotification } from "reducers/notifications";
 import TagsSelector from "components/Proposal/Create/SchemeForms/TagsSelector";
+import TrainingTooltip from "components/Shared/TrainingTooltip";
 import * as css from "../CreateProposal.scss";
 import MarkdownField from "./MarkdownField";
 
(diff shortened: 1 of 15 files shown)


src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
```diff
@@ -229,12 +234,14 @@ class CreateContributionReward extends React.Component<IProps, IStateProps> {
                 className={touched.title && errors.title ? css.error : null}
               />
 
-              <label htmlFor="descriptionInput">
+              <TrainingTooltip overlay={fnDescription} placement="right">
```
tibetsprague: why is this a function?

src/components/Shared/TrainingTooltip.tsx
```diff
@@ -0,0 +1,21 @@
+import {default as Tooltip, RCTooltip} from "rc-tooltip";
+import * as React from "react";
+// import * as css from "./TrainingTooltip.scss";
```
tibetsprague: remove

src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
dkent600: for lint and because it needs to return a component.

Issue #1181, opened 2019-10-21
Scheme URL Params => Proposal Templates


dOrgJelli:
Add the ability to open a proposal template for a user via embedding its pre-filled-out contents within the URL's parameters.

For example:
* User clicks a hyperlink that says "Create A Proposal" on an external webpage.
* User is redirected to alchemy.daostack.io/0xDAO/0xSCHEME?title=...&description=...&...
* When the page loads, the scheme's proposal form is immediately shown with all predefined content inserted. This content is taken from the URL parameters (see bullet above).

Creating these URLs could be done through adding an "export / embed" button on the proposal form. This way any Alchemy user could fill out some proposal content, then export the serialized URL + params and add it as a hyperlink somewhere else.

dOrgJelli:
dOrg would like to use this feature to create proposal templates for the following types of proposals:
* Project Payouts
* Bounties
* Legal Document DAO-Signing
* Reimbursements

PR #1226, opened 2019-11-11
Scheme URL Params => Proposal Templates

mrrobot16:
This PR is for issue: https://github.com/daostack/alchemy/issues/1181

Currently we can now load a proposal by passing params to /dao/0x.../scheme/0x../proposals/create?title=title&description=description. TODO: Export form values to shareable/usable proposal template url


mrrobot16:
We can load proposal data thru parms and also export data from a proposal form to shareable URL. 
What should do with this shareable url?

jellegerbrandy:
Thanks @mrrobot16 
- could you please add a simple test along the lines of the tests in test/integration?  It does not have to be very sophisticated, just a test that opens the proposal form and then checks if the form fields are filled in correctly
- can you fix the lint errors? Run `npm lint`. The build should pass on travis:https://travis-ci.org/daostack/alchemy/builds/610177030?utm_source=github_status&utm_medium=notification

src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
```diff
@@ -150,17 +217,7 @@ class CreateContributionReward extends React.Component<IProps, IStateProps> {
...
-            description: "",
-            ethReward: 0,
-            externalTokenAddress: arc.GENToken().address,
-            externalTokenReward: 0,
-            nativeTokenReward: 0,
-            reputationReward: 0,
-            title: "",
-            url: "",
-          } as IFormValues}
+          initialValues={this.state.initialFormValues}
```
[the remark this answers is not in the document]
mrrobot16: Great idea :) Thanks!

src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
```diff
@@ -97,10 +99,37 @@ class CreateContributionReward extends React.Component<IProps, IStateProps> {
     this.handleSubmit = this.handleSubmit.bind(this);
     this.state = { 
       tags: new Array<string>(),
+      initialFormValues: {
```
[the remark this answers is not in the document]
mrrobot16: Done.

mrrobot16:
question: @tibetsprague @jellegerbrandy

Is there a way I can access Formik form inputted values outside of the handleSubmit() look at https://github.com/daostack/alchemy/blob/e553f75f04a448bf33843c8e18c17b82303ca2ab/src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx#L109.

The reason this test fails https://travis-ci.org/daostack/alchemy/builds/612541109#L3892 is because the first approach done to export form values was being done inside the handleSubmit() meaning that I called a exportFormValues() function inside the handleSubmit() if a conditional is true(exportMode: true) if not it will do what handleSubmit does originally which is grab form values and submit proposals.

So potential solution i am thinking is if we can some how access all form values outside of the submit handler and just have another function do the exportFormValues logic will make the test pass in my opinion.

I was looking in Formik's documentation but could not found a way to access form values outside the submit handler, any ideas?

There other solution I thought was to set the form inputs to the state but I don't feel that's a good idea. 

Thoughts?

tibetsprague:
@mrrobot16  there are a number of suggested techniques for this here: https://github.com/jaredpalmer/formik/issues/214 that may help...

mrrobot16:
@tibetsprague @jellegerbrandy thanks for your input guys it's now working. Now need to add a test, feel free to check it out and give feedback :)

src/components/Shared/UserSearchField.tsx
```diff
@@ -17,6 +17,7 @@ interface IExternalProps {
   name?: string;
   onBlur?: (touched: boolean) => any;
   onChange?: (newValue: string) => any;
+  beneficiary: string;
```
jellegerbrandy: perhaps no name this "beneficiary", but rather "defaultValue" (as the usersarchfield is not always used for "beneificiaries")

mrrobot16: Good idea thanks :)

mrrobot16: Done.

jellegerbrandy:
hey @mrrobot16, I really like this feature but you seem to be ignoring some of my comments :-)

- a test would be really nice
- please do not use this https://github.com/daostack/alchemy/pull/1226/files/d4e08f8c28ae564ed9417f8134b9fe1fbbb66a95#diff-b6f590277383b310441becd4524648b0R195 but use some existing tested code. The reason for this is that we'd like to catch any error in the URL syntax in that line, and not fail in some arbitrary place later, or worse, try do do stuff with the invalid input. (For example, innocent example: pass something like `.../create/?&`)

jellegerbrandy:
@A-Zak were you involved with the UI specs of this? This seems like a very bad place (and coloring) for "exporting the URL":

image

1. Bad because it is a very specialized advanced feature but right next (even before) the "sumbit" button, and with a "stronger" color then the submit button, suggesting this is the action the user shoudl take here.
2. Bad because all our buttons are actions doing stuff (, while this is just copying something to the clipboard

Why is is on the create form, and not on each proposal? Should we have this export functionality at all? who is it for?

A-Zak:
> * Bad because it is a very specialized advanced feature but right next (even before) the "sumbit" button, and with a "stronger" color then the submit button, suggesting this is the action the user shoudl take here.
> * Bad because all our buttons are actions doing stuff (, while this is just copying something to the clipboard

Yeah..... That's not the best place for this button... /:
Creating a template is something that will probably be used very rarely, so it shouldn't be so prominent, and like you said @jellegerbrandy, definitely not more prominent than the submit button... 

How bout if we put a triple dot menu on the right side of the proposal top bar (like we have on the proposal card)... and just have a little menu that has "Export proposal template" button...

A-Zak:
Also, does this work for generic scheme proposals? e.g. the [DutchX controlling scheme](https://alchemy.daostack.io/dao/0x519b70055af55a007110b4ff99b0ea33071c720a/scheme/0xeca5415360191a29f12e1da442b9b050adf22c81b08230f1dafba908767e604f/proposals/create/)? We need to support both custom fields (should be easy, right?), but also selecting the "proposal type" (the method to call)...

mrrobot16:
> hey @mrrobot16, I really like this feature but you seem to be ignoring some of my comments :-)
> 
> * a test would be really nice
> * please do not use this https://github.com/daostack/alchemy/pull/1226/files/d4e08f8c28ae564ed9417f8134b9fe1fbbb66a95#diff-b6f590277383b310441becd4524648b0R195 but use some existing tested code. The reason for this is that we'd like to catch any error in the URL syntax in that line, and not fail in some arbitrary place later, or worse, try do do stuff with the invalid input. (For example, innocent example: pass something like `.../create/?&`)

@jellegerbrandy Yes I am aware of this changes suggested, I am currently working on it as I get time. I have not forgotten this.

@A-Zak As far as UI, we didnt not have any specified way as far were this action button will live. I took initiative and placed there inside of create Proposal form because I didn't know where else to test it. Was that discussed somewhere I missed? Please let me know :) 

This currently works for Contribution Reward Scheme not for all other schemes yet.

Please let me know if anything else :)

dOrgJelli:
@mrrobot16 your utility function can be generalized like so:
```typescript
function getInitialFormValues<Values>(defaultValues: Values) {
  const { search } = window.location;
  const params = new URLSearchParams(search);

  // Warning, seem comment 1. below
  const initialFormValues: any = { ...defaultValues };

  for (const prop in defaultValues) {
    const paramValue = params.get(prop);

    if (paramValue) {
      switch (typeof defaultValues[prop]) {
        case "string":
          initialFormValues[prop] = String(paramValue);
          break;
        case "number":
          initialFormValues[prop] = Number(paramValue);
          break;
        case "boolean":
          initialFormValues[prop] = Boolean(paramValue);
          break;
        case "bigint":
          initialFormValues[prop] = BigInt(paramValue);
          break;
        case "object":
          initialFormValues[prop] = JSON.parse(paramValue);
          break;
        case "undefined":
        case "function":
        case "symbol":
          throw Error(`Unsupported default value ${defaultValues[prop]}`);
      }
    }
  }

  // 1. Warning, this is not a deep clone, only shallow.
  // Will be problematic if Values has a nested object,
  // as it will get overwritten. Use something like lodash
  // to do a deep clone here.
}
```

Feel free to message me for details on how / why this works :smiley:

mrrobot16:
**UPDATE**: This is is still **WIP**

mrrobot16:
@jellegerbrandy @tibetsprague Hey guys i have done changes and added tests, when possible can we get feedback on this latest changes. It currently works for all schemes now.

src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
```diff
@@ -145,22 +156,23 @@ class CreateContributionReward extends React.Component<IProps, IStateProps> {
     const arc = getArc();
 
     const fnDescription = () => (<span>Short description of the proposal.<ul><li>What are you proposing to do?</li><li>Why is it important?</li><li>How much will it cost the DAO?</li><li>When do you plan to deliver the work?</li></ul></span>);
+    const initialFormValues: IFormValues  = {
```
tibetsprague: actually now that i think about it would be better to set this up in the constructor as this.initialFormValues so it doesn't have to get recreated on every render... especially since you are loading arc for it every time. not a dealbreaker though

src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
```diff
@@ -290,6 +302,7 @@ class CreateContributionReward extends React.Component<IProps, IStateProps> {
                     name="beneficiary"
                     onBlur={(touched) => { setFieldTouched("beneficiary", touched); }}
                     onChange={(newValue) => { setFieldValue("beneficiary", newValue); }}
+                    defaultValue={importUrlValues(initialFormValues).beneficiary}
```
tibetsprague: if you do the importUrlValues when you setup initialFormValues one time in the constructor then you dont have to call it multiple times

src/components/Proposal/Create/SchemeForms/CreateKnownGenericSchemeProposal.tsx
```diff
@@ -61,10 +68,11 @@ class CreateKnownSchemeProposal extends React.Component<IProps, IState> {
     }
 
     const actions = props.genericSchemeInfo.actions();
+    const initialActionId = importUrlValues({ currentActionId: ""}).currentActionId;
```
tibetsprague: why not just call importUrlValues once for the whole state here and then you dont have to call it three times?

src/components/Proposal/Create/SchemeForms/CreateKnownGenericSchemeProposal.tsx
```diff
@@ -251,7 +268,7 @@ class CreateKnownSchemeProposal extends React.Component<IProps, IState> {
 
         <div className={css.formWrapper}>
           <Formik
-            initialValues={initialFormValues}
+            initialValues={importUrlValues(initialFormValues)}
```
tibetsprague: if done once in the constructor dont have to import again here

src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
```diff
@@ -64,11 +68,13 @@ class CreateSchemeRegistrarProposal extends React.Component<IProps, IState> {
     this.handleSubmit = this.handleSubmit.bind(this);
 
     this.state = {
-      currentTab: "addScheme",
+      currentTab: this.loadCurrentTab(),
       tags: new Array<string>(),
     };
+    this.state = importUrlValues(this.state);
```
tibetsprague: yeah do this everywhere

src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
```diff
@@ -121,8 +126,51 @@ class CreateSchemeRegistrarProposal extends React.Component<IProps, IState> {
...
+    values = {
+      ...values,
+      ...this.state,
+    };
+    exportUrl(values);
+    this.props.showNotification(NotificationStatus.Success, "Exportable url is now in clipboard :)");
+  }
+  
+  loadCurrentTab(){
+    let initialCurrentTab = importUrlValues({ schemeToAdd: "", schemeToEdit: "", schemeToRemove: ""});
```
tibetsprague: this seems like a strange way to specify the tab, why do you need three different params which override each other? why not just a single tab=XXX parameter?

src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
```diff
@@ -121,8 +126,51 @@ class CreateSchemeRegistrarProposal extends React.Component<IProps, IState> {
...
+    else {
+      initialCurrentTab = "addScheme";
+    }
+
+    return initialCurrentTab;
+  }
+  
   public render(): RenderOutput {
     // "schemes" are the schemes registered in this DAO
+    const initialFormValues: any = {
```
tibetsprague: again seems better to create this and call the import in the constructor so it only happens once

src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
```diff
@@ -335,7 +372,7 @@ class CreateSchemeRegistrarProposal extends React.Component<IProps, IState> {
                         name="schemeToEdit"
                         component="select"
                         className={css.schemeSelect}
-                        defaultValue=""
+                        defaultValue={importUrlValues(initialFormValues)}
```
tibetsprague: call import once in constructor

src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
```diff
@@ -410,7 +447,7 @@ class CreateSchemeRegistrarProposal extends React.Component<IProps, IState> {
                         name="schemeToRemove"
                         component="select"
                         className={css.schemeSelect}
-                        defaultValue=""
+                        defaultValue={importUrlValues(initialFormValues).schemeToRemove}
```
tibetsprague: call import once in constructor

src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
mrrobot16: because this way i am telling importUrlValues to return me only those properties( schemeToAdd, schemeToEdit, schemeToRemove)if found in search url params.

dOrgJelli: But why not just use a property named "tab" like @tibetsprague is suggesting and set it to "addScheme" | "editScheme" | "removeScheme"?

mrrobot16:
@tibetsprague I haved gone thru your comments and in indeed setting up the initial values in constructor works. 

Let me know what else can be done to optimize this feature. Happy Holidays!

src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx
```diff
@@ -1,16 +1,18 @@
+import * as React from "react";
```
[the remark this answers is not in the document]
mrrobot16: I thought it was cleaner to order the imports by npm modules, components, libs, reducers, etc.

dOrgJelli:
Any word on this @mrrobot16 @tibetsprague? Very interested in using this for some dOrg use cases!

jellegerbrandy:
Hi Guys, sorry for not getting back earlier.

As far as i'm concerned, this is good to go, except for the export button in the UI, which nobody is really happy with.

i imagine this feature would be just as useful without having an export button at all. Shall we remove it?

dOrgJelli:
I've updated the UI, please verify this looks good to you. Here's a screenshot:
image


@jellegerbrandy @mrrobot16 @A-Zak @tibetsprague

Theylon:
Hey guys, 
Could export function be added to:
- Competition proposal (for now only the overlapping parameters)
- Competition submission (Title, description, URL)

PR #1235, opened 2019-11-13
remove most arrow function and .bind lint errors

dkent600:
This is in an effort to minimize spurious react component rendering by removing most of the uses of arrow and .bind functions  in `render`. 

Down to only 11 lint warnings.

But I don't honestly see any significant performance improvements.

### DIFF
--- a/src/components/Account/AccountProfilePage.tsx
+++ b/src/components/Account/AccountProfilePage.tsx
@@ -92,7 +92,7 @@ class AccountProfilePage extends React.Component<IProps, null> {
     e.preventDefault();
   }
 
-  public handleFormSubmit = async (values: IFormValues, { _props, setSubmitting, _setErrors }: any): Promise<void>  => {
+  public handleFormSubmit = () => async (values: IFormValues, { _props, setSubmitting, _setErrors }: any): Promise<void> => {
     const { accountAddress, currentAccountAddress, showNotification, updateProfile } = this.props;
 
     if (!await enableWalletProvider({ showNotification })) { return; }
@@ -178,7 +178,7 @@ class AccountProfilePage extends React.Component<IProps, null> {
 
                 return errors;
               }}
-              onSubmit={this.handleFormSubmit}
+              onSubmit={this.handleFormSubmit()}
               // eslint-disable-next-line react/jsx-no-bind
               render={({
                 // eslint-disable-next-line @typescript-eslint/no-unused-vars
--- a/src/components/Account/AccountProfilePage.tsx
+++ b/src/components/Account/AccountProfilePage.tsx
@@ -92,7 +92,7 @@ class AccountProfilePage extends React.Component<IProps, null> {
     e.preventDefault();
   }
 
-  public handleFormSubmit = () => async (values: IFormValues, { _props, setSubmitting, _setErrors }: any): Promise<void> => {
+  public handleFormSubmit = async (values: IFormValues, { _props, setSubmitting, _setErrors }: any): Promise<void> => {
     const { accountAddress, currentAccountAddress, showNotification, updateProfile } = this.props;
 
     if (!await enableWalletProvider({ showNotification })) { return; }
@@ -178,7 +178,7 @@ class AccountProfilePage extends React.Component<IProps, null> {
 
                 return errors;
               }}
-              onSubmit={this.handleFormSubmit()}
+              onSubmit={this.handleFormSubmit}
               // eslint-disable-next-line react/jsx-no-bind
               render={({
                 // eslint-disable-next-line @typescript-eslint/no-unused-vars
--- a/src/components/Account/AccountProfilePage.tsx
+++ b/src/components/Account/AccountProfilePage.tsx
@@ -92,7 +92,7 @@ class AccountProfilePage extends React.Component<IProps, null> {
     e.preventDefault();
   }
 
-  public async handleSubmit(values: IFormValues, { _props, setSubmitting, _setErrors }: any): Promise<void> {
+  public handleFormSubmit = async (values: IFormValues, { _props, setSubmitting, _setErrors }: any): Promise<void>  => {
     const { accountAddress, currentAccountAddress, showNotification, updateProfile } = this.props;
 
     if (!await enableWalletProvider({ showNotification })) { return; }
@@ -134,7 +134,7 @@ class AccountProfilePage extends React.Component<IProps, null> {
     setSubmitting(false);
   }
 
-  public onOAuthSuccess(account: IProfileState) {
+  public onOAuthSuccess = () => (account: IProfileState): void => {
     this.props.verifySocialAccount(this.props.accountAddress, account);
   }
 
@@ -163,6 +163,7 @@ class AccountProfilePage extends React.Component<IProps, null> {
                 description: accountProfile ? accountProfile.description || "" : "",
                 name: accountProfile ? accountProfile.name || "" : "",
               } as IFormValues}
+              // eslint-disable-next-line react/jsx-no-bind
               validate={(values: IFormValues): void => {
                 // const { name } = values;
                 const errors: any = {};
@@ -177,7 +178,8 @@ class AccountProfilePage extends React.Component<IProps, null> {
 
                 return errors;
               }}
-              onSubmit={this.handleSubmit.bind(this)}
+              onSubmit={this.handleFormSubmit}
+              // eslint-disable-next-line react/jsx-no-bind
               render={({
                 // eslint-disable-next-line @typescript-eslint/no-unused-vars
                 values,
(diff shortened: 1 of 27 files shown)


src/components/Proposal/Voting/VoteButtons.tsx
```diff
@@ -91,20 +89,20 @@ class VoteButtons extends React.Component<IProps, IState> {
...
           "Voting requires reputation in " + dao.name :
-          proposal.stage === IProposalStage.ExpiredInQueue || (proposal.stage === IProposalStage.Boosted && expired) || (proposal.stage === IProposalStage.QuietEndingPeriod && expired)  || (proposal.stage === IProposalStage.Queued && expired) ?
+          proposal.stage === IProposalStage.ExpiredInQueue ||
+              (proposal.stage === IProposalStage.Boosted && expired) ||
+              (proposal.stage === IProposalStage.QuietEndingPeriod && expired)  ||
+              (proposal.stage === IProposalStage.Queued && expired) ?
             "Can't vote on expired proposals" :
             proposal.stage === IProposalStage.Executed ?
-              "Can't vote on executed proposals" :
-              `Vote ${vote === IProposalOutcome.Pass ? "for" : "against"}`
```
tibetsprague: im confused, why did you get rid of this one?

src/layouts/AppContainer.tsx
```diff
@@ -163,31 +184,9 @@ class AppContainer extends React.Component<IProps, IState> {
...
-                    timestamp={timestamp}
-                    dismiss={() => dismissNotification(id)}
-                    showNotification={showNotification}
-                    minimize={() => this.setState({notificationsMinimized: true})}
-                  />
-                </div>
-              ))
+              <MinimizedNotifications notifications={sortedNotifications.length} unminimize={this.unminimizeNotif()} />
+              :
+              sortedNotifications.map(this.sortedNotification())
```
tibetsprague: this looks weird, maybe name the function something different?

src/components/Account/AccountProfilePage.tsx
```diff
@@ -177,7 +178,8 @@ class AccountProfilePage extends React.Component<IProps, null> {
 
                 return errors;
               }}
-              onSubmit={this.handleSubmit.bind(this)}
+              onSubmit={this.handleFormSubmit()}
```
jellegerbrandy: This still creates a new function object on each render, and so I'm not sure what is the improvement here.

The pattern to use here is to ` bind` the function int he constructor, as we do for example here: https://github.com/daostack/alchemy/blob/dev/src/components/Proposal/Create/SchemeForms/CreateContributionRewardProposal.tsx#L97

(So you would undo the change you made in `handleFormSubmit`, remove the ` bind` in onSubmit, and instead bind int he constructor)

src/components/Account/AccountProfilePage.tsx
```diff
@@ -258,8 +260,8 @@ class AccountProfilePage extends React.Component<IProps, null> {
                         }
 
                         <h3>Social Verification</h3>
-                        <OAuthLogin editing={editing} provider="twitter" accountAddress={accountAddress} onSuccess={this.onOAuthSuccess.bind(this)} profile={accountProfile} socket={socket} />
-                        <OAuthLogin editing={editing} provider="github" accountAddress={accountAddress} onSuccess={this.onOAuthSuccess.bind(this)} profile={accountProfile} socket={socket} />
+                        <OAuthLogin editing={editing} provider="twitter" accountAddress={accountAddress} onSuccess={this.onOAuthSuccess()} profile={accountProfile} socket={socket} />
```
jellegerbrandy: see comment above about binding

src/components/Account/AccountProfilePage.tsx
```diff
@@ -258,8 +260,8 @@ class AccountProfilePage extends React.Component<IProps, null> {
...
+                        <OAuthLogin editing={editing} provider="github" accountAddress={accountAddress} onSuccess={this.onOAuthSuccess()} profile={accountProfile} socket={socket} />
```
jellegerbrandy: see comment above about binding

src/components/Dao/DaoContainer.tsx
```diff
@@ -80,31 +93,25 @@ class DaoContainer extends React.Component<IProps, null> {
           </div>
           <Switch>
             <Route exact path="/dao/:daoAvatarAddress/history"
-              render={(props) => <DaoHistoryPage {...props} currentAccountAddress={currentAccountAddress} />} />
+              render={this.daoHistoryRoute()} />
```
jellegerbrandy: see comment above about binding

src/components/Dao/DaoContainer.tsx
```diff
@@ -80,31 +93,25 @@ class DaoContainer extends React.Component<IProps, null> {
...
             <Route exact path="/dao/:daoAvatarAddress/members"
-              render={(props) => <DaoMembersPage {...props} daoState={daoState} />} />
+              render={this.daoMembersRoute()} />
```
jellegerbrandy: see comment above about binding

src/components/Proposal/Voting/VoteButtons.tsx
dkent600: That was a really weird piece of code.  Note that `tipContent` was a function before, taking the parameter `vote`.  Where was `vote` supposed to come from, and what did it represent?  I could find no answers so I removed it.

src/layouts/AppContainer.tsx
dkent600: done

src/components/Account/AccountProfilePage.tsx
dkent600: @tibetsprague You once advised me to use the method I'm using.  Can you please confirm @jellegerbrandy's suggestion, to make sure we're all on the same page?  Thanks!

(Just a note BTW, documenting best practices for fixing lint errors like this one is one of the documentation tasks on your plate, Tibet  :-)  )

dkent600: @jellegerbrandy   @tibetsprague  My understanding of the problem here is not that the function is being called on every render, but worse, the fact the react thinks a new function is created actually causes any subcomponents depending on it (to which the new function is given) to be rendered more often.

Reading some more:  There is a similar pattern with the creation of new instances of _anything_, like variables created in the render method, and that are then passed into subcomponents.

src/components/Proposal/Voting/VoteButtons.tsx
tibetsprague: looks pretty obvious in the code, you pass in whether it is a pass button or fail button you are mousing over and then the tip says Vote For or Vote Against. im not sure when that funcionality was removed but it used to exist. dont know if we care, no one has mentioned it.

dkent600: this text is only being used when voting is disabled, so the case you describe can't happen.

src/components/Account/AccountProfilePage.tsx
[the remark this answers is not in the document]
dkent600: The code is correct, even if it is not optimal. `this.handleFormSubmit()` returns the actual handler function, bound to this.  @tibetsprague this is the pattern you long ago recommended to me.  :-)

src/components/Account/AccountProfilePage.tsx
dkent600: no, `this.onOAuthSuccess` is correct, if not optimal.

src/components/Account/OAuthLogin.tsx
```diff
@@ -76,7 +76,7 @@ export default class OAuthLogin extends React.Component<IProps, IState> {
   // Kicks off the processes of opening the popup on the server and listening
   // to the popup. It also disables the login button so the user can not
   // attempt to login to the provider twice.
-  public startAuth(e: any) {
+  public startAuth = () => (e: any) => {
```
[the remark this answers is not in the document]
dkent600: yes, I think as long as it is an arrow function it should be automatically bound to `this` and therefore give us what we want.

src/components/Account/AccountProfilePage.tsx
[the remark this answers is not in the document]
dkent600: @tibetsprague In any case, this is a big "doh" on my part....should've realized this.  My bad.

dkent600:
@tibetsprague   @jellegerbrandy I believe I've corrected all the handler function functions.

src/components/Account/AccountProfilePage.tsx
jellegerbrandy: well, yes, indeed.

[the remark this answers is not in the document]
jellegerbrandy: So what is the conclusion here? 

Methinks we should still replace `handleFormSubmit()`  with `handleformsubmit` (and change the definition of the function and remove the first `() =>` from it.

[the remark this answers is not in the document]
dkent600: > Methinks we should still replace handleFormSubmit() with handleformsubmit (and change the definition of the function and remove the first () => from it.

That is what I've done  :-)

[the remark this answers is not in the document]
jellegerbrandy: "yes, indeed", meaning that passing `this.handleFormSubmit()` will create a new function. 

We want to just pass `handleFormSubmit` to the subcomponent, and either:

-  keep the original handleFormSubmit, and ` bind()`  it in the constructor of the component.
- or define `public hanldeFormSubmit = (values) => { doit }` (so you do not need to `bind()` the function)

or am I misunderstanding something here?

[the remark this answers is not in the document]
jellegerbrandy: exactly like you did below in `handleCloseMenu`, to me it looks like you  forgot this one.

dkent600:
@jellegerbrandy 

> exactly like you did below in handleCloseMenu, to me it looks like you forgot this one

Indeed, my global search missed a case.  I've corrected them

src/components/Account/AccountProfilePage.tsx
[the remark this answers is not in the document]
dkent600: Indeed, my global search missed a case. I've corrected them

PR #1238, opened 2019-11-13
[WIP] instantiate TrainingTooltips only when target is visible

dkent600:
uses react-visibility-sensor to enable us to instantiate TrainingTooltips only when the target html is visible.  This modestly speeds up the display of tooltips on the scheme proposals page when there are a lot of proposals.

**Note** I plan to submit another PR that will do no more that limit proposal card tooltips to the first card, visible or not.  We can choose which of these we want.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22849,6 +22849,14 @@
         "react-lifecycles-compat": "^3.0.4"
       }
     },
+    "react-visibility-sensor": {
+      "version": "5.1.1",
+      "resolved": "https://registry.npmjs.org/react-visibility-sensor/-/react-visibility-sensor-5.1.1.tgz",
+      "integrity": "sha512-cTUHqIK+zDYpeK19rzW6zF9YfT4486TIgizZW53wEZ+/GPBbK7cNS0EHyJVyHYacwFEvvHLEKfgJndbemWhB/w==",
+      "requires": {
+        "prop-types": "^15.7.2"
+      }
+    },
     "read-pkg": {
       "version": "1.1.0",
       "resolved": "https://registry.npmjs.org/read-pkg/-/read-pkg-1.1.0.tgz",
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     "react-through": "^1.1.1",
     "react-toggle": "^4.1.1",
     "react-transition-group": "^2.5.1",
+    "react-visibility-sensor": "^5.1.1",
     "redux": "^4.0.4",
     "redux-devtools-extension": "^2.13.7",
     "redux-thunk": "2.2.0",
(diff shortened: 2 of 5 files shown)


dkent600:
@jellegerbrandy 

> it should not be necessary (the tooltips are really small components and should be very quick to render, so rendering those that are not in the viewport should not add much if they are not 1000s)

It shouldn't be, but it is.  I think there may be two challenges here:

1. when the redux state changes, affecting every existing tooltip, react.js is probably doing much more that it needs to.  This is demonstrated that when I completely remove the tooltip from the non-visible proposal cards, the reaction time is still slower than a page that has fewer proposal cards.
2. when the tooltip state changes, the target element is re-rendered.  Examples are the votebuttons and stakebuttons components that are tooltip targets.  This is a function of the $@*&%! way in which the rc-tooltip component we are using is designed to wrap its target component, and react doing unnecessary renders of the wrapped component.  Eliminating this is how I'm gaining the performance improvement (see the best version fo this in a separate PR:  : https://alchemy-staging-rinkeb-pr-1240.herokuapp.com/dao/0x72939947e7a1c4ac94bb840e3304b322237ad1a8/scheme/0x16b81c56297ae81ffa32ed128f17dbf7b108eb9200d2c8110a9ff2e05e095995)

> it does not address the probelm of having redundant tooltips in the viewport

That is a separate issue that should be addressed separately

> It seems that this solution puts some load on our already-slow app: each time we scroll or resize, the state of all tooltips changes, so they are all re-rendered

Yes, the parts that are targets to tooltips are re-rendered when the card becomes visible (votebuttons and stakebuttons)

jellegerbrandy:
Ok, thanks for the explanation, it's clearer to me now why we need this :-)

Just one more thing:

> That is a separate issue that should be addressed separately

I think not really. If we show the tooltips on toggle only on the first proposal card, we have far far less tooltips, and all this work on optimizing would perhaps not be necessary.

dkent600:
@jellegerbrandy 

>> That is a separate issue that should be addressed separately

> I think not really. If we show the tooltips on toggle only on the first proposal card, we have far far less tooltips, and all this work on optimizing would perhaps not be necessary.

Yeah I thought that after I wrote the comment.  Thing is, with the easy way we are left with showing tooltips on a card that is scrolled off and thus not visible to the user, which defeats the purpose of the feature.

jellegerbrandy:
> Yeah I thought that after I wrote the comment. Thing is, with the easy way we are left with showing tooltips on a card that is scrolled off and thus not visible to the user, which defeats the purpose of the feature.

i agree. It is not perfect, but @A-Zak was ok with this solution, and it is much more straightforward, so I'd be for taking that approach

dkent600:
closing because this is inferior to https://github.com/daostack/alchemy/pull/1240

dkent600:
@jellegerbrandy 

> i agree. It is not perfect,  but @A-Zak was ok with this solution

He preferred this option to that one.

PR #1240, opened 2019-11-14
[WIP] instantiate TrainingTooltips only when target is visible II

dkent600:
This is like https://github.com/daostack/alchemy/pull/1238 but rather than hiding the rc-tooltip inside of TrainingTooltip, leaving all of the TrainingTooltip components existent across the app,  it entirely hides the whole TrainingTooltip on proposal cards that aren't visible.

This method is by far the fastest.


dkent600:
Closing as we are deciding to only show tooltips on the first proposal card.

Issue #1171, opened 2019-10-15
125 % of rep ?


orenyodfat:
<img width="1075" alt="Screen Shot 2019-10-15 at 14 48 35" src="https://user-images.githubusercontent.com/5994398/66829057-31518800-ef5b-11e9-8cd0-acef905e8099.png">

orenyodfat:
- I was the only rep holder in the dao .
- created a proposal 
- vote 
- redeem rep.

jellegerbrandy:
which dao?

orenyodfat:
Necdao

PR #1245, opened 2019-11-16
update DAO total reputation supply when it changes

dkent600:
Resolves:  https://github.com/daostack/alchemy/issues/1171
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1749/silent

* Fixes the 125% issue in the original use case as well as everywhere else where the reputation percentage is displayed.  Accomplished this by subscribing to changes in the DAOstate whereever `DAOstate.reputationTotalSupply` is used.
* cleans up some places where the dao state is being redundantly created.

**Note** there still remain some places that create their own instances of DAO and DAOstate.  These are characterized as being code that is not under `DAOContainer`, for example the app header, redemptions menu and account profile code.  So ideally we would be creating and subscribing to the DAO and DAOstate in `AppContainer` rather than `DAOContainer`.  Simply in the interest of time and a perhaps lazy call that this good-enough, I didn't go that far in refactoring.  **However** note that these are now creating their own subscriptions, thus adding modestly to the number of subscriptions created by the app.

**Question** should we be setting ``fetchAllData` to true on the top-level DAOstate in DaoContainer?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 ### template
   - Features Added
   - Bugs Fixed
-    - update DAO total rep when it changes, so rep percentages remain correct
 
 ## Next release
   - Features Added
   - Bugs Fixed
+    - update DAO total rep when it changes, so rep percentages remain correct
 
 ### 2019-11-12
   - Features added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### template
   - Features Added
   - Bugs Fixed
+    - update DAO total rep when it changes, so rep percentages remain correct
 
 ## Next release
   - Features Added
--- a/src/components/Account/AccountProfilePage.tsx
+++ b/src/components/Account/AccountProfilePage.tsx
@@ -1,5 +1,5 @@
 import { promisify } from "util";
-import { IDAOState, IMemberState } from "@daostack/client";
+import { IDAOState, IMemberState, DAO } from "@daostack/client";
 
 import BN = require("bn.js");
 import * as profileActions from "actions/profilesActions";
@@ -302,9 +302,15 @@ const SubscribedAccountProfilePage = withSubscription({
     const queryValues = queryString.parse(props.location.search);
     const daoAvatarAddress = queryValues.daoAvatarAddress as string;
     const accountAddress = props.match.params.accountAddress;
+    let dao: DAO;
+    if (daoAvatarAddress) {
+      dao = arc.dao(daoAvatarAddress);
+    }
+
     return combineLatest(
-      daoAvatarAddress ? arc.dao(daoAvatarAddress).state() : of(null),
-      daoAvatarAddress ? arc.dao(daoAvatarAddress).member(accountAddress).state() : of(null),
+      // subscribe if only to to get DAO reputation supply updates
+      daoAvatarAddress ? dao.state( {subscribe: true}) : of(null),
+      daoAvatarAddress ? dao.member(accountAddress).state() : of(null),
       arc.ethBalance(accountAddress)
         .pipe(ethErrorHandler()),
       arc.GENToken().balanceOf(accountAddress)
--- a/src/components/Dao/DaoContainer.tsx
+++ b/src/components/Dao/DaoContainer.tsx
@@ -122,7 +122,7 @@ const SubscribedDaoContainer = withSubscription({
   createObservable: (props: IExternalProps) => {
     const arc = getArc(); // TODO: maybe we pass in the arc context from withSubscription instead of creating one every time?
     const daoAddress = props.match.params.daoAvatarAddress;
-    return arc.dao(daoAddress).state();
+    return arc.dao(daoAddress).state({ subscribe: true} );
   },
 });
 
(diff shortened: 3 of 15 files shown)


jellegerbrandy:
About this refactor (adding daostate to props instead of reading it from the cache): it is harmless, but it is definitly not necessary and it is also not a pattern I like.

The thing is that calling (and subscribing ot) `dao.state()` in different components is very cheap: it will only send the query the first time, and then just observe the cache for changes, which is not expensive (it is very similar to observing the redux state)

dkent600:
@jellegerbrandy 

> The thing is that calling (and subscribing ot) dao.state() in different components is very cheap: it will only send the query the first time, and then just observe the cache for changes, which is not expensive (it is very similar to observing the redux state), and keeps our code cleaner (no need to pass the state around in the props)

1)  this is not about simply using `dao.state()`, rather, is about using `dao = new DAO(address)` and `dao.state({subscribe: true})`.  Does that make a difference for you?  We are trying to minimize the number of subscriptions.

2) All of the modified components here are under DaoComponent.  It makes total sense to me to obtain a single DAO subscription for all code executing in the scope of a single DAO, whether or not that subscription is cached.  In general in software engineering it is best to just do things just once.  Fewer lines of code, less redundancy, easier maintenance, less risk.

3) many of the refactored changes involve replacing daoAvatarAddress with daoState.  So there is no change in the number of component properties.

4) Others involve using `props.match.params.daoAvatarAddress`.  I think it is inappropriate for many pieces of code to be relying this way on an application URL for configuration.

5) to reiterate: the only-refactored code that I am submitting here is in fact cleaner.  Fewer lines of code and less redundancy (thus easier maintenance and less risk).

jellegerbrandy:
> 1. this is not about simply using `dao.state()`, rather, is about using `dao = new DAO(address)` and `dao.state({subscribe: true})`.  Does that make a difference for you?  We are trying to minimize the number of subscriptions.

yes, passing the `subscribe: true` will open a new subscription, and so it should be used sparingly. 

So the preferredpattern is this:

```
<Compontent1><C1 ... dao.state(subscribe: true) /><C2 ... dao.state(subcribe: false) /></Component> 
```
which wil only open one subscription to the srever,  C2 will get updates anyway via the cache

and these less so:
```
<Compontent1><C1 ... dao.state(subscribe: true) /><C2 ... dao.state(subcribe: true) /></Component>
```
opens two separate subscriptionson the server for the same data, which is not efficient

```
<Compontent1><C1 ... dao.state(subscribe: false) /><C2 ... dao.state(subcribe: false) /></Component> 
````
will not get any updates when data on the server changes

dkent600:
@jellegerbrandy 

> `<Compontent1><C1 ... dao.state(subscribe: true) /><C2 ... dao.state(subcribe: false) /></Component>`
> which wil only open one subscription to the srever, C2 will get updates anyway via the cache

Forgive this question!  The difference between a subscription to the server and subscribing to updates to the cache?  Is it that the cache is only updated with data that has been queried elsewhere by the dApp? So in other words, the C1 subscription would cause updates to the cache which would then enable C2 to receive the same updates?

dkent600:
@jellegerbrandy 

> `<Compontent1><C1 ... dao.state(subscribe: true) /><C2 ... dao.state(subcribe: false) /></Component>`
> which wil only open one subscription to the srever, C2 will get updates anyway via the cache

I'm not sure this is the best design.

1.  it means that some code may work properly only by virtue of a side-effect of other code.
2.  (I believe) there is no way to avoid the side-effect -- if I don't want the data to be updated, I can't avoid it.

jellegerbrandy:
> 1.  it means that some code may work properly only by virtue of a side-effect of other code.

Well, yes in the sense  redux and apolloCache break the "components are true functions" pattern. But for a reason. But I also tend do say "no"  in this case: the code works "properly"  in the sense that C2 is a completely working component, that reads data from the apollocache or, if it is not there, from the server, and renders the component. 

2.  (I believe) there is no way to avoid the side-effect -- if I don't want the data to be updated, I can't avoid it except I suppose by terminating the observable.

Well (a) this is a  bit of a theoretical case (why would you not want the data to be updated?) and (b) if you _really_ wanted to, you could subscribe to `dao.state().pipe(first())`, which will just read the data once and then not send any updates anymore.

dkent600:
> but I also tend do say "no" in this case: the code works "properly" in the sense that C2 is a completely working component, that reads data from the apollocache or, if it is not there, from the server, and renders the component.

Code that relies on side-effects is not a best practice.  It may work, but for mysterious reasons that can easily and unexpectedly break by totally-unrelated code changes.

jellegerbrandy:
It's not a side effect! The component is perfectly self-contained and has predictable behavior. We just manage the cache in a separate place from the component.

dkent600:
> It's not a side effect! The subcomponent is perfectly self-contained and has predictable behavior. We just manage the cache in a separate place from the component.

Only because you are defining the case that way.  C1, or any other component make a similar query, could be _anywhere_ in the application and the coder may not be aware.

jellegerbrandy:
let's talk about this on a call

tibetsprague:
this is why I wish we still used redux instead of the Apollo cache, its much more predictable, you know what is changing where and when and how it will effect the rest of the app

PR #1253, opened 2019-11-21
add subscriptions for votes, stakes and rewards

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1756/silent

This fixes not just voting feedback, but also staking and anything related to redemptions.

I feel weird about adding subscriptions, but don't know another way to address this unless we can do a fetchAllData on the proposal itself and subscribe directly to it.  But that makes me feel weird too, so will defer to @jellegerbrandy on the best way to do this.

### DIFF
--- a/src/components/Proposal/ProposalData.tsx
+++ b/src/components/Proposal/ProposalData.tsx
@@ -123,7 +123,7 @@ export default withSubscription({
 
     if (currentAccountAddress) {
       return combineLatest(
-        proposal.state({ subscribe: true }), // state of the current proposal
+        proposal.state({ subscribe: props.subscribeToProposalDetails }), // state of the current proposal
         proposal.votes({where: { voter: currentAccountAddress }}, { subscribe: props.subscribeToProposalDetails }),
         proposal.stakes({where: { staker: currentAccountAddress }}, { subscribe: props.subscribeToProposalDetails }),
         proposal.rewards({ where: {beneficiary: currentAccountAddress}}, { subscribe: props.subscribeToProposalDetails })
--- a/src/components/Proposal/ProposalData.tsx
+++ b/src/components/Proposal/ProposalData.tsx
@@ -20,6 +20,10 @@ interface IExternalProps {
   dao: IDAOState;
   proposalId: string;
   children(props: IInjectedProposalProps): JSX.Element;
+  /**
+   * true to subscribe to changes in votes, stakes and rewards
+   */
+  subscribeToProposalDetails?: boolean;
 }
 
 interface IStateProps {
@@ -120,9 +124,9 @@ export default withSubscription({
     if (currentAccountAddress) {
       return combineLatest(
         proposal.state({ subscribe: true }), // state of the current proposal
-        proposal.votes({where: { voter: currentAccountAddress }}, { subscribe: true }),
-        proposal.stakes({where: { staker: currentAccountAddress }}, { subscribe: true }),
-        proposal.rewards({ where: {beneficiary: currentAccountAddress}}, { subscribe: true })
+        proposal.votes({where: { voter: currentAccountAddress }}, { subscribe: props.subscribeToProposalDetails }),
+        proposal.stakes({where: { staker: currentAccountAddress }}, { subscribe: props.subscribeToProposalDetails }),
+        proposal.rewards({ where: {beneficiary: currentAccountAddress}}, { subscribe: props.subscribeToProposalDetails })
           .pipe(map((rewards: Reward[]): Reward => rewards.length === 1 && rewards[0] || null))
           .pipe(mergeMap(((reward: Reward): Observable<IRewardState> => reward ? reward.state() : of(null)))),
 
--- a/src/components/Proposal/ProposalData.tsx
+++ b/src/components/Proposal/ProposalData.tsx
@@ -119,10 +119,10 @@ export default withSubscription({
 
     if (currentAccountAddress) {
       return combineLatest(
-        proposal.state(), // state of the current proposal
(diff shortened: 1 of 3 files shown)


dkent600:
@jellegerbrandy 

> I'm not sure what problem you are trying to solve here

The TP ticket describes this

dkent600:
@tibetsprague I believe you wrote the `ProposalData` component.  I wonder if you can take a look at @jellegerbrandy 's suggestion above and feedback your opinion?  Thanks!

jellegerbrandy:
ah, indeed, the tp issue is very clear.

The `ProposalDetailsPage` just uses `ProposalData` and has no separate subscription query, which is why it does not get updates. To implement the same pattern, we'd need to add a subscription to the page to all the proposal data, maybe using the same pattern as on the ProposalsPage.

dkent600:
@jellegerbrandy 

> to implement the same pattern, we'd need to add a subscription to the page to all the proposal data, maybe using the same pattern as on the ProposalsPage.

There is no "ProposalsPage".

See the code changes I made for this PR.  I have modified the `ProposalData` query to subscribe to the data.  My comment in this PR description is relative to those code changes.

When you suggested doing something like the schemeproposalspage, I infer that it would involve making the ProposalData page work like the schemeproposalspage.

tibetsprague:
so the problem here is that the SchemeProposalsPage does the subscription so that each individual ProposalCard doesn't need to, but since ProposalDetailsPage is not underneath a component that does the subscription it doesn't stay up to date. But adding subscribe to ProposalData means the subscription would also get added to every ProposalCard, so maybe we need a prop to pass in to ProposalData that tells it whether to subscribe or not, and only do that on the ProposalDetailsPage. sound good @dkent600 ?

dkent600:
@tibetsprague   @jellegerbrandy 

> so maybe we need a prop to pass in to ProposalData that tells it whether to subscribe or not, and only do that on the ProposalDetailsPage

OK I went ahead and did this.  It is odd though that the ProposalCards don't seem to require this subscription, though ProposalDetailsPage does.  I don't understand why after spending about 10 minutes trying to find a reason.

tibetsprague:
@dkent600 the reason is exactly what @jellegerbrandy and I both said, the SchemeProposalPage handles the subscription for all the proposals

dkent600:
>  It is odd though that the ProposalCards don't seem to require this subscription, though ProposalDetailsPage does

Just a note:  I didn't catch the  because I assumed that ProposalData was managing the subscriptions.  Sure looks like that unless you look closer and find the counter-intuitive:  Parent passes in a Proposal, the child passes the proposal.id to ProposalData to get back effectively the same Proposal.  This is an example of side-effects in action and the potential challenges in understanding the code that relies on them.  @jellegerbrandy

jellegerbrandy:
I  think that we have a good (or at least woring) way of creating subscxriptions, and this PR is not using it, 

Here is the pattern **without** doing server-side subscriptions. All these withSubscription compionetns will (a) first look int he cache for the data and if it does not find it and send a request to the server for the data and (b) get automatigally updates if the data i the apollo cache changes:
````
<PageComponent>
 ... <withSubscription arc.daos({subscribe: false}) />
     <withSubscription getProposals({subscribe: false}) />
     <withSubscription someotherquery({subscribe: false}) />
</PageComponent />
````
So this is a perfectly fine and working page. 

But if something on the server changes, the UI will not get updated. 

One way of solving that is to change all the {subscribe: false} in {subscribe: true}.This is the apporach of this PR.  This will work, but it will open 3 server-side subscriptions (one for each withSub component). 

The pattern we use for the SchemeProposals Page (which is our most complex page) is the follw


````
<PageComponent **subscribe here to one single superquery that subscribes to all the data we want to get updates from on this page**>
 ... <withSubscription arc.daos({subscribe: false}) />
     <withSubscription getProposals({subscribe: false}) />
     <withSubscription someotherquery({subscribe: false}) />
</PageComponent />
````
So to have some consistency in the code, I'd like for us to use this same pattern on the singel proposal page.

jellegerbrandy:
it is also confusing because of the usage of the word "subscriptions"  both
for server-side subscriptions as well as subs to the apollo cache...

And although I agree that that prposal -> propsal.id -> proposal seems a
bit clumsy, not sure what you mean by "side effects"  here?

dkent600:
>> not sure what you mean by "side effects"  here

the side-effect of the `ProposaData` query automatically receiving changes to the _global_ (apollo) cache solely by virtue of the configuration of some other not-necessarily-related-nor-obvious query (`SchemeProposalPage`).

jellegerbrandy:
But that is fine, or at least, very common. It is the same thing that redux
does: there is a global state that, and if it changes, then all components
that are based on that data get updated.

Also note that these "side effects" also occur without server-side
subscribing: for example, another component may update the apollo cache,
and as an effect from that, other compoenntsns may re-render.

dkent600:
> It is the same thing that redux does: there is a global state that, and if it changes, then all components that are based on that data get updated. - @jellegerbrandy 

The difference between our cache behavior and redux is that any component that wants to depend on a particular redux state must _explicitly_ configure itself to receive those updates.  This is not the case with our components that will receive changes to the cache regardless of whether they explicitly ask for them.  Thus a component's dependencies are neither explicit nor obvious, intentional nor necessarily desired.  This can all lead to difficult challenges in code development, testing and code maintenance.

PR #1259, opened 2019-11-25
minimize rerendering of disqus component

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=bug/1785/silent

BTW, I think we could probably reduce a lot of rendering elsewhere in the app by using the same solution as used in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   - Bugs Fixed
     - update DAO total rep when it changes, so rep percentages remain correct
     - go to error page on non-existent DAOs
+    - eliminate rerendering of Disqus component on the proposal details apge
 
 ### 2019-11-12
   - Features added
--- a/src/components/Proposal/ProposalDetailsPage.tsx
+++ b/src/components/Proposal/ProposalDetailsPage.tsx
@@ -51,6 +51,9 @@ export default class ProposalDetailsPage extends React.Component<IProps, IState>
     };
   }
 
+  // define this here rather than in render to minimize rerendering of discus component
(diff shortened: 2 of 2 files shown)


tibetsprague:
also as far as I know this always used to work, the re-rendering only started recently

dkent600:
@tibetsprague 

> I don't understand why this changes anything, can you explain?

We were passing a new instance of `disqusConfig ` into `DiscussionEmbed` on each render of ProposalDetailsPage.  This made react think that DiscussionEmbed needed to be re-rendered, when in fact it didn't.

tibetsprague:
huh, odd. that should not be an issue, Disqus even does it the way we were doing it in their primary example code: https://github.com/disqus/disqus-react

dkent600:
@tibetsprague 

> that should not be an issue,

It is a generic issue that I've read about in react articles on minimizing renders.

jellegerbrandy:
If I understand this correctly: before this PR, any time the proposaldetailspage is rendered, we create a new disqusconfig, and so also re-render the disqus subcomponent. After this PR, the disqus subcomponent is not re-rendered every time proposaldetailspage changes.

The fact that this was not a problem before, but became a problem now, may be dus to the proposalsdetailspage being rendered more often (?)

dkent600:
> may be due to the proposalsdetailspage being rendered more often (?)

That occurred to me too.  Pretty much the only explanation I think.

PR #1264, opened 2019-11-28
DAOcreator Embed

dOrgJelli:
Things still left to do:
- Optimize bundle size. See comments below.

Please review and let me know if anything else needs fixing. You can test by setting your metamask to Rinkeby (or localhost if you'd like). When you deploy you'll be given a link to view it within Alchemy.

You navigate to the page through the "Create A DAO" button on the home page.


dOrgJelli:
@tibetsprague could you help me understand why CI is failing? It seems to not recognize the default import of DAOcreator. What it is currently:
`import DAOcreator from "@dorgtech/daocreator-ui-v1";`

What it wants:
`import * as DAOcreator from "@dorgtechm/daocreator-ui-v1";`

The downside of doing the "fix" is that when you use the DAOcreator on line # 176, you'd have to write `DAOcreator.default`. This seems unnecessary and it should be able to detect the default export...

Do you know why this is happening?

dOrgJelli:
> one UX thing that i encountered: the DAO is being deployed, I'm impatient amd so I navigate away while waiting. Now do not know how to get back to teh deployment process, which is presumably now stuck halfway and wating for my input. (Or, more likely, just aborted). Can we add a confirmation "you are about to leave this page but we are still deploying your DAO - are you really sure?"

I can certainly add this, is there an example of this being done elsewhere in the application? If not, any recommendations on how you think this should be implemented?

> the build of this version is 45M (up from <12M). This is bc of teh migration dependency (which now is bundled with the compiled arc contract versions, which are enormous). It is jusdt a devdepency in the new client (and alchemy), but apparently you have it as a dependency (?, i did not check)

Ah so unfortunately @daostack/migration is included as a run-time dependency: https://github.com/dOrgTech/DAOcreator/blob/e2c3a8369f06fdbf42f510ac676a89f91cbb8efc/packages/lib/package.json#L13

This is where the ABIs are being brought in from. The migration script being used imports the ABIs like so: https://github.com/daostack/migration/blob/eff2ff5592496a881e1db7bbee148eebb2c8239d/migrate-dao.js#L45

I think we could solve this long-term by using IPFS to store the ABI, and having the ArcFactory store the hash of the ABI. In the short term we could manually delete all `@daostack/migration/contracts/${version}` folders that aren't being used (everything except v33) when Alchemy builds itself.

dOrgJelli:
An optimization that could be made within @daostack/migration is to not include duplicate ABIs in newer releases of Arc. For example, if the Avatar ABI hasn't changed since version 0.0.1-rc.19, the instead of copying it in newer releases we could just omit it or maybe create a "pointer" file like so:
`0.0.1-rc-23/Avatar.json`
```
{
  "is": "0.0.1-rc-19/Avatar"
}
```

dOrgJelli:
@jellegerbrandy I've implemented the above optimization in this PR: https://github.com/daostack/migration/pull/247

Please let me know if you think this is a good solution. From what I'm currently seeing, this would reduce the bundle size by ~27mb.

jellegerbrandy:
@dOrgJelli  I think the optimization of the contract size can be done in two directions:

- your solution that optimizes for duplicates of the code (But I think the metric we should use is **compressed** (gzipped)  bundle size, and the gain may not be as big as it seems now  :-)

-  copy only the ABIs from the contract code in migration, and ignore the bytecode (this bytecode is responsable for most of the bloat, and is only used for deploymentof contracts, not reading them). This is what I have been doing in the client in the prepare step https://github.com/daostack/client/blob/master/package.json#L28 with this script https://github.com/daostack/client/blob/master/scripts/copyABIsFromMigration.js

I have not checked what exaclty you are using from the migration pacakge in DAOCreator, but if you just need the ABIs, you can get them from the client package and not have `migration` as a full dependency. I think this is the preferred road, because it leaves the migration package free to evolve as it wants without having to bother with browser optimization.

(this is a complex issue, if you want to have a quick sync call I'm here)

jellegerbrandy:
As tthe daocreator has a different UI, it also has dependences that the rest of alchemy does not need. 
So it probably makes sense to split the bundle and have a separate chunk with the doacreator, so that a user does not need to download daocreator-specific deps when downloading the rest of the app

dOrgJelli:
Hey @jellegerbrandy, the bytecode is needed, see response here: https://github.com/daostack/migration/pull/247#issuecomment-562635462

dOrgJelli:
> As tthe daocreator has a different UI, it also has dependences that the rest of alchemy does not need.
> So it probably makes sense to split the bundle and have a separate chunk with the doacreator, so that a user does not need to download daocreator-specific deps when downloading the rest of the app

Yes I completely agree. How would I go about doing this? I'm assuming it's a webpack config option?

dOrgJelli:
Ignore the above comment, I was incorrect, see comment here: https://github.com/daostack/migration/pull/247#issuecomment-562654055

dOrgJelli:
@jellegerbrandy the version of the DAOcreator that's now being used is using an optimized version of the contracts directory which is 5.4 mb. Can you please tell me how to calculate the delta of this branch's bundle size with master? I'd like to know how much we're bloating Alchemy with this addition, and if we need to create the separate chunk.

After that's figured out, we can fix the UX issues you had mentioned above.

dOrgJelli:
@jellegerbrandy I've added another optimization and the contracts directory is only 2MB now, and the total DAOcreator build should only be around ~5MB. Please let me know if this is good enough, and what other changes you'd like made before this is merge.

A-Zak:
Hey @dOrgJelli, this is really amazing! :)

Few small comments I had:
1. can you please also add a link to "Create A DAO" to the side menu (above the privacy policy)?
2. once you deploy your DAO, it could be great if you could have a button that would take you to it... especially since there's no way to find it yet... I would even go as far as adding a warning sign and a message to the user to save this url otherwise it'll be really hard to find again 😅
3. The "Create A DAO" button we have now in the All DAOs screen, can you please make it the same style (and position) as the "new proposal" button (filled with blue, rounded sides, some padding from the title separator line, etc)?

src/components/DaoCreator/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+type IExternalProps = RouteComponentProps<any>;
+
+interface IDispatchProps {
+  showNotification: typeof showNotification;
+}
+
+type IProps = IDispatchProps;
+
+const mapStateToProps = (state: IRootState, ownProps: IExternalProps): IExternalProps => {
```
tibetsprague: we don't need mapStateToProps if its not actually importing any new props. you can just pass null to the connect call

tibetsprague:
> @tibetsprague could you help me understand why CI is failing? It seems to not recognize the default import of DAOcreator. What it is currently:
> `import DAOcreator from "@dorgtech/daocreator-ui-v1";`
> 
> What it wants:
> `import * as DAOcreator from "@dorgtechm/daocreator-ui-v1";`
> 
> The downside of doing the "fix" is that when you use the DAOcreator on line # 176, you'd have to write `DAOcreator.default`. This seems unnecessary and it should be able to detect the default export...
> 
> Do you know why this is happening?

sorry i took so long to look at this @dOrgJelli , i assume you worked this out?

tibetsprague:
you can use `npm run analyze` to see the bundle size and compare

tibetsprague:
@dOrgJelli could you fix the conflicts with the latest dev and do one more compare of the bundle size using analyze?

src/components/DaoCreator/index.tsx
```diff
@@ -0,0 +1,60 @@
+import * as React from "react";
+import { connect } from "react-redux";
+import { RouteComponentProps } from "react-router-dom";
+import DAOcreator from "@dorgtech/daocreator-ui-v1";
```
tibetsprague: so here is where we could use dynamic importing (lazy loading) to load this only at runtime, and reduce the size of the initial bundle: https://reactjs.org/docs/code-splitting.html

dOrgJelli:
So sorry for the delay on this @A-Zak @tibetsprague, hopping on this now.

src/App.tsx
```diff
@@ -97,6 +97,7 @@ export class App extends React.Component<{}, {
                 <Route path="/daos" component={AppContainer}/>
                 <Route path="/profile" component={AppContainer}/>
                 <Route path="/redemptions" component={AppContainer}/>
+                <Route path="/dao-creator" component={AppContainer} />
```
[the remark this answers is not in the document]
dOrgJelli: I've changed it to `/daos/create` so that it doesn't conflict with the `/dao/${address}` route. `/daos/create` follows the same pattern as `proposals/create`.

dOrgJelli:
@tibetsprague here's the size of the DAOcreator UI + Lib relative to the rest of Alchemy. I'm going to add lazy loading like you suggested.
Screenshot from 2020-02-05 23-37-34

Screenshot from 2020-02-05 23-36-55

Screenshot from 2020-02-05 23-36-42

Screenshot from 2020-02-05 23-36-16

tibetsprague:
@dOrgJelli thats not too too bad...

dOrgJelli:
@jellegerbrandy @tibetsprague @A-Zak all requested changes should be made. You can test DAO creation by setting your metamask to Rinkeby and going through the deployment process here:
https://alchemy-daocreator-embe-skttt7.herokuapp.com/daos/create

@A-Zak here's are some UI changes I made based on your feedback:
image

A-Zak:
@dOrgJelli looks great, thank you! :) 🙏

PR #1279, opened 2019-12-13
[WIP] Dao debt info

shivgupt:
Show DAOs outstanding contribution reward debt in the sidebar.

### DIFF
--- a/src/layouts/SidebarMenu.tsx
+++ b/src/layouts/SidebarMenu.tsx
@@ -8,12 +8,13 @@ import TrainingTooltip from "components/Shared/TrainingTooltip";
 import BN = require("bn.js");
 import * as classNames from "classnames";
 import withSubscription, { ISubscriptionProps } from "components/Shared/withSubscription";
+import gql from "graphql-tag";
 import * as GeoPattern from "geopattern";
-import { ethErrorHandler, formatTokens, getExchangesList, supportedTokens } from "lib/util";
+import { ethErrorHandler, formatTokens, getExchangesList, supportedTokens, getDaoDebt } from "lib/util";
 import * as queryString from "query-string";
 import * as React from "react";
 import { matchPath, Link, RouteComponentProps } from "react-router-dom";
-import { first } from "rxjs/operators";
+import {  map, first } from "rxjs/operators";
 import { IRootState } from "reducers";
 import { connect } from "react-redux";
 import { combineLatest, of, from } from "rxjs";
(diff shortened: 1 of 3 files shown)


src/layouts/SidebarMenu.tsx
```diff
@@ -264,6 +273,74 @@ const SubscribedEthBalance = withSubscription({
...
+  );
+};
+
+const SubscribedDaoDebt = withSubscription({
+  wrappedComponent: DaoDebt,
+  checkForUpdate: (oldProps: IDebtProps, newProps: IDebtProps) => {
+    return oldProps.dao.address !== newProps.dao.address;
+  },
+  createObservable: async (props: IDebtProps) => {
+    //const daoState = props.dao;
```
tibetsprague: remove comment

src/lib/util.ts
```diff
@@ -391,6 +393,57 @@ export function hasGpRewards(reward: IRewardState) {
...
+  let externalTokenReward;
+  let rewards;
+
+  const supportedExternalTokens = supportedTokens();
+
+  result["ETH"] = new BN(0);
+  for (const proposal of proposals) {
+
+    // Calculate CR Debt
+    if (proposal.contributionReward && proposal.contributionReward.externalToken !== utils.NULL_ADDRESS ) {
```
dOrgJelli: Is the second part of this if statement needed? What if a CR proposal doesn't have an externalToken as a requested reward? For example, if it's only asking for ETH, this if statement would fail no?

shivgupt: Nice eye! The second part of if statement was supposed to be separate check in token debt section and I messed up. Thank you

shivgupt: Nice Eye!

I messed up the loging while trying to exclude the proposals which cannot be redeemed because of null externalToken but non zero ethReward and externalTokenReward.

shivgupt:
Oops I just decreased the size in side bar to fit another block for dao Debt. @A-Zak can you see if this is alright or needs an update

src/layouts/SidebarMenu.tsx
```diff
@@ -264,6 +273,73 @@ const SubscribedEthBalance = withSubscription({
...
+  },
+  createObservable: async (props: IDebtProps) => {
+
+    const arc = getArc();
+    const query = gql`query
+    {
+      proposals(
+        where: {
+          dao: "${props.dao.address}",
+          accountsWithUnclaimedRewards_not: []
```
jellegerbrandy: this is, potentially, an ENORMOUS query, as this includes also all proposals in which some user has not claimed his rep yet. If we cannot do this with a  reasonably sized query, we will need to update the subgraph to keep the number for us.

src/layouts/SidebarMenu.tsx
```diff
@@ -264,6 +273,73 @@ const SubscribedEthBalance = withSubscription({
...
+          executionState_in: ["BoostedTimeOut", "BoostedBarCrossed", "QueueBarCrossed"]
```
jellegerbrandy: whi not "Executed"?

jellegerbrandy:
closing this as it does not see any progress, but cannot be merged as-is

PR #1280, opened 2019-12-14
[WIP] Competitions

dkent600:
Resolves: https://daostack.tpondemand.com/RestUI/Board.aspx#page=board/5209716961861964288&appConfig=eyJhY2lkIjoiQjgzMTMzNDczNzlCMUI5QUE0RUE1NUVEOUQyQzdFNkIifQ==&boardPopup=feature/1858/silent


src/components/Scheme/ContributionRewardExtRewarders/DetailsPageRouter.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+}
+
+export default withSubscription({
+  wrappedComponent: DetailsPageRouter,
+  loadingComponent: <div className={css.loading}><Loading/></div>,
+  errorComponent: null,
+  checkForUpdate: [],
+  createObservable: (props: IProps) => {
+    const arc = getArc(); // TODO: maybe we pass in the arc context from withSubscription instead of creating one every time?
```
jellegerbrandy: `getArc()` does not create a new arc context everytime, only the first time it is called, then it stores it in a global ar.

src/crxRegistry/README.md
```diff
@@ -0,0 +1,9 @@
+# ContributionRewardExt Rewarder Registry
```
jellegerbrandy: I am not in principle against having such a registrty, but it seems that at this point (we are implementing this pattern just a single time, for the first time) it is premature - too early to generalize (or to know how to gneeralize this).

src/components/Scheme/ContributionRewardExtRewarders/DetailsPageRouter.tsx
dkent600: ha, that comment in the code is very old, made it here via blind copy/paste.

src/crxRegistry/README.md
dkent600: yeah, and too early to be commenting on it in the PR  :-)

src/components/Dao/DaoContainer.tsx
```diff
@@ -65,8 +66,14 @@ class DaoContainer extends React.Component<IProps, null> {
   private daoDiscussionRoute = (routeProps: any) => <DaoDiscussionPage {...routeProps} dao={this.props.data[0]} />;
   private daoProposalRoute = (routeProps: any) =>
     <ProposalDetailsPage {...routeProps}
-      daoState={this.props.data[0]}
       currentAccountAddress={this.props.currentAccountAddress}
+      daoState={this.props.data[0]}
+      proposalId={routeProps.match.params.proposalId}
+    />;
+  private daoCrxProposalRoute = (routeProps: any) =>
```
jellegerbrandy: what does "crx" stand for? (perhaps use something more readable, like `competition`?)

src/components/Dao/DaoContainer.tsx
```diff
@@ -104,10 +111,14 @@ class DaoContainer extends React.Component<IProps, null> {
               render={this.daoProposalRoute}
             />
 
+            <Route exact path="/dao/:daoAvatarAddress/crx/proposal/:proposalId"
```
jellegerbrandy: As thi sis probably best thought of as an alternative view of the proposal, perhaps the url can be:

`"/dao/:daoAvatarAddress/proposal/:proposalId/competition"`

src/components/Scheme/ContributionRewardExtRewarders/rewardersProps.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export enum CrxRewarderComponentType {
+  List,
+  Details,
+  CreateProposal
+}
+export const getCrxRewarderComponent = (schemeState: ISchemeState, type: CrxRewarderComponentType): Promise<any> => {
+  const contractName = rewarderContractName(schemeState);
+  if (contractName) {
+    return import(`components/Scheme/ContributionRewardExtRewarders/${contractName}/${CrxRewarderComponentType[type]}`)
```
jellegerbrandy: Doing dynamic imports is a Big Change, with consequences forr the bundling process. I would only do this if I had a Really Good reason... do we have such a reason?

src/components/Scheme/ContributionRewardExtRewarders/Competition/Details.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+    return combineLatest(
+      submissions,
+      submissions.pipe(
+        concatMap((submissions: Array<ICompetitionSuggestion>) => {
+          if (!submissions.length) {
+            return of([]);
+          } else {
+            // seems like there should be a more elegant rxjs way
+            const observables: Array<Observable<boolean>> = [];
```
jellegerbrandy: The problem here is not rxjs, I think. The pattern here would be to have a "Submission" subcomponent that has sets up its own observable (instead of joining the submissios here and then `map`ping them to subcombonents in the render method above)

src/components/Scheme/ContributionRewardExtRewarders/Competition/CreateProposal.tsx
```diff
@@ -0,0 +1,626 @@
...
+      </div>
+    );
+  }
+}
+
+const SubscribedCreateContributionRewardExProposal = withSubscription({
+  wrappedComponent: CreateProposal,
+  checkForUpdate: ["daoAvatarAddress"],
+  createObservable: (props: IExternalProps) => {
+    const arc = getArc(); // TODO: maybe we pass in the arc context from withSubscription instead of creating one every time?
```
jellegerbrandy: `getArc` is not creating a new context but is returin "the" context, so we're already good

src/components/Scheme/ContributionRewardExtRewarders/Competition/Details.tsx
```diff
@@ -0,0 +1,331 @@
...
+const CompetitionDetailsConnected = connect(mapStateToProps, mapDispatchToProps)(CompetitionDetails);
+
+export default withSubscription({
+  wrappedComponent: CompetitionDetailsConnected,
+  loadingComponent: null,
+  errorComponent: (props) => <div>{ props.error.message }</div>,
+  checkForUpdate: [],
+  createObservable: (props: IExternalProps & IStateProps ) => {
+    /**
+     * HEADS UP:  we subscribe here because we can't rely on getting to this component via CompetitionCard
```
jellegerbrandy: ???
(I mean, what you are saying here seems ok, I'm jus tnot sure why we need a "HEADS UP" warning here...)

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+      // concat()  (??)
+      // or:
+      // map((suggestions: Array<CompetitionSuggestion>) => suggestions.map((suggestion) => suggestion.staticState ))
+
+      // work-around hack because CompetitionSuggestion actually contains all we need
+      map((suggestions: Array<CompetitionSuggestion>) => suggestions.map((suggestion) => suggestion as unknown as ICompetitionSuggestion) )
+    );
+};
+
+export const getProposalSubmissions = (proposalId: string, subscribe = false): Observable<Array<ICompetitionSuggestion>> => {
```
jellegerbrandy: If possible, can we avoid one-liner definitions? It just makes the code harder to read (and god knows we have enough complexity as it is :-))

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+  return getSubmissions(proposalId, undefined, subscribe);
+};
+
+export const getProposalSubmission = (proposalId: string, id: string, subscribe = false): Observable<ICompetitionSuggestion> => {
+  /** 
+   * won't subscribe here because the only use case is SubmissionDetails where we can assume the suggestion has
```
jellegerbrandy: won't subscribe, or does it dpeend on the parameter?

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+  return competition.suggestions({ where: options }, { subscribe, fetchAllData: true })
+    .pipe(
+      // FAKE -- until .fetchStaticState() exists on CompetitionSuggestion
+      // mergeMap((suggestions: Array<CompetitionSuggestion>) => suggestions.map((suggestion) => from(suggestion.fetchStaticState()) )),
+      // concat()  (??)
+      // or:
+      // map((suggestions: Array<CompetitionSuggestion>) => suggestions.map((suggestion) => suggestion.staticState ))
+
+      // work-around hack because CompetitionSuggestion actually contains all we need
+      map((suggestions: Array<CompetitionSuggestion>) => suggestions.map((suggestion) => suggestion as unknown as ICompetitionSuggestion) )
```
jellegerbrandy: This seems a workaround around something that should have been handled better in the client...

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+  id?: string; // id of the competition
+  suggestionId?: string; // the "suggestionId" is a counter that is unique to the scheme
+  // - and is not to be confused with suggestion.id
+}
+
+const getSubmissions = (
+  proposalId: string,
+  options?: IGetSubmissionsOptions,
+  subscribe = false
+): Observable<Array<ICompetitionSuggestion>> => {
```
jellegerbrandy: It seems that with the definitions in this `utils` file you are mostly:
- fixing things that should have been handled in the client (which given the history of developing this stuff is understandable and good, but should not be considered the deifnitive solution)
-  wrapping the client functions (which is not a good idea, as it obfuscates the code)

Ideally, we'd not have the `getSubmissions`, `getProposalSubmissions` etc because there would be functios aviable from the client.

src/components/Scheme/ContributionRewardExtRewarders/README.md
```diff
@@ -0,0 +1,5 @@
+# ContributionRewardExt Rewarder Registry
```
jellegerbrandy: So I believe we talked about this, but for the record, I do not think this is necessary, and I also think it is not worth the effort. 

I believe we all  (me, adam, oren, you?) think the "rewarder" pattern used in ContributionRewardExt.sol` was a bad idea, so it is not likely that we are going to re-use it in the future. So it is a wasted effort to try to anticipate other uses (and it does make the code more complex...).

Not sure how finished this modularity part is, but if it is more work to finish this then to remove it, I vote strongly for removing it

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -132,15 +185,40 @@ const SubscribedSchemeContainer = withSubscription({
   checkForUpdate: ["schemeId"],
   createObservable: async (props: IProps) => {
     const arc = getArc();
-    const scheme = arc.scheme(props.schemeId);
+    const scheme = arc.scheme(props.schemeId) as any;
 
     // TODO: this may NOT be the best place to do this - we'd like to do this higher up
-
+    // why are we doing this for all schemes and not just the scheme we care about here?
```
jellegerbrandy: good point. Seems we should at a `scheme: scheme.id` to the where clause here.

src/components/Dao/DaoContainer.tsx
dkent600: ContributionRewardExt

src/components/Scheme/ContributionRewardExtRewarders/rewardersProps.ts
dkent600: The really good reason is so we don't have to statically reference and load all rewarding contract components.  Instead we are lazy loading, which is efficient for memory usage and bandwidth.

I don't see this as a big change, wondering why you say this?  When doing dynamic imports, which we can (and I think likely should) do elsewhere when we import JSON for generic schemes, and which I also do elsewhere in the competition code for JSON, then we want them to be bundled smartly -- they aren't by default.  The plugin I am using manages this nicely.

src/components/Scheme/ContributionRewardExtRewarders/Competition/CreateProposal.tsx
dkent600: please ignore that stupid comment that wasn't mine.  I'll go ahead and find all of them in the app and delete them

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
dkent600: Yes, it is part of my comments to you in discord, and the ticket you created in client repo.

src/components/Scheme/ContributionRewardExtRewarders/README.md
dkent600: The work is 100% finished.  It works.

The current architecture may have been portrayed as a "bad idea", but at this time it is what we have and I don't see it being rewritten any time soon.  The architecture works and no decision has been made to change it.  So I think we maintain optionality by keeping this code, staying ready in case someone does want to add a new rewarder contract, which will be really easy to do.

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
dkent600: Sorry, I don't understand your question

src/components/Scheme/SchemeContainer.tsx
dkent600: the only reason I can think of is that as it is done here it results in one subscription.  if filtered by scheme.id, then would be one subscription per scheme.   Which do you think is best, @jellegerbrandy ?

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
dkent600: Agreed, done.

dkent600: I think this comment may be obsolete

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
dkent600: I think this comment may be obsolete

src/components/Scheme/ContributionRewardExtRewarders/README.md
jellegerbrandy: :+1:

src/components/Scheme/ContributionRewardExtRewarders/rewardersProps.ts
jellegerbrandy: * The problem is that if you import modules in this way (using live code) it is much harder (actually impossible) to calculate the dependencies of the module that does the import, and so it is harder for webpack to split the bundle into coherent chunks (I guess, in fact I do not even really understand how webpack knows what to include)

*  We are curently bundling all our code together, and this import syntax does not change that.So I do not see how this this "lazy loading" or "efficient for memory usage and bandwidth?", and also I do not see how this would help us move towards that  (on the contrary, see above)

any case, this is not blocking, but we may want to change this when we do bunde optimizing...

src/components/Scheme/ContributionRewardExtRewarders/Competition/List.tsx
```diff
@@ -0,0 +1,98 @@
+import * as React from "react";
```
tibetsprague: again would prefer to name this file CompetitionList. also good for filename to match component name

src/components/Scheme/ContributionRewardExtRewarders/Competition/List.tsx
```diff
@@ -0,0 +1,98 @@
...
+      return 0;
+    }
+
+    const statusIndexA = Object.values(CompetitionStatusEnum).indexOf(statusA.status);
+    const statusIndexB = Object.values(CompetitionStatusEnum).indexOf(statusB.status);
+    /**
+     * sort by the ordinal position of the CompetitionStatusEnum values
+     */
+    const retval = SortService.evaluateNumber(statusIndexA, statusIndexB);
+    if (retval)
```
tibetsprague: dont we need to use braces here? has this been linted?

src/components/Scheme/ContributionRewardExtRewarders/Competition/StatusBlob.tsx
```diff
@@ -0,0 +1,22 @@
+import { ICompetitionProposalState, ICompetitionSuggestionState } from "@daostack/client";
```
tibetsprague: what is a StatusBlob?

src/components/Scheme/ContributionRewardExtRewarders/Competition/SubmissionDetails.tsx
```diff
@@ -0,0 +1,165 @@
...
+  wrappedComponent: SubmissionDetails,
+  loadingComponent: null,
+  errorComponent: (props) => <div>{ props.error.message }</div>,
+  checkForUpdate: ["currentAccountAddress"],
+  createObservable: async (props: IExternalProps) => {
+    return combineLatest(
+      getProposalSubmission(props.proposalState.id, props.suggestionId, true),
+      getSubmissionVoterHasVoted(props.suggestionId, props.currentAccountAddress, true),
+      props.currentAccountAddress ? getCompetitionVotes(props.proposalState.id, props.currentAccountAddress, true) : of([])
+      // , props.proposalState.proposal.rewards({ where: { beneficiary: props.currentAccountAddress} }, {subscribe: true })
```
tibetsprague: remove comments

src/components/Dao/DaoContainer.tsx
```diff
@@ -16,6 +16,7 @@ import { ModalRoute } from "react-router-modal";
 import { IRootState } from "reducers";
 import { showNotification } from "reducers/notifications";
 import { IProfileState } from "reducers/profilesReducer";
+import DetailsPageRouter from "components/Scheme/ContributionRewardExtRewarders/DetailsPageRouter";
```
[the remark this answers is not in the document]
tibetsprague: i think i would call it ContributionRewardExtDetailsPage

src/components/Scheme/ContributionRewardExtRewarders/rewardersProps.ts
tibetsprague: dynamic importing is a good idea where possible i think. it makes for smaller bundles up front, though some small amount of waiting later on for the user...

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -30,13 +32,18 @@ interface IExternalProps extends RouteComponentProps<any> {
   daoState: IDAOState;
 }
 
-interface IStateProps {
+interface IExternalStateProps {
```
tibetsprague: what does ExternalStateProps mean?

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -30,13 +32,18 @@ interface IExternalProps extends RouteComponentProps<any> {
...
   schemeId: Address;
 }
 
-type IProps = IExternalProps & IDispatchProps & IStateProps & ISubscriptionProps<ISchemeState>;
+interface IStateProps {
```
tibetsprague: this should just be IState

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -97,34 +139,45 @@ class SchemeContainer extends React.Component<IProps, null> {
...
+          <div className={css.createProposal}>
             <TrainingTooltip placement="topRight" overlay={"A small amount of ETH is necessary to submit a proposal in order to pay gas costs"}>
               <a className={
                 classNames({
-                  [css.createProposal]: true,
                   [css.disabled]: !isActive,
                 })}
               data-test-id="createProposal"
-              href="javascript:void(0)"
+              href="#!"
```
tibetsprague: this is new to me... cool!

src/components/Shared/Countdown.scss
```diff
@@ -1,6 +1,6 @@
 .container {
-  width: 100%;
```
tibetsprague: you sure this CSS change works everywhere the countdown shows? i have a memory of changing to width: 100% sometime recently...

src/lib/util.ts
```diff
@@ -5,15 +5,18 @@ import {
...
   IProposalState,
   IRewardState,
-  ISchemeState } from "@daostack/client";
+  ISchemeState} from "@daostack/client";
 import { GenericSchemeRegistry } from "genericSchemeRegistry";
 import { of } from "rxjs";
 import { catchError } from "rxjs/operators";
 
 import BN = require("bn.js");
+import "moment";
```
tibetsprague: what does this do?

dkent600: See the timezone date functions I added

src/components/Shared/Countdown.scss
dkent600: Note that I renamed the exiting countdown "proposalCountdown" to make room for this one that contains no proposal specific code.

src/components/Scheme/SchemeContainer.tsx
dkent600: yeah, this gets rid of the runtime console warnings

src/components/Scheme/SchemeContainer.tsx
dkent600: Done

src/components/Scheme/SchemeContainer.tsx
dkent600: State that is coming from outside the component (in this case, `match.params.schemeId`)  (see `mapStateToProps`).  To be consistent with your request below, I'm renaming to `IExternalState`.

src/components/Scheme/ContributionRewardExtRewarders/Competition/Card.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const submissions = this.props.data[0];
+    return competitionStatus(competition, submissions);
+  }
+
+  private fireHandleChange(status: CompetitionStatus) {
+    this.props.handleStatusChange(this.props.proposalState, status);
+  }
+
+  private onEndCountdown = () => {
+    // paranoia about timer inprecision
```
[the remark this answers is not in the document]
dkent600: It is why I'm using setTimer before recomputing the competition state which depends on the current browser/javascript time.  I have worried that the timers may be inprecise in arriving at zero at precisely when they are supposed to.  I think this may be paranoia.

src/components/Proposal/ProposalDetailsPage.tsx
```diff
@@ -1,19 +1,20 @@
 import { Address, IDAOState, IProposalStage, Vote } from "@daostack/client";
-import classNames from "classnames";
+import * as classNames from "classnames";
```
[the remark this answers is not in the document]
dkent600: no idea, could this be from jelle's merge?  anyway I just set it back.

src/components/Proposal/Create/CreateProposalPage.tsx
```diff
@@ -51,25 +63,31 @@ class CreateProposalPage extends React.Component<IProps, null> {
     history.push("/dao/" + daoAvatarAddress + "/scheme/" + schemeId);
   }
 
+  public async componentDidMount() {
+    const newState = {};
+
+    if (!this.state.createCrxProposalComponent) {
+      Object.assign(newState, { createCrxProposalComponent: await getCrxRewarderComponent(this.props.data, CrxRewarderComponentType.CreateProposal) });
```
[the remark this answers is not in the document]
dkent600:
```
    /**
     * Get the "CreateProposal" modal dialog component supplied by the rewarder contract associated
     * with this CrExt scheme (if it is a CrExt scheme -- very cheap if not a CrExt).
     */
```

src/components/Proposal/Create/CreateProposalPage.tsx
```diff
@@ -15,28 +14,41 @@ import { BreadcrumbsItem } from "react-breadcrumbs-dynamic";
 import { connect } from "react-redux";
 import { IRootState } from "reducers";
 import { RouteComponentProps } from "react-router-dom";
+import { CrxRewarderComponentType, getCrxRewarderComponent, rewarderContractName } from "components/Scheme/ContributionRewardExtRewarders/rewardersProps";
```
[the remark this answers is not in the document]
dkent600: CrExt or ContributionRewardExt

src/components/Dao/DaoContainer.tsx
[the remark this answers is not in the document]
dkent600: This is configuration by convention.  Its association with CrExt is clear by name of the folder that contains it.  Every reward contract must use those names as described in `src\components\Scheme\ContributionRewardExtRewarders\readme.md`.  I need to confirm  that the readme is up-to-date....

dkent600:
@jellegerbrandy regarding your concern about bundling. chunks, etc, see the `@babel/plugin-syntax-dynamic-import`.  It supposedly makes this quite efficient.

src/components/Scheme/ContributionRewardExtRewarders/Competition/Details.tsx
```diff
@@ -0,0 +1,400 @@
...
+    const competition = this.props.proposalState.competition;
+    const submissions = this.props.data[0];
+    return competitionStatus(competition, submissions);
+  }
+
+  public componentDidMount() {
+    const newState = {};
+    /**
+     * use `window` because a route with these params isn't configured
+     * externally to the Competition code in Alchemy, and thus the params
```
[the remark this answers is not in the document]
dkent600: I tried hard to use `route` but it just wouldn't work as well.  I don't think this bit of code is a big deal considering it work great and gives us the modularity.  For what it is worth, it is not the only place we use `window.location`.

src/actions/arcActions.ts
```diff
@@ -154,11 +153,12 @@ export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: strin
 
     if (privateKey) {
       const reputationFromTokenScheme = scheme.ReputationFromToken as ReputationFromTokenScheme;
+      const agreementHash = await reputationFromTokenScheme.getAgreementHash();
       const state = await reputationFromTokenScheme.scheme.fetchStaticState();
       const contract =  arc.getContract(state.address);
       const block = await arc.web3.eth.getBlock("latest");
       const gas = block.gasLimit - 100000;
-      const redeemMethod = contract.methods.redeem(addressToRedeem);
+      const redeemMethod = contract.methods.redeem(addressToRedeem, agreementHash);
```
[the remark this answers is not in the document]
dkent600: When you say that are you considering that this needs the ABI to sign a tx, etc?I would prefer to let you "fix" this  :-)

CHANGELOG.md
```diff
@@ -4,7 +4,25 @@
 
 ## Next release
   - Features Added
+    - Many tweaks to the language throughout the app
```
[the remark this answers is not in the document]
dkent600: cleaned up

src/components/Scheme/ContributionRewardExtRewarders/Competition/List.tsx
dkent600: This is one of the files that must confirm to a naming convention.  (I explained in an earlier response)

dkent600: It might be possible to define the convention such that the containing folder name must be prepended to the file name.

src/components/Scheme/ContributionRewardExtRewarders/Competition/List.tsx
dkent600: it is linted.  apparently breaks no rules.  I added the braces anyway.

src/components/Scheme/ContributionRewardExtRewarders/Competition/StatusBlob.tsx
dkent600: added:

/**
 * displays the current status of a competition
 */

src/components/Scheme/ContributionRewardExtRewarders/Competition/SubmissionDetails.tsx
dkent600: done

src/components/Proposal/Create/CreateProposalPage.tsx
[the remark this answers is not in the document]
jellegerbrandy: `ContributionRewardExt` has my vote

src/components/Proposal/ProposalDetailsPage.tsx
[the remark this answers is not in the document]
jellegerbrandy: must be the merge, indeed (sorry)

src/components/Scheme/ContributionRewardExtRewarders/Competition/Card.tsx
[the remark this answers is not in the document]
jellegerbrandy: it _is_ indeed paranoid, probably; the ethereum chanin has only a  vague concept of time (one tick every 15 seconds, about, give or take a second), so our times are much more precise (and therefore always wrong, strictly speaking)

src/lib/util.ts
jellegerbrandy: you did not really answer the question, and I also wonder wha ta "bare"  import does here (it it is because importing "moment" hsa side effects that we want, then we need a comment)

dkent600: /**
 * gotta load moment in order to use moment-timezone directly
 */

dkent600: the competition UI displays datetimesthat include the name of the users' local timezone and their GMT offset

src/components/Scheme/ContributionRewardExtRewarders/Competition/Card.tsx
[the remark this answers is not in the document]
dkent600: Before I added the setTimeout I was seeing the UI not always updating on this event.  That stopped after I added this.  Maybe a coincidence.

src/components/Scheme/ContributionRewardExtRewarders/Competition/Card.tsx
```diff
@@ -0,0 +1,139 @@
...
+    } = this.props;
+
+    const competition = proposalState.competition;
+    const submissions = this.props.data[0];
+    const votes = this.props.data[1];
+    const numWinningSubmissions = submissions.filter((submission) => submission.isWinner).length;
+    const overWithWinners = status.overWithWinners;
+
+    return <div className={css.competitionCardContainer} data-test-id={"competition-card-" + proposalState.id}>
+      <StatusBlob competition={competition} submissions={submissions}></StatusBlob>
```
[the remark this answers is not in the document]
dkent600: I added a descriptive comment in the component file and in another comment in this review

src/actions/arcActions.ts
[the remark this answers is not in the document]
jellegerbrandy: Ok, sorry, this was a misunderstanding, it is not even your code I'm talking about but mine :-)
It is a very special case, so I thik it is sort of ok to do this in alchemy

src/components/Dao/DaoContainer.tsx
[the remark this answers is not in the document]
tibetsprague: well, I still prefer a name that is more descriptive for the file, but im not going to push it

src/components/Scheme/SchemeContainer.tsx
tibetsprague: these are not state items though they are props coming from the state (mapStateToProps) that is why it is called IStateProps

tibetsprague: these are not state items though they are props coming from the state (mapStateToProps) that is why it is called IStateProps

src/components/Shared/Countdown.scss
tibetsprague: I saw that, just double checking the layout is working right

src/components/Scheme/ContributionRewardExtRewarders/Competition/List.tsx
tibetsprague: that would be ideal to me. but not a huge deal

src/components/Shared/Countdown.scss
dkent600: working fine  :-)

src/components/Scheme/SchemeContainer.tsx
dkent600: I just like to distinguish between the internal state and the external state.  `IExternalState` does end up being part of `IProps`, so seems like `IExternalStateProps` makes the most sense.  Does that make sense to you?

tibetsprague: what makes this one External though? as opposed to everywhere where we use IStateProps

dkent600: So I guess 'IExternalStateProps' is inconsistent, but `IStateProps` makes me think it is part of the component's state, which it isn't.  Maybe I have to get used to `IState` just meaning the component's state, and `IStateProps` meaning props that come from the external (not internal) state.

dkent600: So I'm renaming to `IStateProps`

tibetsprague: yeah, props are things that come from somewhere else (external), state is the internal state. IStateProps comes from the Redux global state, so maybe we should rename IStateProps to IReduxStateProps. actually I guess IExternalStateProps is ok by me... if you want to rename all of them go for it.

dkent600: LOL, maybe another time  :-)

Issue #1291, opened 2019-12-30
USDC wrong decimals in Redeem 


orishim:
USDC shows correct decimals in proposal view, but not in the redeem view:
<img width="632" alt="Screen Shot 2019-12-30 at 1 16 46 PM" src="https://user-images.githubusercontent.com/10736693/71570338-37d9af80-2b07-11ea-8c26-cc67ee053e76.png">

Related: #795

PR #1312, opened 2020-01-16
Fix display of external token rewards (in particular USDC)

tibetsprague:
Also fix ability to execute proposals when DAO doesnt have funds to redeem for beneficiary.


src/actions/arcActions.ts
```diff
@@ -64,6 +64,11 @@ export function executeProposal(avatarAddress: string, proposalId: string, _acco
 
     // Call claimRewards to both execute the proposal and redeem the ContributionReward rewards,
     //   pass in null to not redeem any GenesisProtocol rewards
+    const originalErrorHandler = observer[1];
```
jellegerbrandy: this works?!

jellegerbrandy: apparently so :-)

jellegerbrandy:
this fixes #1291

PR #1356, opened 2020-01-31
Fix errors about nested <a> tags

tibetsprague:
This means you cant click on the whole proposal card to go to the proposal details but you can click on all of the following:
- proposal title
- proposal description
- tags
- the timer
- the voting and staking section at the bottom of the card


dkent600:
Maybe could make a ref of the div of the entire card, give it an onClick handler, passing through any click that is on any of its children, ie that is not directly on the div.

Issue #1348, opened 2020-01-31
Reputation from tokens does not update correctly when siwthing accounts


jellegerbrandy:
Basically, it seems it always resets to 0, and you need to reload the page to see the actual amount you can redeem..

dkent600:
I'm not sure it updates after redeeming, either.

tibetsprague:
i actually dont know how to test this. if i go to the Nectar DAO on ganache and click on Reputation from Token i see this error: 
Uncaught (in promise) Error: Returned values aren't valid, did it run Out of Gas? You might also see this error if you are not using the correct ABI for the contract you are retrieving data from, requesting data from a block number that does not exist, or querying a node which is not fully synced.

tibetsprague:
with the way we do client calls now i dont know of any good way to detect that redeem succeeded and then set the tokens to redeem to 0 in the component...

tibetsprague:
ohhhh, looking at the code it only fires after 23 confirmations. this is confusing I think, can we make it fire after 3?

Issue #1319, opened 2020-01-21
reputation from token - should disable redeem in the UI if account does not have rep to redeem


orenyodfat:
if account has zero rep to redeem the redeem button should be disable.
- currently it might mislead users - because they can redeem 0 rep and loose the gas for the. tx.
- if user will try to do that and in the future he will have rep (from token) he could not redeemed it because he already redeemed (zero).

PR #1394, opened 2020-02-07
Update # rep to redeem from tokens when switching account

tibetsprague:
Also attempt to set to 0 after redeem succeeds, but hard to do this with current client library, see https://github.com/daostack/client/issues/412

Also disable redeem buttons when 0 rep to redeem

Fixes https://github.com/daostack/alchemy/issues/1348 & https://github.com/daostack/alchemy/issues/1319

### DIFF
--- a/src/actions/arcActions.ts
+++ b/src/actions/arcActions.ts
@@ -144,7 +144,7 @@ export function redeemProposal(daoAvatarAddress: string, proposalId: string, acc
   };
 }
 
-export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: string, privateKey: string|undefined, redeemerAddress: Address|undefined) {
+export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: string, privateKey: string|undefined, redeemerAddress: Address|undefined, redemptionSucceededCallback: () => void) {
   return async (dispatch: Redux.Dispatch<any, any>) => {
     const arc = getArc();
 
@@ -181,6 +181,7 @@ export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: strin
       try {
         await arc.web3.eth.sendSignedTransaction(signedTransaction.rawTransaction);
         dispatch(showNotification(NotificationStatus.Success, "Transaction was succesful!"));
+        redemptionSucceededCallback();
       } catch(err) {
         dispatch(showNotification(NotificationStatus.Failure, `Transaction failed: ${err.message}`));
       }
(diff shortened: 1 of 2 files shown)


src/actions/arcActions.ts
```diff
@@ -181,6 +181,7 @@ export function redeemReputationFromToken(scheme: Scheme, addressToRedeem: strin
       try {
         await arc.web3.eth.sendSignedTransaction(signedTransaction.rawTransaction);
         dispatch(showNotification(NotificationStatus.Success, "Transaction was succesful!"));
```
dkent600: is an `await` needed for `dispatch`?

src/components/Scheme/ReputationFromToken.tsx
```diff
@@ -333,7 +337,7 @@ class ReputationFromToken extends React.Component<IProps, IState> {
                 </div>
                 <div className={schemeCss.redemptionButton}>
                   <button type="submit"
-                    disabled={false}
+                    disabled={this.state.alreadyRedeemed || this.state.redemptionAmount.isZero()}
```
dkent600: It is likely that just the css will not prevent the button from being clicked.  You may also need to set the `onClick` to `undefined`.

src/components/Scheme/ReputationFromToken.tsx
```diff
@@ -343,7 +347,7 @@ class ReputationFromToken extends React.Component<IProps, IState> {
                   <div className={schemeCss.redemptionButton}>
                     <div>Or try our new experimental feature:</div>
                     <button type="submit"
-                      disabled={false}
+                      disabled={this.state.alreadyRedeemed || this.state.redemptionAmount.isZero()}
```
dkent600: Same `onClick` comment as above.

src/actions/arcActions.ts
tibetsprague: no, it is just creating a notification

src/components/Scheme/ReputationFromToken.tsx
tibetsprague: this is not css this is the button disabled property and it will prevent the button from being clicked

src/components/Scheme/ReputationFromToken.tsx
tibetsprague: disabled works fine

src/components/Scheme/ReputationFromToken.tsx
dkent600: OK, I think I need to not review code when I'm not awake yet.

src/actions/arcActions.ts
dkent600: not fully awake when I reviewed this.

PR #1396, opened 2020-02-07
Improve layout of redemptions page header

dkent600:
In the redemptions page header:

* improves positioning of the Redeem All button
* improves layout in general on mobile platforms


dkent600:
Branch should've been based on dev

Issue #1374, opened 2020-02-04
ProposalCard should not show this warning


dkent600:
Seeing this in the console:

> react-dom.development.js:530 Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in WithSubscriptionComponent (created by ProposalCard)
    in ProposalCard (created by SchemeProposalsPage)
    in Transition (created by CSSTransition)
    in CSSTransition (created by Fade)
    in Fade (created by SchemeProposalsPage)
    in div (created by TransitionGroup)
    in TransitionGroup (created by SchemeProposalsPage)
    in div (created by InfiniteScroll)
    in div (created by InfiniteScroll)
    in InfiniteScroll (created by SchemeProposalsPage)
    in div (created by SchemeProposalsPage)
    in div (created by SchemeProposalsPage)
    in div (created by SchemeProposalsPage)
    in div (created by SchemeProposalsPage)
    in SchemeProposalsPage (created by WithSubscriptionComponent)
    in WithSubscriptionComponent (created by Connect(WithSubscriptionComponent))
    in Connect(WithSubscriptionComponent) (created by Route)

@tibetsprague

tibetsprague:
@dkent600 when do you see this?

dkent600:
https://alchemy-staging-rinkeby.herokuapp.com/dao/0x72939947e7a1c4ac94bb840e3304b322237ad1a8/scheme/0x16b81c56297ae81ffa32ed128f17dbf7b108eb9200d2c8110a9ff2e05e095995

PR #1400, opened 2020-02-08
Fix warning about ProposalCard memory leak

tibetsprague:
Fixes https://github.com/daostack/alchemy/issues/1374

Since i couldn't easily reproduce the warning this is a bit of guess, but I think it's the right guess :)

### DIFF
--- a/src/components/Proposal/ProposalData.tsx
+++ b/src/components/Proposal/ProposalData.tsx
@@ -64,6 +64,7 @@ interface IState {
 }
 
 class ProposalData extends React.Component<IProps, IState> {
+  private expireTimeout: any;
 
   constructor(props: IProps) {
     super(props);
@@ -79,8 +80,12 @@ class ProposalData extends React.Component<IProps, IState> {
     // Don't schedule timeout if its too long to wait, because browser will fail and trigger the timeout immediately
     const millisecondsUntilExpires = closingTime(this.props.data[0]).diff(moment());
     if (!this.state.expired && millisecondsUntilExpires < 2147483647) {
-      setTimeout(() => { this.setState({ expired: true });}, millisecondsUntilExpires);
+      this.expireTimeout = setTimeout(() => { this.setState({ expired: true });}, millisecondsUntilExpires);
     }
+  }
+
+  componentWillUnmount() {
+    clearTimeout(this.expireTimeout);
   }
 
   render(): RenderOutput {


src/components/Proposal/ProposalData.tsx
```diff
@@ -79,10 +80,14 @@ class ProposalData extends React.Component<IProps, IState> {
     // Don't schedule timeout if its too long to wait, because browser will fail and trigger the timeout immediately
     const millisecondsUntilExpires = closingTime(this.props.data[0]).diff(moment());
     if (!this.state.expired && millisecondsUntilExpires < 2147483647) {
-      setTimeout(() => { this.setState({ expired: true });}, millisecondsUntilExpires);
+      this.expireTimeout = setTimeout(() => { this.setState({ expired: true });}, millisecondsUntilExpires);
```
dkent600: I don't see a reason for this.  `setState` is mean to be invoked within `componentDidMount`.

src/components/Proposal/ProposalData.tsx
```diff
@@ -79,10 +80,14 @@ class ProposalData extends React.Component<IProps, IState> {
...
     }
```
dkent600: I don't see a reasons for this. 1) `setState` is meant to be called from `componentDidMount`,  2) the console msg indicates that the component has not yet mounted, which would not be the case in `componentDidMount`.

src/components/Proposal/ProposalData.tsx
dkent600: Are you not able to consistently reproduce accoring to the lin [given here](https://github.com/daostack/alchemy/issues/1374#issuecomment-583672201)?

src/components/Proposal/ProposalData.tsx
dkent600: The console message suggests this is happening in `WithSubscriptionComponent`, which calls `setState` in a lot of places.

dkent600: The console message suggests this is happening in `WithSubscriptionComponent`, which calls `setState` in a lot of places.

src/components/Proposal/ProposalData.tsx
tibetsprague: @dkent600 what do you mean? this is inside componentDidMount

tibetsprague: and no I was not able to reproduce. I saw the error happen one time but I dont know when.

src/components/Proposal/ProposalData.tsx
tibetsprague: that error message usually occurs when the component is unmounted, and it usually indicates that something setup in componentDidMount was not removed/unset/cleared in componentWillUnmount. So in this case for example the timeout expires/fires and the component is no longer mounted so then this error fires. Thus this fix.

dkent600: sorry, misread the code, was thinking that you added a call to setTimeout that wasn't there before, but I see that is not the case.

Issue #1329, opened 2020-01-24
bundle optimization: we are including src in the bundle, 


jellegerbrandy:
this adds a useless  667kb

tibetsprague:
hmm @jellegerbrandy, are you sure about this? if i do a build i don't see the src. i think what you are seeing is simply the source map laying out all the files in the app bundle

tibetsprague:
@jellegerbrandy did you see my comment?

jellegerbrandy:
this was in npm run analyze; I'm not seeing it anymore

PR #1401, opened 2020-02-08
Interestingly, setting devtool to source-maps on prod makes the bundl…

tibetsprague:
…e smaller.

The gzipped app bundle goes from 155 KB to 114 KB!
I'm not sure what the default is that it was using before, must have been some inline source maps, so this is better.  And we get to keep source maps on production for debugging

Related to https://github.com/daostack/alchemy/issues/1329


tibetsprague:
yeah so the default is inline source maps, for production one is supposed to change the value so that it generates external .map files which arent in the bundle and only get loaded by the browser when needed. talk about some low hanging fruit...

PR #1420, opened 2020-02-11
ethdenver optimization

jellegerbrandy:
* rewrite retry logic on app load 
* add apollo-retry-link ocnfiguration, so we can tweak it (and tweak it a bit)
*  add a `ETHDENVER_OPTIMIZIATION flag that allows us to tweak things only for ethdenver (we can easily disable it later) . For now, the only thing it does is disable the redemptions button on the top menu on competition urls...


jellegerbrandy:
@dkent600  so the initialize logic in App.tsx throws an error if the connection is not responding or erroring out (after trying several times). But this error does not "bubble up" to the UI - I am not familiar with the logic there. Are you?

dkent600:
@jellegerbrandy 

> so the initialize logic in App.tsx throws an error if the connection is not responding or erroring out (after trying several times). But this error does not "bubble up" to the UI - I am not familiar with the logic there. Are you?

The render code renders according to `this.state.arcIsInitialized`.  In the case of an exception, `this.state.arcIsInitialized` will remain true.  It appears that nothing is reported, except in the console.

This is very old code...predating my work on web3 provider stuff....

src/App.tsx
```diff
@@ -42,19 +45,38 @@ export class App extends React.Component<{}, {
...
+        this.setState({ retryingArc: true });
         // eslint-disable-next-line no-console
+        numberOfAttempts += 1;
+        // retry
+        if (numberOfAttempts >= totalNumberOfAttempts) {
+          // THIS ERROR SHOULD BE SHOWN TO THE USER - where it is intercepted?
+
+          const msg = "Could not connect to the network; please retry later...";
+          this.setState({ error: msg});
+          throw Error(msg);
```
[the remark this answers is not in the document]
dkent600: @jellegerbrandy fixed here: https://github.com/daostack/alchemy/pull/1420/commits/3699812dff6d1e6c3b9d97af66ed529bebe24b4f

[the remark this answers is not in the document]
dkent600: Sorry, somehow that comment got associated with the wrong line of code.  But the commit linked above shows what I was referring to.

src/arc.ts
```diff
@@ -122,59 +124,79 @@ export async function initializeArc(provider?: any): Promise<boolean> {
   let success = false;
   let arc: any;
 
-  try {
```
dkent600: Since this method guarantees no exceptions, it should catch all.  Should restore this try/catch.  Anyone with and interest in seeing the raw details, which would usually only be a technical person, can look at the console.  If we're going to change how this function works, we need to do it thoughtfully, looking at everywhere it is used.

dkent600: fixed here:  https://github.com/daostack/alchemy/pull/1420/commits/fe770fed6f243a02c260cda42220a8d8b957e44a

src/layouts/AppContainer.tsx
```diff
@@ -147,7 +147,7 @@ class AppContainer extends React.Component<IProps, IState> {
   }
 
   private dismissNotif = (id: string) => () => this.props.dismissNotification(id);
-  private headerHtml = ( props: any ): any => <Header {...props} />;
+  private headerHtml = ( props: any ): any => <Header {...props} showRedemptionsButton={false} />;
```
dkent600: why the showRedemptionsButton here?

src/layouts/Header.tsx
```diff
@@ -196,9 +208,11 @@ class Header extends React.Component<IProps, null> {
                 icons={{ checked: <img src='/assets/images/Icon/checked.svg'/>, unchecked: <img src='/assets/images/Icon/unchecked.svg'/> }}/>
             </div>
           </TrainingTooltip>
-          <div className={css.redemptionsButton}>
-            <RedemptionsButton currentAccountAddress={currentAccountAddress} />
-          </div>
+          {
+            this.props.showRedemptionsButton ? <div className={css.redemptionsButton}>
+              <RedemptionsButton currentAccountAddress={currentAccountAddress} />
+            </div> : <div></div>
```
dkent600: should not have an empty div here, can screw up the layout

src/settings.ts
```diff
@@ -1,5 +1,8 @@
 import BurnerConnectProvider from "@burner-wallet/burner-connect-provider";
 
+// hide some data to send less queries
+export const ETHDENVER_OPTIMIZATION = true;
```
dkent600: makes this true for every platform?  should this be just for when the NETWORK is xdai?

src/layouts/Header.tsx
dkent600: fixed with: https://github.com/daostack/alchemy/pull/1420/commits/6e7d4665d341999f477a1009d51ed9dbe48169c7

src/layouts/AppContainer.tsx
dkent600: fixed with:  https://github.com/daostack/alchemy/pull/1420/commits/6e7d4665d341999f477a1009d51ed9dbe48169c7

src/arc.ts
jellegerbrandy: I'm not really agreeing with this at all, but we can discuss at a more opportune time :-)

package.json
```diff
@@ -66,7 +66,7 @@
     "start-docker": "node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js  --config webpack.docker.config.js",
     "start-prod": "cross-env NODE_ENV=production NETWORK=main node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js  --config webpack.dev.config.js",
     "start-staging": "cross-env NODE_ENV=staging NETWORK=rinkeby node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --config webpack.dev.config.js",
-    "start-xdai": "cross-env NODE_ENV=production NETWORK=xdai node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --config webpack.prod.config.js",
+    "start-xdai": "cross-env NODE_ENV=staging NETWORK=xdai node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --config webpack.dev.config.js",
```
tibetsprague: ?

src/components/Shared/withSubscription.tsx
```diff
@@ -95,7 +95,7 @@ const withSubscription = <Props extends ISubscriptionProps<ObservableType>, Obse
     public async setupSubscription(observable?: Observable<any>) {
 
       this.teardownSubscription();
-      console.log(`setup subscription with id ${options.wrappedComponent.name}`);
+      // console.log(`setup subscription with id ${options.wrappedComponent.name}`);
```
tibetsprague: ?

package.json
dkent600: when we run xdai locally we do it like we do when we run staging locally.

src/arc.ts
dkent600: yes, refactor that if desired when we can take the time to do it right.

package.json
jellegerbrandy: yeah - so these commands are not used for deployment, they are just conveient for running locally...

PR #1434, opened 2020-02-13
Fix travis errors

jellegerbrandy:
which should never have been merged to dev :-/


src/arc.ts
```diff
@@ -35,7 +35,8 @@ export function targetedNetwork(): Networks {
       return "rinkeby";
     }
     case "mainnet":
-    case "production" : {
+    case "production":
+    case undefined : {
```
tibetsprague: i think the fix is case 'main' actually, look at package.json... unless you are sure that it is undefined

PR #1435, opened 2020-02-13
Pre fetch contracts on prod

tibetsprague:
not quite sure how to test this exactly on a production server


src/settings.ts
```diff
@@ -15,6 +9,109 @@ function isMobileBrowser(): boolean {
   return check;
 }
 
+function getWeb3ConnectProviderOptions(network: string) {
+  if (typeof(window) === "undefined") {
+    return null;
+  }
+
+  const BurnerConnectProvider: any = null; // require("@burner-wallet/burner-connect-provider");
```
dkent600: Web3Connect won't work without an imported module, last I checked.  Definitely a good idea to lazy-load them and keep them in separate chunks, but I don't think `null` is going to work.

jellegerbrandy:
I create this preview app to test this (node_env production, network xdai): https://test-contractinfo-preloading.herokuapp.com/

Not sure what happend with the wallet loading in this PR - maybe committed by mistake?

src/settings.ts
tibetsprague: oops, i forgot to uncomment that

PR #1437, opened 2020-02-14
Fix travis

tibetsprague:


webpack.dev.config.js
```diff
@@ -88,7 +88,7 @@ module.exports = merge(baseConfig, {
     new webpack.NamedModulesPlugin(),
 
     new webpack.EnvironmentPlugin({
-      NETWORK: process.env.NETWORK,
+      NETWORK: "ganache",
```
dkent600: Does this not break local builds against staging, etc, that are dev builds against non-ganache networks?

webpack.prod.config.js
```diff
@@ -93,7 +93,7 @@ plugins: [
     }),
 
     new webpack.EnvironmentPlugin({
-      NETWORK: process.env.NETWORK,
+      NETWORK: "main",
```
dkent600: What if want to run a prod build against another network?  Can't do it with this change.

webpack.dev.config.js
tibetsprague: no, that is the default value if process.env.NETWORK is not set, this is how you use EnvironmentPlugin

webpack.prod.config.js
tibetsprague: that is the default value if process.env.NETWORK is not set, this is how you use EnvironmentPlugin

Issue #1454, opened 2020-02-19
Incorrect regex to check hex in CreateUnknownGenericSchemeProposal.tsx


shekhar-shubhendu:
Location: https://github.com/daostack/alchemy/blob/dev/src/components/Proposal/Create/SchemeForms/CreateUnknownGenericSchemeProposal.tsx#L140

Used Hex: `0x[0-9a-e]+`

What it should be: `0x[0-9a-f]+`

jellegerbrandy:
Fixed in #1455

PR #1455, opened 2020-02-19
Fixing the hex byte pattern regex in Generic Scheme Proposal Form

dadaphl:
to allow character `f`
see #1454


jellegerbrandy:
Many thanks @dadaphl  :heart:

PR #1467, opened 2020-02-26
963 esc to close modals

tibetsprague:


tibetsprague:
Also fixed a couple other little things here

PR #1469, opened 2020-03-03
Refactor modals

tibetsprague:
Create ModalPopup component to bundle common modal stuff

Share common modal styles across all modals.
All ModalPopups can be closed by pressing esc or clicking on the background
All modals have a header, body and optional footer

### DIFF
--- a/src/components/Proposal/Create/CreateProposalPage.tsx
+++ b/src/components/Proposal/Create/CreateProposalPage.tsx
@@ -61,6 +61,8 @@ class CreateProposalPage extends React.Component<IProps, IStateProps> {
   }
 
   public async componentDidMount() {
+    document.addEventListener("keydown", this.handleKeyPress, false);
+
     Analytics.track("Page View", {
       "Page Name": Page.CreateProposal,
       "DAO Address": this.props.daoAvatarAddress,
@@ -77,6 +79,17 @@ class CreateProposalPage extends React.Component<IProps, IStateProps> {
     }
 
     this.setState(newState);
+  }
+
+  public componentWillUnmount(){
+    document.removeEventListener("keydown", this.handleKeyPress, false);
+  }
+
+  private handleKeyPress = (e: any) => {
+    // Close modal on ESC key press
+    if (e.keyCode === 27) {
+      this.doClose();
+    }
   }
 
   public render(): RenderOutput {
--- a/src/components/Shared/ModalPopup.scss
+++ b/src/components/Shared/ModalPopup.scss
@@ -1,0 +1,56 @@
+.modalWindow {
+  position: absolute;
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%,-50%);
+  transition: all .25s ease;
+  border-radius: 15px 15px 15px 15px;
+  z-index: 99999999999999999;
+}
+
+.header {
+  background-color: rgba(18, 46, 91, 1.000);
+  padding: 12px 20px;
+  color: $white;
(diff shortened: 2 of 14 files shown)


src/components/Proposal/Create/CreateProposalPage.tsx
```diff
@@ -61,6 +61,8 @@ class CreateProposalPage extends React.Component<IProps, IStateProps> {
   }
 
   public async componentDidMount() {
+    document.addEventListener("keydown", this.handleKeyPress, false);
```
dkent600: Would this be better in `ModalPopup`?  If we don't think it appropriate for _all_ popups (why not?)  then it can be optionally enabled.

src/components/Shared/ModalPopup.scss
```diff
@@ -0,0 +1,56 @@
...
+  position: absolute;
+  left: 50%;
+  top: 50%;
+  transform: translate(-50%,-50%);
+  transition: all .25s ease;
+  border-radius: 15px 15px 15px 15px;
+  z-index: 99999999999999999;
+}
+
+.header {
```
dkent600: Can we use nesting (as a general best practice, clarity, robustness)?  I think all the classes here might go under `.ModalWindow`.

src/components/Proposal/Create/CreateProposalPage.tsx
tibetsprague: CreateProposalPage cant use ModalPopup (as currently designed) since it uses <ModalRoute>, not <Modal>

src/components/Shared/ModalPopup.scss
tibetsprague: is nesting better? I often find it more confusing by far

dkent600: My take:  It is less confusing for me.  The nesting of classes maps 1-to-1 with the nesting of html elements using the classes, so it is easier to find the classes associated with specific elements, and vice-versa.

You know exactly there a class is supposed to be used, and nesting enforces this in the compiler.

It cleans up situations where there are duplicate class names used in different contexts.  

Nesting keeps the scss nice and clean, less verbose -- in the editor you can collapse a whole hierarchy of classes that don't relate to the ones you're using.

tibetsprague: so everything should always be nested fully within every single other top level component? it ends up becoming very hard to figure out what is nested where at what level. i guess it would be clean if you perfectly mapped the DOM nesting to the SCSS nesting, but that would be a huge task to get there in our current codebase. and also it feels kind of pointless to nest everything inside the wrapper component, it just adds unnecessary levels of indentation.

dkent600: I don't understand "fully within every single other top level component".  At any rate, feel free to do this as you feel best.  :-)

src/components/Proposal/Create/CreateProposalPage.tsx
dkent600: Ah, I see this is already in `ModalPopup`

dkent600: Does this mean that none of the many popups spawned by CreateProposalPage are use the common code you created here? Or, to what extent do they share code with other popups?

tibetsprague: that is true, they are not using the shared code at all right now. one thing that could potentially be shared is some of the styling... there is certainly more work that could be done there

Issue #1337, opened 2020-01-28
Clicking on proposal in history view should OPEN NEW TAB


orishim:
Not able to right click a proposal and open it in a new tab from the history view.

I would even say that default behavior for clicking a a proposal should be to open in new tab, since scrolling and waiting to load takes a lot of time.

jellegerbrandy:
Default behavior will not be to open in a new tab.

But "crtl-click" should indeed  work

PR #1474, opened 2020-03-03
in history, open proposal in new tab on ctrl-click

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1337


dkent600:
@tibetsprague 

> on a Mac normally shift click would do this, can we add that too?

I added code to do that but don't have a way to test

PR #1478, opened 2020-03-06
Enforce required permissions when adding/editing known Schemes

tibetsprague:
Also move all scheme utilities into a new lib/schemeUtils.ts file


src/lib/schemeUtils.ts
```diff
@@ -0,0 +1,209 @@
...
+
+import { getArc } from "../arc";
+
+export enum SchemePermissions {
+  None = 0,
+  IsRegistered = 1, // Always added by default in the controller
+  CanRegisterSchemes = 2,
+  CanAddRemoveGlobalConstraints = 4,
+  CanUpgradeController = 8,
+  CanCallDelegateCall = 0x10,
```
dkent600: There is also a metadata permission that I'm querying Oren about.

tibetsprague: sounds like this is all good for now

dkent600: I haven't seen a response from Oren about the metadata permission (that it overlaps the delegatecall permission).

tibetsprague: the responded and said they are the same

dkent600: Ah, got it, looks good

Issue #739, opened 2019-05-26
wrong "pre boost" button on queued proposal.


orenyodfat:
![Screen Shot 2019-05-26 at 10 55 59](https://user-images.githubusercontent.com/5994398/58378937-09aa5500-7fa5-11e9-87f0-dee381cf15e6.png)

proposal confidence should be > (not >-=) daothreshold in order to pre boost.

orenyodfat:
@jellegerbrandy @tibetsprague is that already been fixed ?

jellegerbrandy:
yes

orenyodfat:
this bug is still valid 
see https://alchemy.daostack.io/dao/0x440583455bcd85ab2bd429c015d3aabcae135f0a/scheme/0x949943d31b8db2987783563573da3a573ac93b6406ccabe614b89d462fc69965

dkent600:
The PreBoost button is visible when `proposalState.upstakeNeededToPreBoost < 0`.

It seems that `upstakeNeededToPreBoost >= 0` is expected to mean that `score(_proposalId) <= confidenceThreshold`, in other words, that preboosting won't happen.

I think that if there is an error it is in how [this client lib code](https://github.com/daostack/client/blob/6d2a9bf719cba9d88c0900a3b8a9d7fc448470ce/src/proposal.ts#L512) is computing `upstakeNeededToPreBoost`.

The client lib test code that might confirm the logic is currently commented out.

@jellegerbrandy

dkent600:
Client lib ticket:  https://github.com/daostack/client/issues/415

dkent600:
@orenyodfat It is not possible to see what was happening in your latest report above, since the data at the url you gave has now completely changed.  A screenshot would have been much better, one that is very clear about what was incorrect.  In the meantime, if you have any clues you can give about what precisely was happening that should not have, please let us know here, thanks.

dkent600:
I don't honestly find anything wrong in the client lib computations.  I don't see much wrong with the Alchemy UI code either, though I have addressed a few things in [this PR](https://github.com/daostack/alchemy/pull/1482). Otherwise, not knowing what is happening in the GUI, I can't really go much further.

orenyodfat:
looks like this it solved.
[...]

PR #1482, opened 2020-03-06
improve preboostable logic, feedback and layout

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/739
...hopefully.

Refer to the logic described in the code comments added in this client lib PR:  https://github.com/daostack/client/pull/416/files

* eliminates some unnecessary constructions of BN
* adds '=' (for ">=") in text referring to `downStakeNeededToQueue`
* improves layout of amount-needed feedback in StakeButtons.scss
* doesn't show amount-needed feedback unless it is possible to boost, in StakeButtons.tsx
* in PreTransactionModal, when downstaking, don't show amount-to-boost, rather show amount-to-stake-to-unboost


dkent600:
@jellegerbrandy I am premising my findings here based on the logic given in the new comments in https://github.com/daostack/alchemy/issues/739.  Until you've approved those comments, I don't think we should approve this PR.  So I'm blocking this one until that point.

src/components/Proposal/Staking/BoostAmount.tsx
```diff
@@ -26,32 +25,34 @@ export default class BoostAmount extends React.Component<IProps, null> {
       [css.wrapper]: true,
       [css.detailView]: detailView,
     });
+    const nothing = <span className={css.boostedAmount}><b>&nbsp;</b></span>;
 
     return (
       <div className={wrapperClass}>
         {
-          proposal.stage === IProposalStage.Queued && !expired && proposal.upstakeNeededToPreBoost.gte(new BN(0)) ?
+          !expired ? ((proposal.stage === IProposalStage.Queued && proposal.upstakeNeededToPreBoost.gten(0)) ?
```
tibetsprague: I dont think this is right, cant you still boost/unboost a pre-boosted proposal that is expired? or do we want to turn that off

src/components/Proposal/Staking/BoostAmount.tsx
```diff
@@ -26,32 +25,34 @@ export default class BoostAmount extends React.Component<IProps, null> {
...
+                &gt; {formatTokens(proposal.upstakeNeededToPreBoost, "GEN")} on Pass to boost
               </b>
             </span>
-            : proposal.stage === IProposalStage.PreBoosted && proposal.downStakeNeededToQueue.lte(new BN(0)) ?
+            : proposal.stage === IProposalStage.PreBoosted && proposal.downStakeNeededToQueue.lten(0) ?
               <span className={css.boostedAmount}>
                 <b>
                   {detailView ? <img src="/assets/images/Icon/Boost-slate.svg" /> : ""}
-                &gt; {formatTokens(proposal.downStakeNeededToQueue.abs(), "GEN")} Pass to stay boosted
+                &gt; {formatTokens(proposal.downStakeNeededToQueue.abs(), "GEN")} on Pass to stay boosted
```
tibetsprague: does this fit?

dkent600: yes

src/components/Proposal/ActionButton.tsx
```diff
@@ -230,13 +230,13 @@ class ActionButton extends React.Component<IProps, IState> {
           /> : ""
         }
 
-        { proposalState.stage === IProposalStage.Queued && proposalState.upstakeNeededToPreBoost.lt(new BN(0)) ?
+        { proposalState.stage === IProposalStage.Queued && proposalState.upstakeNeededToPreBoost.ltn(0) ?
```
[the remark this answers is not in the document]
dkent600: has always worked for me.  testing the functionality seems to work

src/components/Proposal/Staking/BoostAmount.tsx
dkent600: I was misinterpreting what "expired" meant (forgetting that it often means expired only within the context of the current phase, but not necessarily that the proposal is dead, and not necessarily that the expired state is even represented yet in the blockchain).

So good catch.  I have reverted the change.

Issue #1343, opened 2020-01-30
add tor.us to web3connect


jellegerbrandy:

dkent600:
Torus

dkent600:
Forgive me, I commitred this to the Competition branch by accident.

jellegerbrandy:
Why is this blocked?

dkent600:
blocked because:  https://github.com/torusresearch/torus-embed/issues/100#issuecomment-588980747

dkent600:
Blocked because I have insufficient auth to access heroku to determine why the review app is not building.  @orenyodfat

dkent600:
@orenyodfat FYI pulling in the latest torus does not fix the problem.  I will let them know...

orenyodfat:
Heroku review app should work with this fix https://github.com/daostack/alchemy/commit/5c93b68c2c5b4edd5519bd4bcb99b33886fa3a22 
(tested on other test branch)

PR #1489, opened 2020-03-09
add torus web provider

dkent600:
New PR based against the dev branch (previous was against competition)

Resolves: https://github.com/daostack/alchemy/issues/1343

* also removes the lint rule for sorting imports which is driving me crazy.  need to fix it in [another issue](https://github.com/daostack/alchemy/issues/1533).


dkent600:
@orenyodfat FYI Pulling in the latest version of torus does not fix the CSS problem

package.json
```diff
@@ -53,7 +53,7 @@
   },
   "scripts": {
     "analyze": "NODE_ENV=production NETWORK=main ANALYZE=1 node --max_old_space_size=4096 node_modules/.bin/webpack --config webpack.prod.config.js --progress --colors",
-    "build": "webpack --config webpack.prod.config.js",
+    "build": "node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js --config webpack.prod.config.js",
```
orenyodfat: this does not seems to solve the build issue ?
why this change ?

dkent600: That is an attempt to fix the review app build problem and also the ability to build for prod locally.  It fixes the ability to build for prod locally, but doesn't seem to help with the review app build.

orenyodfat:
Heroku review app should work with this fix 5c93b68
(tested on other test branch) .

orenyodfat:
please use torus.setProvider to support other network than mainnet (default) https://docs.tor.us/api-reference/class#init
https://docs.tor.us/api-reference/ethereum-api#setprovider

package.json
```diff
@@ -84,6 +84,7 @@
     "@fortawesome/react-fontawesome": "^0.1.3",
     "@portis/web3": "^2.0.0-beta.43",
     "@sentry/browser": "^5.0.8",
+    "@toruslabs/torus-embed": "^1.2.0",
```
orenyodfat: please bump to 1.2.1

app.json
```diff
@@ -3,6 +3,10 @@
   "description": "",
   "scripts": {},
   "env": {
+    "NODE_OPTIONS": {
+      "required": true,
+      "value": "--max_old_space_size=2560"
```
orenyodfat: is this needed ?

dkent600:
@orenyodfat The problem with using rinkeby appears to be a bug in web3modal (formerly web3connect).  I reported it here: https://github.com/Web3Modal/web3modal/issues/130

package.json
dkent600: Bumping to 1.2.3, the latest.

dkent600: Also grabbing the latest web3modal, no long in beta.

PR #1492, opened 2020-03-10
Fix Windows Install

dOrgJelli:
The fix needed was to upgrade the `web3` dependency DAOcreator's Lib uses to 1.2.4. The old version included a `preInstall` script that used the `rm` unix command.


dkent600:
Thanks @dOrgJelli !

PR #1494, opened 2020-03-11
release-2020-03-11

dkent600:
release-2020-03-11


test/integration/proposal-genericSchemeStandardBounties.ts
```diff
@@ -0,0 +1,77 @@
+/*
```
orenyodfat: why this test  file is comment out ? @tibetsprague @dkent600 ?

tibetsprague: good question, I guess he never got it to work? or maybe never worked on it. probably we shouldn't have let this PR through without the tests... but we did this time

orenyodfat: ok. please open an issue to enable this test (and make it work:)) .

Issue #1488, opened 2020-03-09
high severity vulnerabilities


dkent600:
found 185 vulnerabilities (156 low, 13 moderate, 16 high)

PR #1500, opened 2020-03-11
improve package vulnerabilities

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1488

Best we can do for now:

image


package.json
```diff
@@ -53,7 +53,7 @@
   },
   "scripts": {
     "analyze": "NODE_ENV=production NETWORK=main ANALYZE=1 node --max_old_space_size=4096 node_modules/.bin/webpack --config webpack.prod.config.js --progress --colors",
-    "build": "webpack --config webpack.prod.config.js",
+    "build": "node --max_old_space_size=4096 ./node_modules/webpack/bin/webpack.js  --config webpack.prod.config.js",
```
tibetsprague: @jelle changed this because builds on Heroku were breaking with this....

dkent600: I had multiple reasons to believe it would fix the problem I'm seeing with review apps.  but I'm not sure it has.  waiting to get permissions on heroku to see what is going on.

dkent600: But note that this *does* fix the ability to build for prod locally.

tibetsprague:
Which we may not need to do. But we can try again and if it breaks heroku again then find a different solution? Maybe two different build commands?

Sent by m

package.json
dkent600: It is not so unusual for me to want to run a local prod build

dkent600:
> this branch give same result :
> Screen Shot 2020-03-15 at 14 27 45

See the original ticket that reports the original numbers.  16 severe.

PR #1516, opened 2020-03-15
fix CreateKnowGenericSchemeProposal - uint256 field should be rendered as "string" (not as "number")

### LINKED ISSUE #1507: Submit Proposal on ETHregistrar does not do anything
Screenshot 2020-03-13 at 17 57 52

nothing happens after clicking on submit proposal:

Screenshot 2020-03-13 at 18 15 47

orenyodfat:
fix #1507


src/genericSchemeRegistry/index.ts
```diff
@@ -98,6 +98,11 @@ export class ActionField {
       userValue = userValue.trim();
     }
 
+    if (this.type == "uint256") {
+      return userValue.toString();
+    }
+    console.log(userValue.toString(),this.type);
```
tibetsprague: remove console.log

src/genericSchemeRegistry/index.ts
```diff
@@ -98,6 +98,11 @@ export class ActionField {
       userValue = userValue.trim();
     }
 
+    if (this.type == "uint256") {
```
dkent600: Should be `===`

dkent600: Should confirm that this doesn't break anything.

Particularly:  What if `this.decimals` is set?  It won't be executed.

Issue #1484, opened 2020-03-06
DAO search should not stop responding after loading more DAOs


dkent600:
In staging:

1. go to All DAOS page
1. scroll down 2 or 3 times, enough to pull in some more DAOs
1. type something in the DAOs filter textbox

RESULT:
The filter will work on the first character typed, but after that, no longer responds

dkent600:
@tibetsprague

tibetsprague:
cannot reproduce

dkent600:
If I mess with it enough I can still get it to happen

dkent600:
Another issue that is easier to reproduce:

In staging,

1. go to the AllDaos page, fresh
1. don't scroll down
1. type to filter on "Buffi"

Result:
Nothing happens

Expect:
Should see the BuffiDaos

tibetsprague:
hmm yeah thats trickier, having the search access the server... I guess I can do that

dkent600:
Maybe we can just disable the filter until all the daos have appeared, and always fetch all the daos on page load?

tibetsprague:
we cant fetch all DAOs, the subgraph seems to limit fetching to 100. also I dont think we would want to load more than 100 at a time. so we have to do server side searching if we want it to work right

dkent600:
I was imagining automatically fetching one page at a time until there is nothing left to fetch, _as if_ the user were scrolling down.

tibetsprague:
that could work...

PR #1532, opened 2020-03-23
When searching DAOs check server not just current ones in client

tibetsprague:
To make sure we find all DAOs even if they have not yet been loaded on the client yet

Also only search after more than 2 characters typed in box

TODO: search different cases of the search string?

Fixes: https://github.com/daostack/alchemy/issues/1484

### DIFF
--- a/src/components/Daos/DaosPage.tsx
+++ b/src/components/Daos/DaosPage.tsx
@@ -14,6 +14,7 @@ import { Link } from "react-router-dom";
 import { IRootState } from "reducers";
 import { IProfileState } from "reducers/profilesReducer";
 import { combineLatest, of } from "rxjs";
+import { first } from "rxjs/operators";
 
 import DaoCard from "./DaoCard";
 import * as css from "./Daos.scss";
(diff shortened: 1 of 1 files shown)


src/components/Daos/DaosPage.tsx
```diff
@@ -56,15 +59,34 @@ class DaosPage extends React.Component<IProps, IState> {
...
+      this.setState({ searchDaos: foundDaos });
+    } else {
+      this.setState({ searchDaos: [] });
+    }
   }
 
   public render(): RenderOutput {
     const { currentAccountProfile, data, fetchMore } = this.props;
-    const search = this.state.search.toLowerCase();
+    const search = this.state.search.length > 2 ? this.state.search.toLowerCase() : "";
```
dkent600: Why do we need `state.search` when we have `state.searchDAOs`?  Does the input control really need to get its value from the state?

We don't appear to be using the local `search` variable at all.

tibetsprague: the search var is used to actually filter the DAOs in the UI, yes it needs to be tracked in the state

dkent600: OK, I see now that it is being used, couldn't see that before, maybe because the lines are long.  But the check for the length of the array is redundant.

PR #1534, opened 2020-03-25
upgrade node version

dkent600:
Upgrades to 11.15.0, the highest version that builds.  I spent a lot of time trying to get 12.x to work, without ultimate success.  I think 11.15.0 accomplishes the spirit of the request.

Removes the `cheap` devtool in the prod webpack config and still builds.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     "url": "git+https://github.com/daostack/alchemy.git"
   },
   "engines": {
-    "node": "10.16.0",
-    "npm": "6.11.3"
+    "node": "11.0.0",
+    "npm": "6.4.1"
   },
   "jest": {
     "testURL": "http://127.0.0.1:3000/",
--- a/package.json
+++ b/package.json
@@ -10,8 +10,8 @@
     "url": "git+https://github.com/daostack/alchemy.git"
   },
   "engines": {
-    "node": "11.0.0",
-    "npm": "6.4.1"
+    "node": "11.15.0",
+    "npm": "6.7.0"
   },
   "jest": {
     "testURL": "http://127.0.0.1:3000/",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@
     - On the All DAOS page, make sure the filter finds all DAOs even if they have not yet been loaded on the client
     - Fix error when navigating to the All Redemptions when there exist more than 100 DAOs.
 
+## 0.10.4
+  - Features Added
+    - Recommended node.js version to build Alchemy is set to 11.15.0.
+
+  - Bugs Fixed
+
 ## 0.10.2
   - Features Added
     - Fix bug when entering numbers for uint256 data types in the GenericScheme new proposal modal
--- a/package-lock.json
+++ b/package-lock.json
@@ -243,9 +243,9 @@
       }
     },
     "@babel/generator": {
-      "version": "7.9.3",
-      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.9.3.tgz",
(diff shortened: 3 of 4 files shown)


dkent600:
> what issues were you seeing with 12?

Errors all relating to node-gyp, scrypt and node-boilerplate.  Some package or packages are out-of-sync.

dkent600:
> yes, 11 might be better that 12 ..
> though we must have the ability to build alchemy with the latest recommended node . 12.6.1.
> 
> which action (or dependencies)are required to solve it ?

@orenyodfat I am proposing to do this now and worry about 12 in a separate issue.  11 takes care of our immediate needs.

orenyodfat:
> > yes, 11 might be better that 12 ..
> > though we must have the ability to build alchemy with the latest recommended node . 12.6.1.
> > which action (or dependencies)are required to solve it ?
> 
> @orenyodfat I am proposing to do this now and worry about 12 in a separate issue. 11 takes care of our immediate needs.

why ? what are our immediate needs ?  which action (or dependencies)are required to solve it ?

dkent600:
> why ? which immediate needs it is solving? 

This upgrade was originally motivated by needing to solved the problem with the size of the heroku slug.  You, I and heroku support (see the [most recent comment here](https://help.heroku.com/tickets/837443#event_fa9027b8-99d4-4937-9cea-75eabc8d0311)) all recommended upgrading beyond v10 as possibly the best way to do this.  At the moment we are working-around the problem with a potentially fragile hack.

Upgrading node to version 11 is all we need to obtain memory reductions that may reduce the fragility of the work-around and maintain our position below the slug size limit.  We don't need version 12 to obtain these memory reductions.

This is not a huge thing since we are already below the slug size limit.  But upgrading to version 12 has even less immediate benefit.

I have already spent a over a day trying to get v12 to build after two weeks of working on nothing but super boring infrastructure issues.  I would like a break from that.

dkent600:
> why ? which immediate needs it is solving? 

@orenyodfat 

This upgrade was originally motivated by needing to solved the problem with the size of the heroku slug.  You, I and heroku support (see the [most recent comment here](https://help.heroku.com/tickets/837443#event_fa9027b8-99d4-4937-9cea-75eabc8d0311)) all recommended upgrading beyond v10 as possibly the best way to do this.  At the moment we are working-around the problem with a potentially fragile hack, a hack that also _significantly diminishes our ability to debug the preview apps_.  We would like to eliminate that hack.

Upgrading node to version 11 may all by itself be able to obtain memory reductions sufficient to enable us to eliminate the hack.  We don't need version 12 to obtain these memory reductions.

I have already spent a over a day trying to get v12 to build after two weeks of working on nothing but super boring infrastructure issues.  I would like a break from that.

orenyodfat:
`Upgrading node to version 11 may all by itself be able to obtain memory reductions sufficient to enable us to eliminate the hack. We don't need version 12 to obtain these memory reductions.` 

please remove the cheap-source-maps patch and confirm that torus review app can be build without an issue.

package.json
```diff
@@ -10,8 +10,8 @@
     "url": "git+https://github.com/daostack/alchemy.git"
   },
   "engines": {
-    "node": "10.16.0",
-    "npm": "6.11.3"
+    "node": "11.15.0",
+    "npm": "6.7.0"
```
orenyodfat: why npm is downgraded ?

dkent600: Probably because that is the minimum version that will work with node 11.15.0.

dkent600:
@orenyodfat Forgive me, in my comment above I related the node version to the slug size problem, but in fact is related to the problem running out of memory when building the review app for the Torus PR.

package.json
orenyodfat: but before we used 6.11.3 . why not keeping that ?

dkent600: Because this is better for our users, gives them more options.

dkent600:
It is looking like using node v11.15.0 fixes the problem without having to change the sourceMap devtool.

PR #1543, opened 2020-03-26
Windows Paths Fix ('\' vs '/')

dOrgJelli:


dkent600:
> huh, why does this fix things? wasn't it working like this previously on windows?

We are pulling in a new version of @dorgtech/daocreator-ui

dOrgJelli:
@tibetsprague @dkent600 so it turns out it wasn't working before and this issue was already present, it was just failing on non-blocking things. If you look at the code changes, I also updated the source-map-loader. If you leave the source-map-loader includes as they used to be, it would generate a bunch of warnings because those regex epxressions weren't being picked up and used. This is because the regex expressions used forward slashes instead of back slashes, causing them to fail when used with windows fs paths.

What needs to be done is the path string with forward slashes needs to be given first to node's runtime so it can correct the path string to match the OS node is running on (windows).

Issue #1540, opened 2020-03-26
Torus login doesn't let you switch accounts


EzraWeller:
After logging in once with Torus on rinkeby (Chrome), I see no way to switch which Torus account I am logged in with (this is a problem in this case because the default account has no Eth/Rep/etc.).

Logging out and then logging back in just gives me the same default account, and even going to the Torus site, importing one of the test accounts using private key (which seems to work on that end) doesn't show up in Alchemy.

I hope this is a clear enough explanation--I wasn't able to take good screenshots of the mouseover information.

dkent600:
This seems to be a limitation of Torus.  Will try to find out more.

dkent600:
Currently there are only two ways to supposedly get this functionality:

1) enable the Torus module to automatically display a button in our GUI somewhere.  I assume we could write CSS to control it's position and appearance, but limited in position (like not in a dropdown menu).

2) write our own custom button that we could put anywhere (like in a menu) that would manually make the call into the Torus module to display the necessary GUI.

I have let Torus know that we would prefer that they manage this button themselves.

Issue #1542, opened 2020-03-26
Torus should not list Metamask as the provider


dkent600:
When logged in with Torus on rinkeby, hovering over the account in the top right lists Metamask as the provider.  It should list "Torus" or something.

I created a ticket on this in Web3Modal:  https://github.com/Web3Modal/web3modal/issues/130#issuecomment-602942907

And Torus has submitted a PR on it:
https://github.com/Web3Modal/web3modal/pull/133

dkent600:
The Web3Modal PR was merged, presumedly giving us the fix for this, but it breaks our code as `getProviderInfo` is not longer being exported.  Have contacted them about this via PM.

orenyodfat:
any alternative they provide that ?

dkent600:
> any alternative torus provide that ?

@orenyodfat I'm sorry I don't understand the question.  Can you rephrase?

FYI the method doesn't come from Torus, it comes from Web3Modal, and that is where it must come from.

PR #1545, opened 2020-03-27
add custom Torus configuration button

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1540

### DIFF
--- a/src/layouts/Header.tsx
+++ b/src/layouts/Header.tsx
@@ -1,6 +1,6 @@
 import * as uiActions from "actions/uiActions";
 import { threeBoxLogout } from "actions/profilesActions";
-import { enableWalletProvider, getAccountIsEnabled, getArc, logout, getWeb3ProviderInfo } from "arc";
+import { enableWalletProvider, getAccountIsEnabled, getArc, logout, getWeb3ProviderInfo, getWeb3Provider, providerHasConfigUi } from "arc";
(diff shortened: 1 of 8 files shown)


orenyodfat:
it say "open metamask configuration" . is that issue #1542 ?

dkent600:
> it say "open metamask configuration" . is that issue #1542 ?

@orenyodfat  Yes

orenyodfat:
should we add tool tip or something near this button..so the user will be aware of that..? it is difficult to find it.

dkent600:
> should we add tool tip or something near this button..so the user will be aware of that..? it is difficult to find it.

The button already has a tooltip.    Where are you looking for the button and not finding it?

EzraWeller:
This looks great to me, was able to import an account and switch between them, with the result showing up in Alchemy.

src/layouts/Header.tsx
```diff
@@ -175,6 +176,7 @@ class Header extends React.Component<IProps, null> {
     const daoAvatarAddress = dao ? dao.address : null;
     const accountIsEnabled = getAccountIsEnabled();
     const web3ProviderInfo = getWeb3ProviderInfo();
+    const web3Provider = getWeb3Provider();
```
tibetsprague: kind of confusing to have both web3ProviderInfo and web3Provider, but ok

Issue #1419, opened 2020-02-11
alchemy should warn if the subgraph is not getting updates


jellegerbrandy:
This requires getting block info from the subgraph (which is not available form the subgraph itself, but from the "subgraph of subgraphs")

tibetsprague:
any more info to get started on this? how do i do this @orenyodfat @ben-kaufman ?

ben-kaufman:
Hi @tibetsprague 

The URL to access this "subgraph of subgraphs" is: `https://api.thegraph.com/index-node/graphql`. You can see here an example of how I interact with it: `https://github.com/daostack/alchemybots/blob/aaff67cb9f88e7476fdb1eedc9b46fab3aca51ec/check-subgraph-status.js#L142`.

dkent600:
@ben-kaufman This github link you gave above is 404.

ben-kaufman:
@dkent600 They're POST only.

dkent600:
@ben-kaufman Can you please send a url for "an example of how I interact with it" that we can GET?

ben-kaufman:
No, there is none that allows GET requests. But you could use http://altair.sirmuel.design/ which lets you play with it nicely.

tibetsprague:
@ben-kaufman the link i want to see is the one you said shows an example of how to interact with it. that is also a 404

tibetsprague:
oh, @ben-kaufman i think i just dont have access to that repo. @leviadam can you add me to alchemybots?

dkent600:
> No, there is none that allows GET requests. But you could use http://altair.sirmuel.design/ which lets you play with it nicely.

@ben-kaufman @tibetsprague that Altair app was blocked by my virus detector (Windows Defender) as containing a threat.

ben-kaufman:
Strange, so you can either ignore the warning or try another one. This worked for me too and is online soo no download: https://lucasconstantino.github.io/graphiql-online/

imolorhe:
> that Altair app was blocked by my virus detector (Windows Defender) as containing a threat.

@dkent600 I believe you mean that Windows considered the file as not trustworthy because the application isn't signed 🙂not necessarily that it contains a virus.

Ps: You also have the options of using Altair as a browser extension in both Chrome and Firefox.

orenyodfat:
pr #1546 have an issue that it shows errors where not suppose to.

dkent600:
[...]

PR #1546, opened 2020-03-29
Show warning if subgraph is not updating anymore

tibetsprague:
Polls every 10 seconds

Fixes: https://github.com/daostack/alchemy/issues/1419


src/arc.ts
```diff
@@ -80,6 +81,18 @@ async function _getCurrentAccountFromProvider(web3?: any): Promise<string> {
   return accounts[0] ? accounts[0].toLowerCase() : null;
 }
 
+/**
+ * Return the most recently synced block from web3
+ */
+export async function getCurrentBlock(web3?: any): Promise<any> {
```
orenyodfat: should this function exported ? why ?

src/arc.ts
```diff
@@ -586,6 +599,90 @@ export function pollForAccountChanges(currentAccountAddress: Address | null, int
...
+            totalEthereumBlocksCount
+            synced
+            failed
+          }
+        }
+      }
+    }`;
+
+    try {
+      const subgraphURL = settings.ganache.graphqlHttpProvider.split("/name/daostack")[0];
```
orenyodfat: ganache ?

src/arc.ts
```diff
@@ -586,6 +599,90 @@ export function pollForAccountChanges(currentAccountAddress: Address | null, int
   });
 }
 
+async function monitorGraphNodeSubgraph() {
+  if (targetedNetwork() === "ganache") {
```
orenyodfat: why condition that with "ganache"?

src/arc.ts
```diff
@@ -586,6 +599,90 @@ export function pollForAccountChanges(currentAccountAddress: Address | null, int
   });
 }
 
+async function monitorGraphNodeSubgraph() {
```
orenyodfat: what this function monitor?
can it be documented ?

src/arc.ts
```diff
@@ -586,6 +599,90 @@ export function pollForAccountChanges(currentAccountAddress: Address | null, int
...
+    }
+  }
+
+  const subgraphName = (settings as any)[process.env.NETWORK]["graphqlHttpProvider"].split("https://api.thegraph.com/subgraphs-daostack/name/daostack/")[1];
+  const query = `{
+    indexingStatusesForSubgraphName(subgraphName: "daostack/` + subgraphName + `") { subgraph synced failed chains { network ... on EthereumIndexingStatus { latestBlock { number hash } chainHeadBlock { number hash } } } }
+  }`;
+
+  try {
+    const { data } = (await axios.post("https://api.thegraph.com/index-node/graphql", { query })).data;
```
orenyodfat: index-node ?

src/arc.ts
orenyodfat: why it is under arc.ts file ? how about adding that under health.ts ?

src/arc.ts
tibetsprague: oh i was using it externally. now im not, though i could easily imagine it being used externally in the future. for now i can remove

src/arc.ts
tibetsprague: that is the name of the key for ganache settings yes

src/arc.ts
tibetsprague: because checking the subgraph status is done with different queries then checking on graphnode

src/arc.ts
tibetsprague: sure i can move to a different file

src/arc.ts
tibetsprague: hey i took this code from the alchemybots repo, and it works, so if there is a different way to monitor thegraph then let me know

tibetsprague:
@orenyodfat can you rereview?

src/arc.ts
orenyodfat: why it is needed here? in the production code ?

src/arc.ts
orenyodfat: why not making that general via the web3 provider ?
why there is unique functionality for ganache ?

src/arc.ts
dkent600: I suspect `getCurrentBlock` should be contained in subgraphHelpers, and not be exported.  Seems like this may be the only place it is used.

tibetsprague: of course changing the means I now have to export getWeb3, so not really any benefit here. to me putting getCurrentBlock in the same place as the other web3 functions makes more sense to me

src/arc.ts
tibetsprague: im not sure what you mean. this code checks if the ganache subgraph is running or not. that seems useful for devs to know

src/arc.ts
tibetsprague: how do I check the graph status via the web3 provider?

orenyodfat: my question was why there is a need to explicitly check for the network type ("ganache") .
the web3 provider already knows that

src/arc.ts
dkent600: makes more sense to me to export getWeb3.

dkent600: but I'm not attached to it

src/arc.ts
tibetsprague: because we do different queries when interacting with ganache or the graph nodde

tibetsprague:
@orenyodfat @dkent600 can i get a re-review?

tibetsprague:
Right now it is showing this warning if you are connected to the wrong network which is not correct behavior

orenyodfat:
should not show subgraph error msg if the user is in the wrong network..it is different case.

tibetsprague:
@orenyodfat it was actually a different bug but i think this should be fixed now!

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
+## 0.10.6
+  - Features Added
+    - Show warning if subgraph is not updating
+
 ## 0.10.5
   - Features Added
-    - change home page to show DAOs again instead of feed. Show separate "Your DAOs" section containing DAOs the user follows or is a member of, then the list of other DAOs below.
+    - Change home page to show DAOs again instead of feed. Show sepatate Your DAOs section of DAOs the user follows or is a member of, with the list of other DAOs below.
```
dkent600, line 7: this change shouldn't be here

tibetsprague: weird

tibetsprague:
@orenyodfat i need your approval here to merge

Issue #1558, opened 2020-04-02
support alchemy kovan 


orenyodfat:
https://github.com/daostack/alchemy/blob/dev/src/settings.ts#L126

tibetsprague:
@dkent600 I could also take this one

PR #1564, opened 2020-04-03
support for kovan

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1558

* doesn't support any of the genericscheme contracts


dkent600:
Blocked, waiting for upgrades to migration and client

src/components/Scheme/SchemeOpenBountyPage.tsx
```diff
@@ -41,9 +41,8 @@ export default class SchemeOpenBounty extends React.Component<IProps, IState> {
   private getApi = async (): Promise<void> => {
     // query all open bounties issued by current dao ordered by decending price
     // rinkeby api and mainnet api are different subdomains
-
     const network = (await getNetworkName()).toLowerCase();
-    const testnet = network === "rinkeby" ? "rinkeby." : "";
+    const testnet = network === "main" ? "" : `${network}.`;
```
dkent600, line 45: will this work for xDai?

orenyodfat:
@dkent600  https://alchemy-supportkovan-nuejtqnov.herokuapp.com/dao/0x7b32d2d56f32ce6523a4dee9008ccec2619a6122

src/arc.ts
```diff
@@ -1,5 +1,5 @@
 import { NotificationStatus } from "reducers/notifications";
-import { getNetworkId, getNetworkName } from "./lib/util";
+import { getNetworkId, getNetworkName, targetedNetwork } from "./lib/util";
```
orenyodfat, line 2: why this change ?

src/arc.ts
```diff
@@ -19,37 +19,6 @@ let selectedProvider: any;
 let web3Modal: Web3Modal;
 let initializedAccount: Address;
 
-export type Networks = "main"|"rinkeby"|"ganache"|"xdai"
```
orenyodfat, line 22: why this is removed ?

src/lib/util.ts
```diff
@@ -268,12 +299,12 @@ export async function getNetworkName(id?: string): Promise<string> {
     case "main":
     case "1":
       return "main";
-    case "morden":
-    case "2":
-      return "morden";
-    case "ropsten":
-    case "3":
-      return "ropsten";
+    // case "morden":
```
orenyodfat, line 302: where is "kovan" ?

src/settings.ts
```diff
@@ -66,6 +66,46 @@ function getWeb3ConnectProviderOptions(network: string) {
...
+      return {
+        network: "kovan",
+        torus: {
+          package: Torus,
+          options: { network: "kovan" },
+        },
+        walletconnect: {
+          package: isMobileBrowser() ? null : WalletConnectProvider,
+          options: {
+            infuraId: "e0cdf3bfda9b468fa908aa6ab03d5ba2",
```
orenyodfat, line 79: does this id is the same for all networks ?

src/arc.ts
dkent600: see the corresponding change and the comment where I explained, in lib/util.ts

src/lib/util.ts
dkent600: in the code, line 315

src/arc.ts
orenyodfat: is that a must for supporting kovan ?

src/settings.ts
dkent600: I am assuming so, am planning to test the assumption.  We should of course test all of the wallets.

src/arc.ts
dkent600: see the corresponding change and the comment where I explained in the code, line 136 in lib/util.ts

data/tokens.json
```diff
@@ -91,6 +91,37 @@
...
+      "0x9f13dbe2ef9733a4bb580412e9cb6f463e9bdf1a": {
+        "decimals": 18,
+        "name": "KyberNetwork",
+        "symbol": "KNC"
+      },
+      "0x9be1001d601102ae0f24ab4764dd5ce2f3e5b096": {
+        "decimals": 6,
+        "name": "USD Coin",
+        "symbol": "USDC"
+      }
```
orenyodfat, line 122: please exclude all these tokens from kovan . 
it is not exist or the addresses are different at kovan

src/arc.ts
dkent600: Without that change we would be worsening the problem by adding a dependency to the circle. 
 There is less risk in fixing this than in not (which is easy to say because this change is virtually risk-free).

src/components/Scheme/SchemeOpenBountyPage.tsx
orenyodfat: you are asking if we might have the bounty scheme in xdai ?
so ,might be though not now ..and if yes so any how the scheme json will need to be updated.
so I suggest to remove this update.

src/arc.ts
orenyodfat: is this change is a must to support kovan or any other network ?

src/components/Scheme/SchemeOpenBountyPage.tsx
dkent600: is bounty scheme in kovan?

src/settings.ts
orenyodfat: why this infura id is in the code at all ? it should come as env var . it is private info , no ?

src/lib/util.ts
orenyodfat: why there is rinkeby support in this function and not kovan ?

src/settings.ts
dkent600: I'm not sure all the wallets are going to work.

src/components/Scheme/SchemeOpenBountyPage.tsx
orenyodfat: not yet. and if there will be it will require updating the json for this scheme. 
so again , please remove this update

src/settings.ts
orenyodfat: why we have the infura token spread all over the code ?
it should get into env var . see #1569

src/components/Scheme/SchemeOpenBountyPage.tsx
orenyodfat: leaving this update will require us to verify that the rinkeby bounty scheme is working as before with this change . did you ?

src/lib/util.ts
dkent600: the support for kovan is there, please check the code

src/arc.ts
dkent600: same answer as above because this is part of the same change.

Without that change we would be worsening the problem by adding a dependency to the circle.  There is less risk in fixing this than in not (which is easy to say because this change is virtually risk-free).

It is not a must for this PR, unless you're worried about the potential for webpack to behave badly because of circular module dependencies.

src/settings.ts
dkent600: that is for another PR

data/tokens.json
dkent600: done

src/lib/util.ts
orenyodfat: at the getNetworkName ?

src/components/Scheme/SchemeOpenBountyPage.tsx
dkent600: The logic is so simple, and the way to test it not trivial (in fact not possible since we can't use rinkeby), that I haven't worried about it.

dkent600: Maybe @EzraWeller can test viewing bounties when he gets to this.

src/lib/util.ts
dkent600: Here: https://github.com/daostack/alchemy/blob/a84a2ae762e869cc3a107644f3e5bd35f9a16335/src/lib/util.ts#L314

Issue #1551, opened 2020-03-31
user image/avatar override its address at the history tab


orenyodfat:
<img width="1083" alt="Screen Shot 2020-03-31 at 18 08 34" src="https://user-images.githubusercontent.com/5994398/78043078-923f1200-737b-11ea-8dc4-de8797043a43.png">

PR #1565, opened 2020-04-03
Make sure proposer name doesn't overlap avatar on proposal history rows

tibetsprague:
Fixes: https://github.com/daostack/alchemy/issues/1551

This took much longer to figure out than it looks :)


orenyodfat:
why https://alchemy-1551-history-ro-rffoae.herokuapp.com/ does not SHOW_ALL_DAOS as other review app are ?  @tibetsprague

tibetsprague:
@orenyodfat i dont know, but why does that effect this review?

orenyodfat:
> @orenyodfat i dont know, but why does that effect this review?

because review app does not work as expected .even though it rebased on dev

tibetsprague:
@orenyodfat slug size is too big, thats why

PR #1578, opened 2020-04-08
correct display of Torus provider name

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1542


package.json
```diff
@@ -86,7 +86,7 @@
     "@fortawesome/react-fontawesome": "^0.1.3",
     "@portis/web3": "^2.0.0-beta.43",
     "@sentry/browser": "^5.0.8",
-    "@toruslabs/torus-embed": "^1.2.3",
+    "@toruslabs/torus-embed": "^1.2.6",
```
orenyodfat, line 89: should this be 1.3.0 ? (latest)

dkent600: that is coming in another PR.  No need here.

dkent600:
> duplicate in #1587
> this should be close?

Not a dup

package.json
dkent600: in any case, 1.30 is badly broken for us.  we have to wait on it.

Issue #1577, opened 2020-04-06
Cannot follow/unfollow DAOs anymore?


tibetsprague:
Eric reported here: https://daostack.zendesk.com/agent/tickets/140

Seems pretty high priority to me

dkent600:
I can't seem to log in to that zendesk link

PR #1580, opened 2020-04-08
Fix (un)following DAOs

tibetsprague:
Required updating 3box to latest version


tibetsprague:
note that upgrading 3box does increase the number of high level security vulnerabilities from 2 to 12 in our repo

orenyodfat:
is that fix #1577 ?

tibetsprague:
@orenyodfat yes

Issue #1585, opened 2020-04-10
Genesis Alpha is duplicated in DAO search


dkent600:
Searching for Genesis Alpha. Once you search for it it never goes away. And if you search twice it always shows up twice. Three times... four....

EzraWeller:
Still seeing this on rinkeby and mainnet. It seems to only effect the "Other DAOs" list and not "My DAOs"?

<img width="1009" alt="Duplications" src="https://user-images.githubusercontent.com/5414803/80136221-6a546000-856f-11ea-99e6-7f0dcf9976d6.png">
<img width="1023" alt="Duplications 2" src="https://user-images.githubusercontent.com/5414803/80136225-6aecf680-856f-11ea-9012-e165ff325bef.png">

orenyodfat:
maybe clean your browsing cache.

tibetsprague:
i do see it happen sometimes locally though too

PR #1586, opened 2020-04-11
Home page and feed changes

### LINKED ISSUE #1581: Member DAOs First on Alchemy Home Page
Eric, Eylon, and I view this as a high priority for April. We are seeking feedback from the implementation team on the following plan:
## Summary
We think the main user story right now and in the near future that is relevant for Alchemy’s home page is a user coming to Alchemy to visit a specific DAO. Right now, they must sift through the feed and the searchable DAO list to find this DAO. Instead, we should just show them links to DAOs they are a member of directly.
## Product Design
Move the feed to its own tab or remove it completely (I am agnostic about this and open to feedback, but happy to make a call if no one has a strong opinion).

Create a new DAO list showing any DAOs the logged in user holds reputation in. Label this “Your DAOs.” This list should go above the current full, searchable list of DAOs (“All DAOs”).

tibetsprague:
Separate section at the top of the DAOs page with DAOs user is a member of or follows first. Then below that is an Other DAOs section with all other DAOs.
Note that this will currently not show more than 100 DAOs that a user is a member of and 100 that they folllow.

Depends on new client daostack/client#443

Fixes: #1581


package.json
```diff
@@ -77,7 +77,7 @@
     "test:unit": "jest"
   },
   "dependencies": {
-    "3box": "1.16.1",
+    "3box": "1.17.1",
```
orenyodfat, line 80: is this related to this pr ?

tibetsprague: this is from the Dao follow/unfollow PR. it will go away when dev is merged in again

tibetsprague:
Also fixes https://github.com/daostack/alchemy/issues/1585

src/lib/daoHelpers.ts
```diff
@@ -0,0 +1,55 @@
+import { DAO, IDAOState, Reputation, Token } from "@daostack/client";
+import { getArc } from "arc";
+import BN = require("bn.js");
+
+// TODO: all this should really be in the client library, after new client is done i will move it there if its not already
```
orenyodfat, line 5: please open an issue in the client for that please.

orenyodfat:
decision : 
-move the search box on top of "Your DAOs" .
-leave only one search box

src/lib/daoHelpers.ts
tibetsprague: https://github.com/daostack/client/issues/447

src/components/Daos/DaosPage.tsx
```diff
@@ -154,69 +182,65 @@ class DaosPage extends React.Component<IProps, IState> {
...
     );
   }
 }
 
+const createSubscriptionObservable = (props: IStateProps, data: SubscriptionData = null) => {
+  const arc = getArc();
+  const { currentAccountAddress, followingDAOs } = props;
+
+  // TODO: right now we don't handle a user following or being a member of more than 100 DAOs
+  //       it was too hard to figure out the UI with infinite scrolling in this case we would need a different UI
```
dkent600, line 198: Is infinite scroll more feasible after the change that has only one search box?

tibetsprague: no

tibetsprague: cant really do two lists with infinite scroll on both on the same page easily... unless they are in separate columns or tabs

tibetsprague:
@EzraWeller i switched to one search box at the top of the page. it could definitely still use a bit of stylistic polish

arsena21:
I think we will need a link to the FEED... (see below for inspiration). We don't have a link anywhere else from the home page.

image

tibetsprague:
@arsena21 there is a link in the menu at the bottom left. what if we just moved that menu up to the top left so its more visible?

arsena21:
Yeah, I think that could work

orenyodfat:
does the "Look for dao to follow or join" msg should appears after finding something in the search box ?

Screen Shot 2020-04-21 at 11 45 36

arsena21:
Maybe we should not have the "Your DAOs" block (including the msg Oren posted above around "Look to join or follow below" unless there are any DAOs in that top block?

orenyodfat:
> Maybe we should just hide the "Your DAOs" block (including the msg Oren posted above around "Look to join or follow below" unless there are any DAOs in that top block?

Lets have a quick decisions re that , otherwise it will not get into next release.

tibetsprague:
ok I can hide if there is nothing
@EzraWeller are you down with that? And are you down with the idea to move the sidebar links that are normally at the bottom left to the top left so they are more visible?

tibetsprague:
ok I can hide if there is nothing
@EzraWeller are you down with that? And are you down with the idea to move the sidebar links that are normally at the bottom left to the top left so they are more visible?

tibetsprague:
ok i just went ahead and hid the Your DAOs section if it empty

tibetsprague:
but i didnt move the sidebar links yet

PR #1592, opened 2020-04-14
Import torus using require

tetratorus:
our build process was changed slightly

this import still works if you enable esModuleInterop: true in your tsconfig but there are other issues if you do that in the alchemy repo. 

Instead, if you import it via require, it just works.

Ran this via npm run start-prod and was able to see Torus in the Web3Modal modal, and also login successfully.


dkent600:
Thanks @tetratorus !

PR #1594, opened 2020-04-15
Biconomy Integration for Meta Transactions

tomarsachin2271:


tomarsachin2271:
Demo screencast can be found [here](https://biconomy-header.nyc3.digitaloceanspaces.com/Screen%20Recording%202020-04-15%20at%208.26.25%20PM.mov)

Changes involve Biconomy integration in alchemy to enable gasless operations.
Biconomy integration is a two step process:
1. Alchemy DApp needs to be registered on biconomy dashboard https:///dashboard.biconomy.io 
In this example i registered the DApp and uploaded GenesisProtocol, Redeemer and ContributionReward smart contracts and enabled meta transactions on Vote, Redeem and ProposeContributionReward method.

2. Integration Biconomy SDK(Mexa) on client code (This has already been done in this PR) DApp Id and ApiKey used to initialize biconomy here should be registered by Alchemy owners and those values should be used in this repository.

Since this approach uses Smart Contract Wallet approach for meta transactions, so whenever user connects his wallet say Metamask, a new smart contract wallet is created for the User.

So some UI changes should be done to make the UX better.
1. User should be notified when a new contract wallet is created and we should show the wallet address on UI somewhere.
2. Since vote transaction go through contract wallet and msg.sender is contract wallet address, reputation should be in contract wallet address.
3. On proposal details page, code should check if user contract wallet address has reputation or not and enable/disable the vote buttons on UI for the user.

Suggestion: There should be global mode NORMAL or GASLESS. In normal mode user client wallet address is shown to user. 
In GASLESS mode, user contract wallet is shown to user and checked for any reputation.

PR #1597, opened 2020-04-15
Improve package footprint

dkent600:
- lowers our bundle size in package.json
- forces "production" in app.js


package.json
```diff
@@ -144,8 +143,7 @@
     "utility-types": "^3.7.0",
     "web3": "1.2.4",
     "web3modal": "^1.3.0",
-    "webpack": "^4.36.1",
-    "ws": "^7.1.0"
```
orenyodfat, line 148: please bring back ws

src/components/Daos/DaosPage.tsx
```diff
@@ -98,7 +98,8 @@ class DaosPage extends React.Component<IProps, IState> {
     // Always show Genesis Alpha first
     let finalDAOList = allDAOs.filter((d: DAO) => d.staticState.name === "Genesis Alpha" && d.staticState.name.toLowerCase().includes(search));
 
-    if (process.env.NODE_ENV === "staging") {
+    // eslint-disable-next-line no-extra-boolean-cast
+    if (!!process.env.SHOW_ALL_DAOS) {
```
orenyodfat, line 102: if (process.env.SHOW_ALL_DAOS === "false")

package.json
dkent600: done

dkent600: removed it again

package.json
```diff
@@ -223,6 +219,7 @@
     "typings-for-css-modules-loader": "^1.7.0",
     "uglifyjs-webpack-plugin": "^2.2.0",
     "webdriverio": "^5.11.6",
+    "webpack": "^4.36.1",
```
orenyodfat, line 222: how about using latest 4.43.0. 
moving this do devDependency might do the slug diff staff 👍

dkent600: it is already in devDependency.  That is what this PR does.  I'll look into upgrading it.

dkent600: We are already getting 4.43.0 by default:

> $ npm list webpack
alchemy-client@0.10.4 [...]alchemy
+-- @storybook/react@5.3.18
| +-- @storybook/core@5.3.18
| | +-- corejs-upgrade-webpack-plugin@2.2.0
| | | `-- webpack@4.43.0  deduped
| | `-- webpack@4.43.0  deduped
| `-- webpack@4.43.0  deduped
`-- webpack@4.43.0

orenyodfat: yes because it is with ^ ..
please set that explicitly : `"webpack": "4.43.0"`

dkent600: I am skeptical that moving this into devDependencies really did the trick.  It wasn't working last time I tried.  May justr be the usual unpredictable Heroku behavior.

The difference this time is that I rebased on the latest dev and rebuilt package-lock.json.  Now I see something which I _think_ has changed here:

> alchemy-client@0.10.4 C:\Users\dkent\Documents\Projects\GitHub\DAOStack\alchemy
`-- 3box@1.17.1
  `-- ipfs@0.40.0 
    `-- UNMET DEPENDENCY ipfs@0.38.1
>
> npm ERR! missing: ipfs@0.38.1, required by ipfs@0.40.0
npm ERR! extraneous: ipfs@0.38.1 C:\Users\dkent\Documents\Projects\GitHub\DAOStack\alchemy\node_modules\ipfs\node_modules\ipfs

I don't remember it saying "UNMET DEPENDENCY ipfs@0.38.1" before.  ipfs@0.38.1 in fact does appear to be missing now under ipfs/node_modules.  I further confirm the size of IPFS in the slug has vastly diminished: 

image

dkent600:
> I Gott say I suspect at some point we will want a staging env back but ok...

@tibetsprague We may _want_ to, but will have to keep in mind that setting `NODE_ENV` to "staging" was breaking both the heroku build and behavior in underscore-protocol, each of which assumes the only possible values are "production" and "development".  In other words, they always treat "staging" (or any other value != "production")  like "development".

Issue #1589, opened 2020-04-13
use subgraph v39_3 


orenyodfat:

dkent600:
@orenyodfat What does this new version bring to Alchemy?

orenyodfat:
stability of subgraph .

PR #1600, opened 2020-04-16
upgrade to subgraph v39_3

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1589


src/components/Daos/DaosPage.tsx
```diff
@@ -98,7 +98,8 @@ class DaosPage extends React.Component<IProps, IState> {
     // Always show Genesis Alpha first
     let finalDAOList = allDAOs.filter((d: DAO) => d.staticState.name === "Genesis Alpha" && d.staticState.name.toLowerCase().includes(search));
 
-    if (process.env.NODE_ENV === "staging") {
+     // eslint-disable-next-line no-extra-boolean-cast
+    if (process.env.SHOW_ALL_DAOS === "true") {
```
dkent600, line 102: This should be removed.  It doesn't work and is done correctly in another PR

orenyodfat: it is doing what it suppose to do .

dkent600: but it will always return false.

dkent600: plus everything else I pointed out

orenyodfat: NO . it is return TRUE!
and it is working

PR #1575, opened 2020-04-05
Add OmenDisputeResolution Generic Scheme

nicoelzer:
This commit adds a GenericScheme to integrate the DxDAO with realit.io as arbitrator to resolve disputes. More information can be found:

- https://daotalk.org/t/omen-mvp-overview/1229 (Section Arbitration)
- https://realitio.github.io/docs/html/arbitrators.html

This PR provides the GenericScheme for Kovan & Rinkeby for testing purposes. Mainnet will be added later.

PR #1601, opened 2020-04-16
Added dxDAO Token Registry Generic Schemes

nicoelzer:
Added Generic Schemes to be used by the dxDAO:

- OmenTokenRegistry
- MixTokenRegistry
- ExchangeTokenRegistry


orenyodfat:
is that override #1575 ?

nicoelzer:
> is that override #1575 ?

Just closed #1575 as this also contained the DisputeResolution GenericScheme which is currently lower priority then this one here. Will create a separate PR for DisputeResolution once we got this live.

orenyodfat:
travis failed

Issue #1567, opened 2020-04-04
DAO front-ends


arsena21:
See designs here: https://www.figma.com/file/fq10yjxy4EaaESPzqBGkEG/DAO-front-end

EzraWeller:
# Simple DAO Front-ends
Eric, Eylon, and I view this as a high priority for April. We are seeking feedback from the implementation team on the following plan:
## Summary
The home page for each DAO in Alchemy should more clearly highlight the few simple actions users should take in that DAO. We would like to accomplish this by making each DAO’s homepage a full page with content controlled by the DAO.

This should happen in two stages:
1. Give each DAO a default home page, with default home page content created by us.
2. Give each DAO the ability to change this content via proposal (or the option to add that ability).
## Business Case / Target Audience
This feature should mainly benefit:
- DAOstack and other service providers building on the stack, since it will make active DAOs clearer to understand for anyone new looking at them.
- DAOs looking to onboard new users, since this will let them include the most important information for members clearly and up front.
## User Stories
A new user hits the home page of a DAO in Alchemy. What they see makes it obvious what this DAO is about and how they can take their first actions in the DAO -- whether that means proposing, commenting, voting, or some other action.
## Product Design
[An initial mock up created by Eric](https://www.figma.com/file/fq10yjxy4EaaESPzqBGkEG/DAO-front-end?node-id=0%3A1)

The details here are not final, but this mock does give a fairly clear visual direction.
## Functional Spec
We have the progress made by dOrg and Ape Unit on the DAO wiki and header respectively. We should evaluate if these projects can and should be repurposed for this patch.
### Stage 1: 
Every DAO should be given a simple homepage that lives in the “Home” tab and includes this information:

> [DAO Name]

> Welcome to [DAO name], a decentralized organization built on DAOstack.

> Visit the [Proposals page](link to proposals page) to make a proposal to the DAO or vote on existing proposals.

> Visit the [DAO Wall](link to DAO Wall) to participate in general discussion about this DAO.

The current “Home” tab should be relabeled to “Proposals.” 
### Stage 2:
Create a scheme/plugin called “Edit Homepage” that lets the DAO change the content on its Homepage. 
- This does not need to reuse the DAO wiki or DAO header work, but it can, if they will help (up to the implementation team).
- The main text of the homepage should support markdown or some other rich text that can embed images and videos if at all possible.
[...]

PR #1602, opened 2020-04-17
DAO Landing Page

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1567  (Phase I)

Mock is here: https://www.figma.com/file/qsog4uCensMYaaSclxlmVi/DAO-Front-End-Mocks?node-id=0%3A1

Also:

* cleans up the DAO sidebar menu item text a little

**Note** one bit isn't finished:  Still haven't seen the scrolling behavior with lots of comments in disqus.


EzraWeller:
I'm not sure about the disqus convos not working -- I don't see any errors on that I've seen.

Everything looks good, except there should still be a "Home" link in each DAO, and what is currently labeled "Proposals" should be "Plugins" (at least for now, because that page's header is "All Plugins"). This is for the "Menu" on the left hand side. Does that make sense?

dkent600:
So the decision is to rename the header on the DaoSchemesPage from "All Plugins" to "Proposal Plugins"

EzraWeller:
The sidebar doesn't seem quite right:
Screen Shot 2020-04-29 at 9 53 25 AM

We still want the "Home" tab to be there, I think (which should go with the house icon), but now it should lead to the new DAO homepage. The "Proposals" tab should have a different icon, maybe a scroll or document if we have that. (So four side bar tabs: Home, Proposals, Members, History)

EzraWeller:
Approved!

PR #1603, opened 2020-04-18
add subgraph_endpoints.json + hide infuraId

### LINKED ISSUE #1590: Make `setting.ts` a JSON/ Separate subgraph versions into a JSON
Currently, the Alchemy monitoring bot takes the subgraph version from the `setting.ts ` file. The problem is that since it is a `ts` file it's problematic for the bot and this functioning breaks with every change to that file. The file could (at least partially) be turned to a JSON file, which will allow the bot to work well with it.

orenyodfat:
fix #1590

### DIFF
--- a/src/settings.ts
+++ b/src/settings.ts
@@ -8,6 +8,7 @@ const Torus = require("@toruslabs/torus-embed");
 
 const Portis = require("@portis/web3");
 const Fortmatic = require("fortmatic");
+const SubgraphEndpoints = require("./subgraph_endpoints.json");
 
 function isMobileBrowser(): boolean {
   // if (!window) {
@@ -36,7 +37,7 @@ function getWeb3ConnectProviderOptions(network: string) {
         walletconnect: {
           package: isMobileBrowser() ? null : WalletConnectProvider,
           options: {
-            infuraId: "e0cdf3bfda9b468fa908aa6ab03d5ba2",
+            infuraId: process.env.INFURA_ID,
           },
         },
         burnerconnect: {
@@ -76,7 +77,7 @@ function getWeb3ConnectProviderOptions(network: string) {
         walletconnect: {
           package: isMobileBrowser() ? null : WalletConnectProvider,
           options: {
-            infuraId: "e0cdf3bfda9b468fa908aa6ab03d5ba2",
+            infuraId: process.env.INFURA_ID,
           },
         },
         burnerconnect: {
@@ -130,7 +131,7 @@ function getWeb3ConnectProviderOptions(network: string) {
         walletconnect: {
           package: isMobileBrowser() ? null : WalletConnectProvider,
           options: {
-            infuraId: "e0cdf3bfda9b468fa908aa6ab03d5ba2",
+            infuraId: process.env.INFURA_ID,
           },
         },
         burnerconnect: {
@@ -165,8 +166,8 @@ function getWeb3ConnectProviderOptions(network: string) {
 
 export const settings = {
   ganache: {
-    graphqlHttpProvider: "http://127.0.0.1:8000/subgraphs/name/daostack",
-    graphqlWsProvider: "ws://127.0.0.1:8001/subgraphs/name/daostack",
+    graphqlHttpProvider: SubgraphEndpoints.ganache_http,
+    graphqlWsProvider: SubgraphEndpoints.ganache_ws,
     graphqlSubscribeToQueries: false,
     web3Provider: "ws://127.0.0.1:8545",
     web3ProviderRead: "ws://127.0.0.1:8545",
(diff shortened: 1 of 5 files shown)


src/subgraph_endpoints.json
```diff
@@ -0,0 +1,12 @@
+{
```
dkent600, line 1: 1) these names should be in the form: http[networkname] and  and ws[NetworkName]

2) "mainnet" should be "main"

so like:

http_main, ws_main
rinkeby_main, ws_,main

etc

orenyodfat: why does it matter ?

dkent600: 1) this naming pattern is standard best practice in coding.

2) main is that we're using almost everywhere in Alchemy, including in settings.ts

src/settings.ts
```diff
@@ -205,15 +206,14 @@ export const settings = {
     web3ConnectProviderOptions: getWeb3ConnectProviderOptions("xdai"),
   },
   main: {
-    graphqlHttpProvider: process.env.ARC_GRAPHQLHTTPPROVIDER || "https://api.thegraph.com/subgraphs-daostack/name/daostack/v39_3",
-    graphqlWsProvider: process.env.ARC_GRAPHQLWSPROVIDER || "wss://api.thegraph.com/subgraphs-daostack/name/daostack/v39_3",
+    graphqlHttpProvider: process.env.ARC_GRAPHQLHTTPPROVIDER || SubgraphEndpoints.mainnet_http,
+    graphqlWsProvider: process.env.ARC_GRAPHQLWSPROVIDER || SubgraphEndpoints.mainnet_wsssss,
```
dkent600: oooooops

src/settings.ts
```diff
@@ -175,28 +176,28 @@ export const settings = {
...
   },
   rinkeby: {
-    graphqlHttpProvider: process.env.ARC_GRAPHQLHTTPPROVIDER || "https://api.thegraph.com/subgraphs-daostack/name/daostack/v39_3_rinkeby",
-    graphqlWsProvider:  process.env.ARC_GRAPHQLWSPROVIDER || "wss://api.thegraph.com/subgraphs-daostack/name/daostack/v39_3_rinkeby",
+    graphqlHttpProvider: process.env.ARC_GRAPHQLHTTPPROVIDER || SubgraphEndpoints.http_rinkeby,
+    graphqlWsProvider:  process.env.ARC_GRAPHQLWSPROVIDER || SubgraphEndpoints.ws_rinkeby,
     graphqlSubscribeToQueries: false,
-    web3Provider:  process.env.ARC_WEB3PROVIDER || "wss://rinkeby.infura.io/ws/v3/e0cdf3bfda9b468fa908aa6ab03d5ba2",
-    web3ProviderRead:  process.env.ARC_WEB3PROVIDERREAD || "wss://rinkeby.infura.io/ws/v3/e0cdf3bfda9b468fa908aa6ab03d5ba2",
+    web3Provider:  process.env.ARC_WEB3PROVIDER || "wss://rinkeby.infura.io/ws/v3/"+ process.env.INFURA_ID,
```
dkent600, line 182: please use best practice form:

`wss://rinkeby.infura.io/ws/v3/${process.env.INFURA_ID}`

src/settings.ts
```diff
@@ -165,8 +166,8 @@ function getWeb3ConnectProviderOptions(network: string) {
 
 export const settings = {
   ganache: {
-    graphqlHttpProvider: "http://127.0.0.1:8000/subgraphs/name/daostack",
-    graphqlWsProvider: "ws://127.0.0.1:8001/subgraphs/name/daostack",
+    graphqlHttpProvider: SubgraphEndpoints.http_ganache,
+    graphqlWsProvider: SubgraphEndpoints.ws_ganache,
     graphqlSubscribeToQueries: false,
     web3Provider: "ws://127.0.0.1:8545",
```
dkent600, line 172: no change required here?

orenyodfat: no

webpack.prod.config.js
```diff
@@ -108,6 +108,7 @@ plugins: [
       ARC_IPFSPROVIDER_PORT : "",
       ARC_IPFSPROVIDER_PROTOCOL : "",
       ARC_IPFSPROVIDER_API_PATH : "",
+      INFURA_ID : "",
```
tibetsprague, line 111: I think you need to add this to app.json too for review apps to work

orenyodfat: though review app is working without it ..

dkent600: > I think you need to add this to app.json too for review apps to work

@tibetsprague That is exactly why I thought yesterday that SHOW_ALL_DAOS would not work in @orenyodfat's PR, because he hadn't added SHOW_ALL_DAOS here.  But then I see that he has in fact added it.  So I wonder when he did, before or after he proved that the functionality was working on review apps.

One thing I know:  You have to have these set here for running locally, where heroku doesn't exist to insert values.

PR #1631, opened 2020-04-23
Fix duplicate DAOs when searching using capital characters

tibetsprague:
Fixes https://github.com/daostack/alchemy/issues/1585

### DIFF
--- a/src/components/Daos/DaosPage.tsx
+++ b/src/components/Daos/DaosPage.tsx
@@ -67,11 +67,13 @@ class DaosPage extends React.Component<IProps, IState> {
     // If search string greater than 2 search on server for any other DAOs not yet loaded that match this search
     if (searchString.length > 2) {
       const arc = getArc();
+      const firstChar = searchString.charAt(0);
       const foundDaos = await combineLatest(
         // eslint-disable-next-line @typescript-eslint/camelcase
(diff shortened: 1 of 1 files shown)


src/components/Daos/DaosPage.tsx
```diff
@@ -67,11 +67,13 @@ class DaosPage extends React.Component<IProps, IState> {
     // If search string greater than 2 search on server for any other DAOs not yet loaded that match this search
     if (searchString.length > 2) {
       const arc = getArc();
+      const firstChar = searchString.charAt(0);
```
dkent600, line 70: Is it not feasible to make the entire DAO name search be case insensitive?

tibetsprague: I dont think so

tibetsprague: I mean it is if the DAOs are already loaded but for ones we look up on the server I didn't find how to do that with the graphnode

dkent600: OK I approved, but this looks kinda like a hack.  Maybe could add a code comment explaining the situation?

tibetsprague: i did below

orenyodfat: why not lower case the search string and and dao names and then do the search ?

Issue #1623, opened 2020-04-23
Cosmetic issues on landing page


EzraWeller:
#1617 

A few cosmetic issues with the latest landing page:

<img width="1326" alt="Home page" src="https://user-images.githubusercontent.com/5414803/80134478-d08bb380-856c-11ea-8954-c42dbd3a8802.png">

tibetsprague:
@EzraWeller I made the text big on purpose so its obvious, why do you think it should be smaller?
The Create a DAO button lining up with edge of DAO list is slightly tricky... I will try

PR #1633, opened 2020-04-24
Tweak cosmetics of the DAOs page

tibetsprague:
Line up create button with right of the daos list.
Make search box font size smaller
Make the search box sticky
Improve how we handle stickiness
Get rid of empty space
Definitely check mobile!

Fixes https://github.com/daostack/alchemy/issues/1623

### DIFF
--- a/src/components/Daos/Daos.scss
+++ b/src/components/Daos/Daos.scss
@@ -6,6 +6,7 @@
 
 .headerWrapper {
   position: relative;
+  top: 45px;
 }
 
 .headerTitle {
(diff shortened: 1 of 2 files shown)


src/components/Daos/Daos.scss
```diff
@@ -19,29 +20,31 @@
...
     padding: 2px;
   }
 }
 
 .createDaoButton {
-  position: fixed;
-  right: 20px;
-  top: 122px;
+  position: sticky;
+  float: right;
```
dkent600, line 46: We can pretty much always do things better without using float.  Dunno what you're attempting here, but can it not be done with css flex styles?

src/components/Daos/DaosPage.tsx
```diff
@@ -48,17 +49,31 @@ class DaosPage extends React.Component<IProps, IState> {
     super(props);
 
     this.state = {
+      isMobile: window.innerWidth <= 550,
```
dkent600, line 52: wow this is starting to look like a big hack, definitely not a "tweak".  Can this not be done with css and @media rules?

src/components/Daos/Daos.scss
```diff
@@ -19,29 +20,31 @@
...
     width: 425px;
     padding: 2px;
   }
 }
 
 .createDaoButton {
-  position: fixed;
-  right: 20px;
-  top: 122px;
+  position: sticky;
```
dkent600, line 45: Sticky has sketchy browser support:  https://caniuse.com/#feat=css-sticky

src/components/Daos/Daos.scss
tibetsprague: maybe it could. but this works

src/components/Daos/DaosPage.tsx
tibetsprague: not this part no because its in the react code

src/components/Daos/Daos.scss
tibetsprague: hmm. well, is it good enough @EzraWeller ?

Issue #1593, opened 2020-04-14
daos members with zero reputation at the moment of a proposal creation should get a proper error msg


orenyodfat:
- Currently the voting button is enable for such members and they are getting wired error msg
![image](https://user-images.githubusercontent.com/5994398/79276575-f5b75c80-7eb0-11ea-8dad-ecebc95d45c5.png)

- This case can be checked by compering the proposal creation date and the dao member creation date.

tibetsprague:
shouldn't voting just be disabled for them? then we dont need error messages

orenyodfat:
yes, the voting button should be disable in this case... though the user should understand why :) .
- . ["Requires reputation in this DAO"](https://github.com/daostack/alchemy/blob/dev/src/components/Proposal/Voting/VoteButtons.tsx#L115) is not a good one for this specific case.
- for this case the user should get a msg with something like : "Requires reputation at the time of proposal creation" ..or something in this direction...

tibetsprague:
right

tibetsprague:
not sure why this was closed, alchemy has to be updated to use the new arc.js version here: https://github.com/daostack/alchemy/pull/1634

dkent600:
It was automatically closed because the arc.js ticket had this in the description:  "Needed to fix daostack/alchemy#1593"

PR #1634, opened 2020-04-25
Can't vote if didn't have reputation in DAO when proposal was created

### LINKED ISSUE #1655: view all 8 > ??
Screen Shot 2020-05-04 at 17 59 17

tibetsprague:
Fixes https://github.com/daostack/alchemy/issues/1593
Depends on https://github.com/daostack/arc.js/pull/451

fixes #1655

### DIFF
--- a/src/components/Proposal/Voting/VoteButtons.tsx
+++ b/src/components/Proposal/Voting/VoteButtons.tsx
@@ -117,13 +117,13 @@ class VoteButtons extends React.Component<IProps, IState> {
            * The following condition deduces that the user could not have had rep when the
            * proposal was created because of the following behavior in the subgraph:
            *
-           * 1) `currentAccountState` (`ReputationHolder` in the subgraph) is an entity that associates
+           * 1) `currentAccountState` (`ReputationHolder` in the subgraph) represents an entity that associates
            * a single DAO with an ethereum account
            * 2) currentAccountState can only exist in the subgraph when the account has > 0 rep in the DAO
            * 3) `currentAccountState.createdAt` is set only once:  When currentAccountState is being instantiated
-           * for the first time, in response to a Mint event
+           * for the first time, in response to a Mint event that brings the rep > 0
            * 4) when a Burn event brings the rep <= 0, then the entity is removed from the subgraph
-           * 5) when `currentAccount` is not found in the subgraph, then a fake currentAccountState is created with
+           * 5) when `currentAccount` is not found in the subgraph, then a fake `currentAccountState` is created with
            * rep == 0
            */
           (currentAccountState && (proposal.createdAt < currentAccountState.createdAt)) ?
--- a/src/components/Proposal/Voting/VoteButtons.tsx
+++ b/src/components/Proposal/Voting/VoteButtons.tsx
@@ -1,4 +1,4 @@
-import { Address, IDAOState, IMemberState, IProposalOutcome, IProposalStage, IProposalState } from "@daostack/client";
+import { Address, IDAOState, IMemberState, IProposalOutcome, IProposalStage, IProposalState } from "@daostack/arc.js";
 import { voteOnProposal } from "actions/arcActions";
 import { enableWalletProvider } from "arc";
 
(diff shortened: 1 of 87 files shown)


src/components/Proposal/Voting/VoteButtons.tsx
```diff
@@ -113,13 +113,15 @@ class VoteButtons extends React.Component<IProps, IState> {
...
         (currentAccountState && currentAccountState.reputation.eq(new BN(0))) ?
           "Requires reputation in this DAO" :
-          proposal.stage === IProposalStage.ExpiredInQueue ||
-              (proposal.stage === IProposalStage.Boosted && expired) ||
-              (proposal.stage === IProposalStage.QuietEndingPeriod && expired) ||
-              (proposal.stage === IProposalStage.Queued && expired) ?
-            "Can't vote on expired proposals" :
-            proposal.stage === IProposalStage.Executed ?
-              `Can't vote on ${proposal.winningOutcome === IProposalOutcome.Pass ? "passed" : "failed"} proposals` : "";
+          (currentAccountState && (proposal.createdAt < currentAccountState.createdAt)) ?
```
dkent600, line 116: This check doesn't guarantee that the account had reputation on proposal.createdAt.

Neither does `(currentAccountState && currentAccountState.reputation.eq(new BN(0)))` which seems like completely incorrect logic that needs to be corrected (it is checking the current balance but should be checking the balance at proposal creation time).

Am I missing something?

orenyodfat: - if this is true `(currentAccountState && currentAccountState.reputation.eq(new BN(0)))` so `"Requires reputation in this DAO"` should be the error msg. 
-otherwise if this is true `(currentAccountState && (proposal.createdAt < currentAccountState.createdAt)) ` so 
`"Must have had reputation in this DAO when the proposal was created"` should be the error msg.
whats wrong with that ?

dkent600: If `currentAccountState` can only ever be defined when the current account currently has rep in the DAO, then why are we checking `currentAccountState.reputation.eq(new BN(0)))`?  That case should never be able to be true.

dkent600: BTW, we should also be modifying the logic on line 98:

`const votingDisabled = [....]`

Also line 62:

`if (currentAccountState.reputation.gt(new BN(0)))`

orenyodfat: because if the current account is not a member of the dao a default currentAccountState value is returned from arc.js . [see here](https://github.com/daostack/arc.js/blob/51ce56ad8691dab37f0dfc7a13a9967cdfe7ac5b/src/member.ts#L184)

dkent600: Looking at the subgraph, it appears that `createdAt` represents the date of the last block in which the account's rep was minted or burned (pretty misleading if you ask me).

(see [here](https://github.com/daostack/subgraph/blob/c34fdbbdebea46a95a42c84609ec2b0900d425ed/src/mappings/Reputation/mapping.ts#L84)).

So `proposal.createdAt < currentAccountState.createdAt` is true when the proposal was created prior to the last time the amount of the account's reputation changed.  This tells us nothing about what the balance was at the time the proposal was created.

orenyodfat: yes, but it is set only at the first time see https://github.com/daostack/subgraph/blob/c34fdbbdebea46a95a42c84609ec2b0900d425ed/src/mappings/Reputation/mapping.ts#L42

dkent600: There remains an issue:   Suppose:

1) DAO and Reputation contract are created on date 1
2) User account is minted rep on date 2
3) User rep is burned (rage quit, penalty by vote of the DAO, or really poor voting patterns) to zero on date 3
4) Proposal is created on date 4
5) User rep is minted again to > zero on date 5
6) User tries to vote on date 6

The line in question will return false even though the user had zero rep on the date the proposal was created.

orenyodfat: if user rep is burned the rep member entity is deleted and the user is kiked out of the dao.
when it will be minted rep again the createDate will be upon this event.

orenyodfat:
Screen Shot 2020-05-06 at 0 00 52

PR #1637, opened 2020-04-27
Dev Docs For Environment Variables

dkent600:


docs/development.md
```diff
@@ -77,3 +77,10 @@ npm run build:watch
     docker rm $(docker ps -a -q)
     docker rmi $(docker images -a -q)
     ```
+# Environment Variables
+
+Heroku uses app.json to supply default values to the Heroku app environment variable configuration GUI.
```
orenyodfat, line 82: where does it mentioned other place that we are using Heroku for the builds ?

why does the normal dev care at all that we are using Heroku ?

dkent600: any dev (normal or not) who wants to add a new environment variable, and I know of two external PRs already who do, need to know how to do it correctly.

PR #1641, opened 2020-04-28
Update travis.yml

orenyodfat:
add travis_wait to solve build longer than 10 minutes
see https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received


orenyodfat:
not sure it is a must. will close that for now.

PR #1625, opened 2020-04-23
Biconomy Integration for Meta Transactions

orenyodfat:

PR #1644, opened 2020-04-29
Don't show weird confirmation popups when using the "Pending" notification type

tibetsprague:
Fixes a weird "bug" in the Biconomy PR


tibetsprague:
There's no way to test this right now except by merging into https://github.com/daostack/alchemy/pull/1625/ and seeing if it stops showing the alert message saying "Often transactions get approved after 24h...."

PR #1649, opened 2020-04-30
handle errors in getLatestBlock

dkent600:


dkent600:
@tibetsprague 

>just to be clear this doesn't solve the issue of the subgraph not syncing error because if we return null then we will display that message. which is probably a correct message to show?

When we return null the entire iteration is aborted.  We do nothing.  (This is commented in the code)

src/arc.ts
```diff
@@ -126,7 +134,7 @@ export async function initializeArc(provider?: any): Promise<boolean> {
     const retryLink = new RetryLink({
       attempts: {
         max: 5,
-        retryIf: (error, _operation) =>  {
+        retryIf: (error, _operation) => {
```
orenyodfat, line 137: why this change ? how the linter did not catch that before..is it catching that at all?
if linter does not care about so leave these update for other "lint": pr.. it is difficult to review with all these changes.

src/lib/subgraphHelpers.ts
```diff
@@ -81,7 +81,10 @@ export function pollSubgraphUpdating(): Observable<boolean> {
           // This only works if user is connected to web3, that's why we need the try
           let atCurrentBlock = true;
           try {
-            const curBlock = (await getCurrentBlock()).number;
+            const curBlock = (await getCurrentBlock())?.number;
+            if (curBlock === undefined) {
+              return; // try again next time
```
orenyodfat, line 86: maybe log something to the console..here?

dkent600: done

src/arc.ts
dkent600: 1) On this github page you can tell the github to ignore whitespage differences
2) I already created a separate PR to make the linter catch these:  https://github.com/daostack/alchemy/pull/1650

PR #1662, opened 2020-05-05
Update to use subgraph v39_4 

orenyodfat:
due to a [bug]( https://github.com/daostack/subgraph/pull/538) in subgraph

- isRegistered flag is false if scheme is unregistered and then re registered. -- cause alchemy to not display the scheme


orenyodfat:
> @orenyodfat Can you please put in the description how this bug would affect Alchemy? The subgraph ticket similarly says nothing.

see description

orenyodfat:
thanks @dkent600

Issue #1674, opened 2020-05-06
"New proposal..." buttons should line up vertically with page


EzraWeller:
The "new proposal" buttons should share the margin of the rest of the page (currently looking at release 0.10.6):
<img width="1146" alt="Screen Shot 2020-05-06 at 3 44 40 PM" src="https://user-images.githubusercontent.com/5414803/81221417-cd52e780-8fb0-11ea-811e-d628ec77e806.png">

dkent600:
Broken by: https://github.com/daostack/alchemy/pull/1598

See css.createproposal, it is "fixed" in place (!)

dkent600:
Note that the requested appearance does not conform to what is currently in production:

![image](https://user-images.githubusercontent.com/1821666/81330461-1a48c380-906e-11ea-8a5c-3bc80eb11dd4.png)

dkent600:
To be consistent, making the same layout change and refactor with the Edit Plugin button

![image](https://user-images.githubusercontent.com/1821666/81334766-8a5a4800-9074-11ea-887f-179767991c4b.png)

dkent600:
In terms of code, but not layout, this one should be refactored as well, but will not do it now because I'm afraid it would prevent the PR from being accepted.  But what is nice is that now these buttons will be more consistent in appearance across these pages.

![image](https://user-images.githubusercontent.com/1821666/81335031-f76ddd80-9074-11ea-8d9f-571502f438a2.png)

PR #1681, opened 2020-05-07
refactor and fix the layout of New Proposal button

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1674

- This moves the code for the New Proposal and Edit Plugin buttons into the components where they belong, instead of hacking them from the wrong components.
- The appearance of the two buttons is now consistent with the appearance of a similar button:  Edit Plugin on the Proposal Plugins page
- I removed the name of the plugin from the "New Proposal" button because the button is harder to layout when it becomes very wide, and the name is redundant.
- responsive behavior is better than it was, I think, but still not great

There are more notes in the ticket this fixes.

Potentially affected parts, to test:

- the SchemeContainer page for any scheme, (url: /dao/:daoAvatarAddress/scheme/:schemeId) for any scheme, not neglecting special cases of ReputationFromToken, Competition and Bounty
  - Proposals tab (add Proposal button)
    - with no proposals
    - with proposals
  - the Information tab (Edit Plugin button)
    - it is possible that ReputationFromToken, Competition and Bounty don't have an Edit Plugin button.  If that is an issue, then please create a separate ticket.

### DIFF
--- a/src/components/Scheme/Scheme.scss
+++ b/src/components/Scheme/Scheme.scss
@@ -8,7 +8,7 @@
   color: #689ad6;
   padding: 0;
   margin: 0;
-  padding-bottom: 15px;
+  margin-bottom: 30px;
   background-color: #e5ebf2;
   padding-top: 10px;
 }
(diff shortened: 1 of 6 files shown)


orenyodfat:
why moving a button require so much code ?
I would like to avoid refactoring the code on this phase. too risky .

PR #1690, opened 2020-05-09
Delete "ChainID 100" optional parameter which alchemy does not support

Theylon:
I get an error when trying to submit a proposal with 100 as the chainid, this is an optional parameter.

image


orenyodfat:
please submit an issue , not a pr . and describe what you are seeing and how to reproduce.
the chain id might be needed.

orenyodfat:
This error come from MetaMask ..? 
Screen Shot 2020-05-09 at 15 42 01

orenyodfat:
seems like a torus issue https://github.com/torusresearch/torus-website/pull/1077

Issue #1684, opened 2020-05-08
proposal thinks I had no rep when I did


dkent600:
This is on ganache.  I created a proposal and I had plenty rep.  But:

![image](https://user-images.githubusercontent.com/1821666/81405760-4a43a580-9106-11ea-809e-39334873ccdb.png)

![image](https://user-images.githubusercontent.com/1821666/81405794-5a5b8500-9106-11ea-952a-3493c799077b.png)

orenyodfat:
which ganache dao is it ?

dkent600:
You can see the DAO in the screen shot above:  Pervasive Hot air balloon

But it happens for me on any DAO

dkent600:
@orenyodfat What is the status of this?

PR #1691, opened 2020-05-09
temorary workaround for gh issue #1684

orenyodfat:
till subgraph (or migration) issue will be solved https://github.com/daostack/subgraph/issues/548


orenyodfat:
> Approving, but please add an appropriate reference to #1684 here

it already done by the commit name.

PR #1704, opened 2020-05-14
add simple message popup

dkent600:
Replaces any need for `alert`.  Motivated by the Biconomy PR.

DaoLandingPage is now using it, thus simplifying the code in that component.

DAOCreator should be made to use it: 
 https://github.com/dOrgTech/DAOcreator/issues/502#issuecomment-634805595


dkent600:
Regarding linting, there aren't many changes and the only changes that I see involve whitespace, which you can eliminate yourself from the diff using the github GUI:

image

PR #1651, opened 2020-05-01
Switch to 3box comments

tibetsprague:
Fixes https://github.com/daostack/alchemy/issues/1553

- removes App.spec.ts because it was [failing due to this error](https://github.com/ipfs-shipyard/ipfs-companion/issues/475).  It contained no tests anyway.  And merely attempting to import a react component in the absence of a browser seems like a weird thing to do.

PR #1705, opened 2020-05-14
upgrade 3box

dkent600:
Attempts to fix heroku slug size issues


dkent600:
The build still fails with this.  We will continue to track this with this PR:  https://github.com/daostack/alchemy/pull/1651

Issue #1667, opened 2020-05-06
Allow "Recipient address" in proposal form to be empty (and be populated with proposer address)


Theylon:
The need is for onboarding web2 users using Tor.us / Fortmatic.  They have never been exposed to an Ethereum address and wouldn't know where to find their own + the complexity of pasting it into the DAO. 

User Story of web 2 user:
- User clicks on a DAO onboarding proposal URL which sends him to the DAO with a pre-filled template form.
- It asks him to login with Tor.us / Fortmatic (email)
- User fills in (replaces) the relevant details in the proposal form
- User submits in order to onboard to the DAO.

EzraWeller:
Doug suggests having a button for filling in your own address in the recipient box, because auto-filling it or allowing it to be left empty could easily lead to people accidentally submitting a recipient they didn't intend. Since the user already has to log in, that seems okay to me. Thoughts, @Theylon ?

dkent600:
Need resolution on the spec for this, thanks.  @EzraWeller

Theylon:
I think this adds an unnecessary layer of complexity in an already complex system. Some of these people won't what an Ethereum address is, let alone what it means to copy it. 
 
In the edge case mentioned this accidental recipient could only be themselves, I'm estimating 90%+ of Alchemy proposals have the recipient as the proposer so this is highly unlikely.

Alchemy is too complex (and practically unusable to newcomers) precisely because of maximizing optionality (10% of uses) rather than XU (90% of uses)
I think we need to engineer some things to be more UX friendly. If you believe this is a critical issue for all of Alchemy I suggest either using the xDAI deployment for this (there are no other DAOs using xDAI) or creating a whole new deployment.

EzraWeller:
To be clear, the user would not need to know their own address -- they would just need to click a button that says "Make me the recipient!" or something.

Theylon:
"Recipient of what? this is my introduction to join the cooperative"

dkent600:
I am reluctant to add a feature that increases the risk of causing significantly bad consequences (bad proposals and everything that follows, like loss of funds and reputation) to even a small number of users.  Yes, this is tyranny of the minority, but the consequences are significant enough IMHO to justify accommodating them.

dkent600:
Alternative:  automatically populate with one's own account address (don't even show the textbox input), and replace with a button entitled "Beneficiary is not you?" that when clicked would display the textbox field, empty.  There could be a '?' button that explains.

[...]

PR #1708, opened 2020-05-15
default user account when creating CR proposal

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1667


dkent600:
@orenyodfat 

>  think that by default the user account address should be fill out with gray

Done

Issue #1656, opened 2020-05-04
Extend Scheme URL Params => Proposal Templates to Competition proposal & competition submission 


Theylon:
Extend this PR to Competition proposal and competition submission.

This will allow an extra layer of abstraction to Alchemy, sending DAO members directly to the submission screen with a template and filling instructions.

https://github.com/daostack/alchemy/pull/1226

PR #1712, opened 2020-05-15
url params for Competitions/Submissions

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1656

### DIFF
--- a/src/lib/proposalUtils.ts
+++ b/src/lib/proposalUtils.ts
@@ -1,4 +1,5 @@
 import { copyToClipboard } from "./util";
+import moment = require("moment-timezone");
 
 const cloneDeep = require("clone-deep");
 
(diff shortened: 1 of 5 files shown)


src/lib/proposalUtils.ts
```diff
@@ -1,4 +1,5 @@
 import { copyToClipboard } from "./util";
+import moment = require("moment-timezone");
```
orenyodfat, line 2: how this is related to this pr ?

src/lib/proposalUtils.ts
```diff
@@ -23,7 +24,11 @@ export function importUrlValues<Values>(defaultValues: Values) {
           initialFormValues[prop] = BigInt(paramValue);
           break;
         case "object":
-          initialFormValues[prop] = JSON.parse(paramValue);
+          if (moment.isMoment(defaultValues[prop])) {
+            initialFormValues[prop] = moment(paramValue);
+          } else {
+            initialFormValues[prop] = JSON.parse(paramValue);
```
orenyodfat, line 30: how this is related to this pr ?

src/lib/proposalUtils.ts
dkent600: Competition parameters include dates.  This is to support them.

src/lib/proposalUtils.ts
dkent600: Competition parameters include dates.  This is to support them.

PR #1717, opened 2020-05-17
Fix plugin manage edit "plugin manager" scheme . permission

### LINKED ISSUE #1716: Plugin manager -Edit - wrong permission set for edit plugin manager .

orenyodfat:
Fix "plugin manager" Edit scheme  permission ,for the case editing the "PluginManager" it self 

fix #1716

### DIFF
--- a/src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
+++ b/src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
@@ -158,7 +158,6 @@ class CreateSchemeRegistrarProposal extends React.Component<IProps, IState> {
     };
     setSubmitting(false);
     await this.props.createProposal(proposalValues);
-
     Analytics.track("Submit Proposal", {
       "DAO Address": this.props.daoAvatarAddress,
       "Proposal Title": values.title,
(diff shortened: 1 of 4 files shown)


src/lib/schemeUtils.ts
```diff
@@ -148,6 +148,24 @@ export function schemeNameAndAddress(address: string) {
   }
 }
 
+/**
+ * given the address (of a scheme), return scheme's name
+ * @param  address [description]
+ * @return         [description]
+ */
+export function schemeNameFromAddress(address: string) {
```
dkent600, line 156: Can you please refactor `schemeNameAndAddress` (just above) to use this new function?

src/components/Proposal/Create/SchemeForms/CreateSchemeRegistrarProposal.tsx
```diff
@@ -138,12 +138,18 @@ class CreateSchemeRegistrarProposal extends React.Component<IProps, IState> {
     } else {
       proposalType = IProposalType.SchemeRegistrarEdit;
     }
+
+    let permissionString = "0x" + permissions.toString(16).padStart(8, "0");
```
dkent600, line 142: Not a big deal, but I prefer an if/else or switch/case/default pattern here, to save an unnecessary assignment to `permissionString` when schemeName === "Plugin Manager"

dkent600: Can you please add a comment in the code explaining why the special case hack is required?  Explaining why `permissions` is good for every scheme except "Plugin Manager".  This should not be the case, seems like a bug elsewhere.  Need to explain in the code.

Issue #1732, opened 2020-05-19
Show "value" propose along with any generic scheme proposal.


orenyodfat:
From @orenyodfat :

> when proposing to generic scheme you can also add value param to the proposal. this value eventualy send ether from the avatar along with the generic call
> we need to show this value .,.and maybe alert if this value is not 0, i mean color that in red

dkent600:
@orenyodfat Need more precision about what you want here

dkent600:
**Notes**

According to the code, when creating the proposals, for what it is worth:

1. with known schemes, `value` cannot be set by the user
2. with unknown schemes `value` **can** be set by the user

When displaying the proposal details, which I think is the more relevant thing here:

3. arc.js is not supplying the value of `value`.  It does appear, however, to be available from the subgraph.  Created this ticker on arc.js:  https://github.com/daostack/arc.js/issues/476

dkent600:
**Notes**

According to the code, when creating the proposals, for what it is worth:

1. with known schemes, `value` cannot be set by the user
2. with unknown schemes `value` **can** be set by the user

When displaying the proposal details, which I think is the more relevant thing here:

3. arc.js is not supplying the value of `value`.  It does appear, however, to be available from the subgraph.  Created this ticker on arc.js:  https://github.com/daostack/arc.js/issues/476

orenyodfat:
https://github.com/daostack/arc.js/releases/tag/0.2.69

orenyodfat:
here is a gs scheme with value = 1 . https://alchemy.daostack.io/dao/0x519b70055af55a007110b4ff99b0ea33071c720a/proposal/0x4d28e845a25f9b335aed4942067adb9ddffdb29364da04982233f538610c226a

PR #1733, opened 2020-05-20
warn when GenericScheme `value` is > 0

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1732

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1685,9 +1685,9 @@
       }
     },
     "@daostack/arc.js": {
-      "version": "0.2.68",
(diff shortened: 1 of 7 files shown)


EzraWeller:
So, I think the UI does not need to blink red unless it is above a certain proportion of the treasury, and I also think we should strive to use the same system for all proposal types. Here is a suggestion:
- make sure all plugins are displaying any assets being sent in a proposal in a regular display format (not blinking red)
- If the amount of those assets is above a threshold percent (e.g. >5% of the DAO's holdings of that asset, >2% of total reputation), have it blink red

This way, it should be unified for all plugin types. How about that?

PR #1755, opened 2020-05-26
page the preboosted proposals

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1748

- upgrades react-infinite-scroll-component because react kept issuing messages that we should, and to be sure we're getting the best behavior, as this was a bit of a hack.  The upgrade touches all pages that do infinite scroll:  DAO History, DAO Members, DAO list, Feed, and of course, Scheme Proposals

- moves regular and preboosted rendering into separate components so that they can fetch their pages independently.

- there is no good way to test this without creating more than 100 proposals and staking on them.  I tested on a local build running against ganache and set the page sizes to 2 instead of 100.


orenyodfat:
>  in ganache you can easily boost/preboost 100 proposals.

orenyodfat:
use this [multipropose script](https://gist.github.com/orenyodfat/123656e5be9d79f3a4546dec8c5fc349)  to multi propose and preboost

orenyodfat:
1. `pending boosted proposals (101)` is not updated unless there is a page refresh ?  is that how it suppose to be ?
2.  this console error popped up after around 100 pre boosted proposals.
  
Screen Shot 2020-06-08 at 11 06 01

dkent600:
@orenyodfat I reported the console error you are seeing a while ago:  https://github.com/daostack/alchemy/issues/1756.  In my view, this is pretty high priority to fix.  I don't know if it is associated with this PR code change.

I will look into the problem you are seeing with the preboosted section not updating.

orenyodfat:
> @orenyodfat I reported the console error you are seeing a while ago: #1756. In my view, this is pretty high priority to fix. I don't know if it is associated with this PR code change.

I can reproduce that only on this pr.

dkent600:
> I can reproduce that only on this pr.

Thanks, I was wondering...

dkent600:
> pending boosted proposals (101) is not updated unless there is a page refresh ? is that how it suppose to be ?

No it is not supposed to be that way.

I just staked up proposals on both ganache and the rinkeby review app they instantly and correctly appeared in the pre-boosted queue.  Is this your repro case?

dkent600:
@orenyodfat 

> pending boosted proposals (101) is not updated unless there is a page refresh ? is that how it suppose to be ?

OK, now I understand that this is actually about the proposal _count_ not changing as proposals appear and disappear in the queue.  I find that this is happening in the regular queue as well, and it is a pre-existing problem (happens on the regular staging app).  Question is: do you want me to work this in this PR, or as a separate PR?

orenyodfat:
> @orenyodfat
> 
> > pending boosted proposals (101) is not updated unless there is a page refresh ? is that how it suppose to be ?
> 
> OK, now I understand that this is actually about the proposal _count_ not changing as proposals appear and disappear in the queue. I find that this is happening in the regular queue as well, and it is a pre-existing problem (happens in staging). Question is: do you want me to work this in this PR, or as a separate PR?

if it is happened on the regular q ..so lets do that on other pr.

orenyodfat:
so the thing left for this pr is the memory leak issue

Issue #1789, opened 2020-06-01
improve splitting of camel case unknown plugin names


dkent600:
In the code here:  https://github.com/daostack/alchemy/blob/1f1da9d7d7b9784a91037722642e95a4bc9a8b93/src/lib/schemeUtils.ts#L120

1. convert first letter to uppercase
2. handle adjacent uppercase letters better.

Example of what can work better:

```
      /**
       * Add spaces before capital letters to approximate a human-readable title.
       * Note if the name already contains spaces, they will be left alone.
       * If there are adjacent uppercase characters, they will not be split, which
       * sometimes will be correct (like "ID") and sometimes not (like "AScheme").
       * (The previous version of this, `/([A-Z])/g, ' $1'`, would split adjacent uppercase characters,
       * which when wrong would be more wrong than not splitting (like "I D").)
       **/
      name = `${name[0].toUpperCase()}${name.slice(1).replace(/([a-z])([A-Z])/g, "$1 $2")}`;
```

Issue #1766, opened 2020-05-28
GenericSchemeRegistry : support different plugin names(aliases) for different target contracts using the same ui components. 


orenyodfat:
The situation currently that a DAO might have multiple genericSchemes using the same UI , which might confuse the user. 
see for example ENSPublicResolver at https://alchemy.daostack.io/dao/0x519b70055af55a007110b4ff99b0ea33071c720a/schemes/ 

To make this a generic solution , there are 2 options : 
1. map an optional "alias" for each address. This `alias` if exist will be used as the Plugin name in the UI.
something like adding "aliases" entity to the generic scheme registry json.
for example : 
`
 {
  "name": "EnsPublicResolver",
  "addresses": {
    "main": [
      "0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41",
      "0x1da022710df5002339274aadee8d58218e9d6ab5",
      "0x226159d592e2b063810a10ebf6dcbada94ed68b8"
    ],
 },`
`"aliases":{
  {"0x4976fb03c32e5b8cfe2b6ccb31c09ba78ebaba41","NewENSPublicResolver"},
  {"0x226159d592e2b063810a10ebf6dcbada94ed68b8","OldENSPublicResolver"}
},
...
`
2. Fetch the alias name of the plugin from the subgraph in real time. this is indexed at the contractInfo . https://github.com/daostack/subgraph/blob/master/daos/mainnet/dxdao.json#L40  

I feel that option 2 is preferable as this is the original intention of the "alias" in the contractInfo subgraph, it is already there.. and it will make alchemy more generic. 
@dkent600 , thoughts ?

dkent600:
Yeah, option 2 makes sense except for the fact that I still don't understand why the subgraph is maintaining a list of registered contracts per DAO in the first place.

dkent600:
Awaiting arc.js PR:  https://github.com/daostack/arc.js/pull/492

PR #1790, opened 2020-06-01
use contract alias for plugin name

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1766
Resolves: https://github.com/daostack/alchemy/issues/1789

Resolving the 1789 was necessary because splitting camel casing wasn't working right.

But need to confirm that splitting camel-casing is what we want.  We've always been doing it.  It will now be more apparent.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -82,13 +82,37 @@
       }
     },
     "3id-blockchain-utils": {
-      "version": "0.3.3",
-      "resolved": "https://registry.npmjs.org/3id-blockchain-utils/-/3id-blockchain-utils-0.3.3.tgz",
-      "integrity": "sha512-BSufokzlyxM/35lNtSKupZid/v9YodIEYCk2W1TJwQ6SbqiOzZk+05Q0Jh2VKVYDhns/5S+oIcud6G2cEOGxlw==",
+      "version": "0.3.4",
+      "resolved": "https://registry.npmjs.org/3id-blockchain-utils/-/3id-blockchain-utils-0.3.4.tgz",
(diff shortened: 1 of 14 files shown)


orenyodfat:
travis is red @dkent600

orenyodfat:
whats the logic behind parsing the scheme alias?  : 
for example , how [GenericSchemeDXTokenRegistry](https://github.com/daostack/subgraph/blob/master/daos/mainnet/dxdao.json#L34)  show up as `Generic Scheme DXToken Registry`

dkent600:
> whats the logic behind parsing the scheme alias? :

Mostly the same as it always has been.  But herein I created a function for it, and documented it.

See the changes here:  
https://github.com/daostack/alchemy/pull/1790/files#diff-57d57172ecb5dd8b766a8fa1431a4dd4

orenyodfat:
> > whats the logic behind parsing the scheme alias? :
> 
> The same as it always has been. But herein I created a function for it, and documented it.
> 
> See the changes here:
> https://github.com/daostack/alchemy/pull/1790/files#diff-57d57172ecb5dd8b766a8fa1431a4dd4

not sure that this logic is good. 
Users/daos  should have the freedom to give what ever title they wish. in that case they should pr that to subgraph, currently ,  later on there will be other way to do that.

dkent600:
> not sure that this logic is good.

OK, just note in the code for `schemeName` that we already hardcode the renaming of several Arc contracts.  Let me know how you'd like it to work.

orenyodfat:
> > not sure that this logic is good.
> 
> OK, just note in the code for `schemeName` that we already hardcode the renaming of several Arc contracts, when there is no meaningful alias. Let me know how you'd like it to work.

when ever `alias` is exist for a scheme set the name of the plugin as the alias .
otherwise use what is done so far.

Issue #1741, opened 2020-05-21
Creation of DAOs via the DAOcreator on the xDAI network 


Theylon:
Currently we are able to deploy DAOs using the DAOcreator on standard main and test networks. 

As gas prices skyrocket resulting in voting transactions to cost $5+ to be mined within an hour, many DAOs are opting for xDAI deployment rather than mainnet. 

Currently xDAI dao deployment is not available through the DAOcreator. According to @dOrgJelli some support is required to add this.

dkent600:
@Theylon 

> some support is required to add this

In Alchemy?  So DaoCreator itself is go-to-go as for as this is concerned?

@dOrgJelli What support is needed?

Theylon:
I think the goal is to use DAOcreator for most needs. 

Gas fees are skyrocketing at the moment and most new DAOs want to initialize on xDAI.

dtbx:
I join @Theylon in requesting support. It's an urgent need, currently multiple DAOpreneurs cannot get their DAO experiments progressed because the Ethereum network is clogged.  Launching a DAO using DAOcreator would usually be easy on main and test networks. 

DAOs will potentially opt to xDAI deployment rather than mainnet to save expenses on gas.

@orishim @dOrgJelli

dkent600:
I have no idea what I meant by "go-to-go" above. Doh.  Bottom line I think it that this work needs to be done in the @orgTech repo.  There is already a ticket on that:  https://github.com/dOrgTech/DAOcreator/issues/387

PR #1791, opened 2020-06-02
Daocreator xdai support

cbrzn:
Resolves: https://github.com/daostack/alchemy/issues/1741


cbrzn:
- Contracts used for creation are now from version 0.0.1-rc.41
- UGenericScheme is no longer supported
- Migration script updated to support xdai

@dkent600

dkent600:
@orenyodfat Can you please review that we're OK with taking the changes that @cbrzn listed in this comment:  https://github.com/daostack/alchemy/pull/1791#issuecomment-637681591?

orenyodfat:
Heroku build fail @cbrzn

cbrzn:
@orenyodfat not sure how can I fix that... maybe sharing the heroku logs might help to know what the problem is?

orenyodfat:
it looks like an out of memory issue 
`"FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory" `
does the new DAOCreator might  include any dramatic change re memory usage ?

dkent600:
Change `devtool: 'source-map' to `devtool: 'cheap-source-map' in webpack.prod.config.js.  This will fix the problem.  This change will soon likely be in dev in any case.

cbrzn:
> does the new DAOCreator might include any dramatic change re memory usage ?

Negative sir

> Change devtool: 'source-map' to devtool: 'cheap-source-map' in webpack.prod.config.js. This will fix the problem. This change will soon likely be in dev in any case.

So this is a problem with alchemy and not dao creator, right? Doing it now, thanks Doug!

dkent600:
Weird, the travis build succeeded, but here was it did not complete

Issue #1794, opened 2020-06-03
Visual bug for GEN in staking modal II


dkent600:
See: https://github.com/daostack/alchemy/issues/1629#issuecomment-637967166

and: https://github.com/daostack/alchemy/issues/1629#issuecomment-638312225

dkent600:
@orenyodfat Can you please go to this page in your Brave browser and post the results here:  https://whichbrowser.net/

dkent600:
I note that this is the second time in the last month or so we've seen a bug in Brave on @orenyodfat's machine that was not reproducible elsewhere.  On Brave on my machine:

![image](https://user-images.githubusercontent.com/1821666/83756857-a3d3bd00-a63d-11ea-9915-ab78cdc0244b.png)

dkent600:
Ah, this  error message isn't actually a validation error (which I did test), but a whole separate code path that I never tested:  @EzraWeller 

<img width="1154" alt="Screen Shot 2020-06-03 at 12 19 36 PM" src="https://user-images.githubusercontent.com/5414803/83663181-826bc600-a596-11ea-9ea4-6f0232b98e01.png">

dkent600:
Strangely, I don't see the error message in red, I see it in black:

![image](https://user-images.githubusercontent.com/1821666/83760222-a2f15a00-a642-11ea-905f-fcd8ecf6bb07.png)

dkent600:
@EzraWeller In you insufficient GEN error message bug, I am unable to reproduce the state of the "Buy GEN" text being wrapped underneath the input field.  Can you please give steps to repro?  Thanks.

dkent600:
@EzraWeller Actually, even in production I can't reproduce the error *at all*:

![image](https://user-images.githubusercontent.com/1821666/83763760-11d0b200-a647-11ea-8722-c4b53e927451.png)

Issue #1629, opened 2020-04-23
Visual bug for GEN in staking modal


EzraWeller:
on #1617

GEN display box slightly broken (mainnet, 0.10.5 release):

<img width="1074" alt="Staking" src="https://user-images.githubusercontent.com/5414803/80147587-b6a89b80-8581-11ea-8219-34fb9498d39a.png">

orenyodfat:
this issue is not fixed 

<img width="559" alt="Screen Shot 2020-06-03 at 8 39 56" src="https://user-images.githubusercontent.com/5994398/83599944-1e390a00-a576-11ea-955d-f826ba62f993.png">

dkent600:
@orenyodfat Can you please provide a URL where this can be reproduced?  Works fine for me.

Also:  What browser are you using?

EzraWeller:
Sounds like Oren is using Brave.

I'm seeing this in Chrome, but it's not the original issue:
<img width="1154" alt="Screen Shot 2020-06-03 at 12 19 36 PM" src="https://user-images.githubusercontent.com/5414803/83663181-826bc600-a596-11ea-9ea4-6f0232b98e01.png">

orenyodfat:
I am using brave

dkent600:
created a new ticket to represent the new behaviors and work flow to address them.

dkent600:
@EzraWeller In you insufficient GEN error message bug, I am unable to reproduce the state of the "But GEN" text being wrapped underneath the input field.  Can you please give steps to repro?  Thanks.

PR #1801, opened 2020-06-04
fix insufficient GEN error message layout

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1794

Only resolves the insufficient GEN error message layout, making sure that, whether or not we can reproduce it now, it can't possibly happen again, and cleaning up the code in the process.

The GEN label error (https://github.com/daostack/alchemy/issues/1629#issuecomment-637967166) is not reproducible.

### DIFF
--- a/src/components/Shared/PreTransactionModal.scss
+++ b/src/components/Shared/PreTransactionModal.scss
@@ -725,28 +725,13 @@ input[type="number"] {
       }
 
       .genError {
-        position: absolute;
-        width: 300px;
-        text-align: center;
-        color: rgba(80, 97, 118, 1);
-        left: 41px;
-        top: 81px;
+        color: $accent-2;
+        font-size: 14px;
+        font-family: $body-font;
+        font-weight: normal;
 
-        h4 {
-          color: $accent-2;
-          font-size: 14px;
-          font-family: $body-font;
-          text-align: left;
-          font-weight: normal;
-          margin: 0;
-          padding-top: 8px;
-        }
-
-        span {
-          display: block;
-          font-size: 10px;
-          font-weight: normal;
-          color: rgba(80, 97, 118, 1);
+        &.hidden {
+          display: none;
         }
       }
     }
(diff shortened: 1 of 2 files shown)


dkent600:
Waiting to receive a full repro case for the insufficient GEN error from @EzraWeller

dkent600:
> Waiting to receive a full repro case for the insufficient GEN error from @EzraWeller . See: [#1794 (comment)](https://github.com/daostack/alchemy/issues/1794#issuecomment-638849543)

Ezra declared it no longer reproducible

orenyodfat:
> > Waiting to receive a full repro case for the insufficient GEN error from @EzraWeller . See: [#1794 (comment)](https://github.com/daostack/alchemy/issues/1794#issuecomment-638849543)
> 
> Ezra is not finding how he reproduced it. But this code change will nevertheless ensure it can't happen again.

@dkent600 please explain why this pr introduce better code than we have now .why the issue can pop up again without this pr ? and how this pr insure that it will not.

dkent600:
> > > Waiting to receive a full repro case for the insufficient GEN error from @EzraWeller . See: [#1794 (comment)](https://github.com/daostack/alchemy/issues/1794#issuecomment-638849543)
> > 
> > 
> > Ezra is not finding how he reproduced it. But this code change will nevertheless ensure it can't happen again.
> 
> @dkent600 please explain why this pr introduce better code than we have now .why the issue can pop up again without this pr ? and how this pr insure that it will not ?

@orenyodfat 

* First note that there is more code deleted than added.  That typically means less complexity, easier and less risky to maintain.  We only have code that is absolutely needed (I often find myself deleting CSS code in Alchemy that is simply not needed and thus wastes developer time).  It also means I'm using CSS that is designed to handle the given situation, rather than trying to force the wrong code to work in situations it wasn't meant for.  Efficient code.
* The old code was using absolute positioning, which was absolutely unnecessary.  Absolute positioning is for us an anti-pattern (according to me) but was in common use in the past.  It is a pattern often used by CSS programmers when they just don't know CSS very well.  I am slowly getting rid of it where possible.  The problems with it?
   1. it is inflexible when it comes to varying browser sizes on desktop or when trying to be responsive to mobile platforms
   1. it is fragile to layout changes that occur to the elements around it
   1. it is harder to understand the code and more likely to cause mistakes when you have a situation like in this case, where element A supposed to be laid out below element B but in the code is actually defined _before_ element B.

PR #1816, opened 2020-06-11
Add dxswap governance scheme

nicoelzer:
- Add a generic scheme that let's the DXdao govern the dxswap protocol


nicoelzer:
@orenyodfat could we get this deployed on Rinkeby Alchemy?

PR #1818, opened 2020-06-11
Release 0.10.10

dkent600:


dkent600:
@EzraWeller I'm not sure we should release if we can't see anything on the Redemptions dropdown menu.  I can't reproduce the issue, but I think we need to understand how you were able to.

EzraWeller:
Sorry, there must be a misunderstanding, the redemptions page is mostly working fine! The only place I'm seeing any error is that one situation I submitted. On mainnet, for example, I see the normal behavior: 
Screen Shot 2020-06-12 at 7 36 56 AM

orenyodfat:
@dkent600 why `DutchX` scheme name changed to `Dutch X` ?

Screen Shot 2020-06-15 at 1 00 01

src/lib/schemeUtils.ts
```diff
@@ -98,7 +105,7 @@ export function schemeName(scheme: ISchemeState|IContractInfo, fallback?: string
       }
       const genericSchemeInfo = genericSchemeRegistry.getSchemeInfo(contractToCall);
       if (genericSchemeInfo) {
-        name = genericSchemeInfo.specs.name;
+        name = splitCamelCase(genericSchemeInfo.specs.name);
```
orenyodfat, line 108: 1. calling splitCamelCase on all genericScheme with no alias is not good.  for example : no need to call splitCamelCase on `DutchX` .
2. why using splitCamelCase at all here  ?

dkent600: Yeah, sorry, thought I reverted all of it.  Created a PR to fix, here: https://github.com/daostack/alchemy/pull/1822

PR #1821, opened 2020-06-12
Add the GEMS token to the new proposal token list, for i4

arsena21:
This PR is needed for a proposal within i4 Art DAO, the sooner we can get this the better :) 

Thanks!

cc @EzraWeller

### DIFF
--- a/data/tokens.json
+++ b/data/tokens.json
@@ -62,6 +62,11 @@
         "decimals": 18,
         "name": "Centrality Token",
         "symbol": "CENNZ"
+      },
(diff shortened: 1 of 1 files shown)


orenyodfat:
@arsena21  I think the pr title is confusing .
Does your intention is to add GEMS token in the list of the DAO holding tokens ?

EzraWeller:
I think the intention is for GEMS to appear in proposal forms, but @arsena21, perhaps you can specify exactly what you're looking for here.

arsena21:
Sorry, yes, it is for GEMS to show up in the proposal (within the token list dropdown)

PR #1842, opened 2020-06-16
Fix UI issue in staking modal

roienatan:
**Note**
- I fixed the issue only by changing styling.
- Tested on Chrome, Firefox and Brave browsers.
- If this issue will pop up again I'll make a more robust fix with the template itself and not only with CSS.
- I also checked the error message itself and not only the "GEN" text and it's fine, although I saw in previous discussions that others had problems with this.

### DIFF
--- a/src/components/Shared/PreTransactionModal.scss
+++ b/src/components/Shared/PreTransactionModal.scss
@@ -4,7 +4,7 @@
 
 .genLabel.genSymbol {
   top: 3px;
-  left: 143px;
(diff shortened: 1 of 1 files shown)


dkent600:
Is there a screenshot to show what it looked like before you fixed it?  Like what did you fix?

roienatan:
@dkent600 
Capture

dkent600:
This is already fixed in the dev branch.

PR #1843, opened 2020-06-16
BUIDLHub proposal notifications integration changes

mdcoon:
Proposed changes for users to register with BUIDLHub for DAO proposal notifications


arsena21:
video of demo: https://www.youtube.com/watch?v=5I3Czdd0gp8&feature=youtu.be

mdcoon:
> is there a reason that the dao Follow button lead to user transaction .(It ask for wallet)?
> https://alchemy-dev-mradczgvxpkcjdiuxh.herokuapp.com/daos/
> @mdcoon

It's not something I changed. But I would imagine it's because 3Box needs you to sign some information being pushed to IPFS/OrbitDB.

orenyodfat:
> > is there a reason that the dao Follow button lead to user transaction .(It ask for wallet)?
> > https://alchemy-dev-mradczgvxpkcjdiuxh.herokuapp.com/daos/
> > @mdcoon
> 
> It's not something I changed. But I would imagine it's because 3Box needs you to sign some information being pushed to IPFS/OrbitDB.

yes , though to were gone the 3box popup ?
Screen Shot 2020-06-17 at 17 02 43

orenyodfat:
> > > is there a reason that the dao Follow button lead to user transaction .(It ask for wallet)?
> > > https://alchemy-dev-mradczgvxpkcjdiuxh.herokuapp.com/daos/
> > > @mdcoon
> > 
> > 
> > It's not something I changed. But I would imagine it's because 3Box needs you to sign some information being pushed to IPFS/OrbitDB.
> 
> yes , though to were gone the 3box popup ?
> Screen Shot 2020-06-17 at 17 02 43

ok.. ignore . it looks like an xdai issue which is not related to this pr.

orenyodfat:
should there be an "envelope" here ? I do not see that .
here is the review app https://alchemy-dev-mradczgvxpkcjdiuxh.herokuapp.com/daos/ 
@mdcoon 
Screen Shot 2020-06-17 at 17 22 59

mdcoon:
> should there be an "envelope" here ? I do not see that .
> here is the review app https://alchemy-dev-mradczgvxpkcjdiuxh.herokuapp.com/daos/
> @mdcoon
> Screen Shot 2020-06-17 at 17 22 59

Yes. I did a push this morning to the BUIDLHub fork that was missing assets. I wonder if it's not pulled into the PR.

src/components/Daos/Daos.scss
```diff
@@ -66,13 +66,26 @@
 
       h2 {
         color: rgba(104, 155, 214, 1);
-        display: block;
+        display: inline;
```
dkent600, line 69: this change will affect other stuff.  Why is it needed?

src/components/Shared/SimpleMessagePopup.tsx
```diff
@@ -15,6 +15,7 @@ const mapDispatchToProps = {
 
 export enum EnumButtonSpec {
   Ok = 1,
+  HIDE = 2
```
dkent600, line 18: Can you please replace all the changes in this file with the following:

```
import * as uiActions from "actions/uiActions";
import ModalPopup from "./ModalPopup";
import * as React from "react";
import * as css from "./SimpleMessagePopup.scss";
import { connect } from "react-redux";
import { IRootState } from "reducers";

interface IDispatchProps {
  hideSimpleMessage: typeof uiActions.hideSimpleMessage;
}

const mapDispatchToProps = {
  hideSimpleMessage: uiActions.hideSimpleMessage,
};

export enum EnumButtonSpec {
  Ok = 1,
  // eventually will turn this into a confirmation modal, adding OkCancel, Yes/No and stuff
}

export interface ISimpleMessagePopupProps {
  closeHandler?: (event: any) => void;
  body: string | JSX.Element;
  buttonSpec?: EnumButtonSpec;
  /**
   * `buttonSpec` is ignored if `hideFooter` is true
   */
  hideFooter?: boolean;
  title?: string | JSX.Element;
}

interface IStateProps {
  options: ISimpleMessagePopupProps;
  showing: boolean;
}

const mapStateToProps = (state: IRootState): IStateProps => {
  return {
    options: state.ui.simpleMessageOptions,
    showing: state.ui.simpleMessageOpen,
  };
};

class SimpleMessagePopup extends React.Component<IDispatchProps & IStateProps, null> {

  private closeHandler = (event: any) => {
    this.props.hideSimpleMessage();
    if (this.props.options.closeHandler) {
      this.props.options.closeHandler(event);
    }
  }

  private renderButtons = (): JSX.Element => {
    const okButton = (<button className={css.closeButton} onClick={this.closeHandler}>OK</button>);
    switch (this.props.options.buttonSpec) {
      case EnumButtonSpec.Ok:
      default:
        return okButton;
    }
  }

    if (!this.props.showing) {
      return "";
    }

    if (!this.props.options.body) {
      throw new Error("message body is required");
    }

    return (
      <div className={css.modalContainer}>
        <ModalPopup
          closeHandler={this.closeHandler}
          header={
            <div className={css.modalHeader}>
              <div className={css.title}>{this.props.options.title ?? "Alchemy"}</div>
              <div className={css.closeButtonX} onClick={this.closeHandler}><img src={"/assets/images/Icon/close-grey.svg"} />
              </div>
            </div>
          }
          body={
            <div className={css.modalBody}>{this.props.options.body}</div>
          }
          footer={
            this.props.options.hideFooter ?
              <div className={css.modalFooter}>{this.renderButtons()}</div>
              : undefined
          }
        />
      </div>
    );
  }
}

export default connect(mapStateToProps, mapDispatchToProps)(SimpleMessagePopup);
```

src/components/Daos/Daos.scss
dkent600: would be safer to use `inline-block` I think.

mdcoon: In order to get the envelope to show up next to the H2 title of "Your DAOs", the H2 has to be inline. Note that the H2 is nested in this case so it only effects the H2 element of that class. If you have other suggestions on how to get it to layout properly, I'm all ears.

src/components/Shared/SimpleMessagePopup.tsx
mdcoon: This will basically remove my changes. The "OK" button being displayed in the modal is something I brought up early on and was told to go ahead and make the change. I didn't want both an "OK" button and "Send" button displayed in the modal. The user can click away from the modal or click the 'X' close in the popup to get rid of it.

mdcoon: The change is actually saying that if "hideFooter" is set, it will show the button. I think you want that reversed, correct?

orenyodfat: > > should there be an "envelope" here ? I do not see that .
> > here is the review app https://alchemy-dev-mradczgvxpkcjdiuxh.herokuapp.com/daos/
> > @mdcoon
> > Screen Shot 2020-06-17 at 17 22 59
> 
> Yes. I did a push this morning to the BUIDLHub fork that was missing assets. I wonder if it's not pulled into the PR.

it was a review app issue . 
the envelop showing up now ! https://alchemy-dev-oxyfc3mtejjbgoiq9d.herokuapp.com/

dkent600: > The change is actually saying that if "hideFooter" is set, it will show the button. I think you want that reversed, correct?

Oops, sorry, lost that in an edit.  Yes please, reverse the logic.

src/components/Daos/Daos.scss
dkent600: I would use inline-block to get the effect you want while minimizing potential side-effects.  If it looks good visually, I'm OK with that.

src/components/Daos/DaosPage.tsx
```diff
@@ -66,13 +68,24 @@ class DaosPage extends React.Component<IProps, IState> {
...
     if (nowMobile !== this.state.isMobile) {
       this.setState({ isMobile: nowMobile });
     }
   }
 
+  private registerForMonitoring = () => {
+    showSimpleMessage(
+      {
+        title: "BUIDLHub Proposal Monitoring",
+        body: <BHubReg />,
```
dkent600, line 83: Suddenly realized what you're doing here.  `showSimpleMessage` wasn't originally envisioned as being a container for any arbitrary form, rather to be just for simple OK/Cancel messages.  Normally for a complicated form with its own logic and idiosyncratic buttons, etc, we would go straight to `ModalPopup`, like `SimpleMessagePopup` itself is doing.

So I am nervous about asking `showSimpleMessage` to support this use case.  But I suppose if it is working then OK.

mdcoon: Right. I was advised to use what's there vs. creating a whole new custom modal popup. I saw the "hideFooter" property but didn't think to just do what you suggested to make it actually work. But committed changes as you suggested.

orenyodfat:
unauthorise domain ? how to authorise ?
Screen Shot 2020-06-18 at 8 20 13
@mdcoon

mdcoon:
> unauthorise domain ? how to authorise ?
> Screen Shot 2020-06-18 at 8 20 13
> @mdcoon

This has to do with the referrer domain. I've disabled referrer validation on our end until this is launched. Please try again.

mdcoon:
> > unauthorise domain ? how to authorise ?
> > Screen Shot 2020-06-18 at 8 20 13
> > @mdcoon
> 
> This has to do with the referrer domain. I've disabled referrer validation on our end until this is launched. Please try again.

Did this work?

mdcoon:
Not sure why CI is failing after changing css property.

PR #1846, opened 2020-06-17
use @daostack/arc.js - draft

roienatan:
- Using @daostack/arc.js instead of @dorgtech/arc.js.
- Updated import in files.
- Updated docker-compose.yml to fetch test env 4.0.11

Facing errors with metamask while trying to create a new proposal.
**Pushing as draft for debug**


roienatan:
Looks like it's kind of a cache issue with metamask. Re-installation of meta mask extension solves the problem.

orenyodfat:
please update to testenv exp 4.0.11 at package.json https://github.com/daostack/alchemy/blob/9a79727e4551ee1d9d010eea6ad1f83a7cfe6130/package.json#L152

test/integration/utils.ts
```diff
@@ -14,7 +14,7 @@ const chai = require("chai");
 global.expect = chai.expect;
 chai.Should();
 
-export const LATEST_ARC_VERSION = "0.1.1-rc.20";
+export const LATEST_ARC_VERSION = "0.1.1-rc.21";
```
orenyodfat, line 17: This can be automated by fetching the version from @daostack/test-env-experimental/package.json @roienatan

test/integration/utils.ts
```diff
@@ -1,4 +1,4 @@
-import { Arc, IArcOptions } from "@dorgtech/arc.js";
+import { Arc, IArcOptions } from "@daostack/arc.js";
 
 const settings: IArcOptions & { txSenderServiceUrl: string } = {
   graphqlHttpProvider: "http://127.0.0.1:8000/subgraphs/name/daostack",
```
orenyodfat, line 4: These value should be taken from subgraph_endpoints.json @roienatan

test/integration/utils.ts
roienatan: Here it supposed to be from `migration-experimental` insted of `test-env-experimental`.

PR #1854, opened 2020-06-18
Automate params in test/integration/utils

roienatan:
- `LATEST_ARC_VERSION` is now fetched from `@daostack/test-env-experimental/package.json`
- Subgraph endpoints are now fetched from `subgraph_endpoints.json`

### DIFF
--- a/test/integration/utils.ts
+++ b/test/integration/utils.ts
@@ -16,7 +16,7 @@ const chai = require("chai");
 global.expect = chai.expect;
 chai.Should();
 
-export const LATEST_ARC_VERSION = require("@daostack/test-env-experimental/package.json")["dependencies"]["@daostack/migration-experimental"];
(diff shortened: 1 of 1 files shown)


test/integration/utils.ts
```diff
@@ -14,7 +16,7 @@ const chai = require("chai");
 global.expect = chai.expect;
 chai.Should();
 
-export const LATEST_ARC_VERSION = "0.1.1-rc.21";
+export const LATEST_ARC_VERSION = require("@daostack/test-env-experimental/package.json")["dependencies"]["@daostack/migration-experimental"];
```
orenyodfat, line 19: migration-experimental version is not the same as arc version. 

it should be `require("@daostack/migration-experimental/package.json")["dependencies"]["@daostack/arc-experimental"];`

roienatan: Right, in the other coment it was written `test-env-experimental`
I'll fix it.

Issue #1856, opened 2020-06-18
Proposal data/info cannot be seen


orenyodfat:
**Describe the bug**
Proposal data/info cannot be seen

**To Reproduce**
https://alchemy.daostack.io/dao/0x519b70055af55a007110b4ff99b0ea33071c720a/proposal/0x3e579ae173062b057d09ca5a1791e36d62989c59071b5b065801792a845b0dc6

this proposal has a relative long description : 
[...]

PR #1857, opened 2020-06-19
handle wide GenericScheme calldata

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1856

- truncates the callData value and adds CopyToClipboard button
- makes contract address fully visible (consistent with unknown schemes)
- makes a few needed cosmetic improvements consistent with other proposal details pages
- makes the etherscan link open safely, and in a new tab

### DIFF
--- a/src/components/Proposal/ProposalSummary/ProposalSummary.scss
+++ b/src/components/Proposal/ProposalSummary/ProposalSummary.scss
@@ -111,6 +111,10 @@
     padding: 20px;
     margin: 0;
 
+    a {
+      color: #0071ff;
+    }
+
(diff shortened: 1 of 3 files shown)


orenyodfat:
- Does other (none generic) proposals pages handling long desctiption / titles ?

dkent600:
> * Does other (none generic) proposals pages handling long desctiption / titles ?

See the PR title and description.  This has nothing to do with description and title.  It has to do with the calldata being very long (you can tell that just by carefully inspecting the page that has the problem).

I don't see any other places where we display the raw callData.

PR #1864, opened 2020-06-20
Added default future date for activation time

roienatan:
When creating a proposal, activation time is now given a default date for the same hour the next day.

### DIFF
--- a/src/components/Proposal/Create/PluginForms/CreatePluginManagerProposal.tsx
+++ b/src/components/Proposal/Create/PluginForms/CreatePluginManagerProposal.tsx
@@ -20,6 +20,7 @@ import * as React from "react";
 import * as css from "../CreateProposal.scss";
 import MarkdownField from "./MarkdownField";
 import { PluginInitializeFields } from "./PluginInitializeFields";
+import * as moment from "moment";
 
 interface IExternalProps {
   daoAvatarAddress: string;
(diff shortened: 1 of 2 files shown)


src/components/Proposal/Create/PluginForms/CreatePluginManagerProposal.tsx
```diff
@@ -162,7 +163,7 @@ class CreatePluginManagerProposal extends React.Component<IProps, IState> {
       votersReputationLossRatio: 4,
       minimumDaoBounty: 150,
       daoBountyConst: 10,
-      activationTime: new Date(),
+      activationTime: moment().add(1, "day").format("YYYY-MM-DDTHH:mm"), //default date for the next day
```
orenyodfat, line 166: how about the default date value will be 0 (1.1.1970 00:00) ?
in that case there will be no constraint re the activation time by default .at the contract level .

roienatan: Tha nature of activation time is to be a future date, so I think it'll be confusing if the default date will be a past date.

roienatan: Also from a UX perspective, as the rest of the default params that are initialized are all valid, so this one also must be valid.

orenyodfat: ok. lets have it this way for now. 👍

PR #1865, opened 2020-06-21
Decode pluginToRegisterData and extract the real plugin name 

roienatan:
When creating a new plugin, if the plugin type is "ContributionRewardExt" the real name is extracted from the decoded pluginToRegisterData.

### DIFF
--- a/src/lib/pluginUtils.ts
+++ b/src/lib/pluginUtils.ts
@@ -16,6 +16,14 @@ import "moment";
 import * as moment from "moment-timezone";
 
 import { getArc } from "../arc";
+const WEB3 = require("web3");
+const web3 = new WEB3();
+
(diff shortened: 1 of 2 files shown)


src/lib/pluginUtils.ts
```diff
@@ -16,6 +16,14 @@ import "moment";
 import * as moment from "moment-timezone";
 
 import { getArc } from "../arc";
+const WEB3 = require("web3");
+const web3 = new WEB3();
```
orenyodfat, line 20: how about moving this inside the function as local var ?

roienatan: Right.

orenyodfat: also .. `WEB3` -> `Web3`

PR #1870, opened 2020-06-22
Support arc v.23

roienatan:


roienatan:
Travis fails on test `rewards.ts`. That's interesting since it's not consistent because it didn't fail in other tests. The test suggests that when resizing the browser we should be redirected from `http://127.0.0.1:3000` to `http://127.0.0.1:3000/redemptions`. Why resizing a window should redirect us to other URL? What's the point here?
@dkent600 maybe you know this code? 

*The test also fails on another thing that I'm pretty sure I know how to fix.

roienatan:
Ok I figured out why the test failed. 
When the redemption button is rendered it asks for `window size`.
`window size` = mobile ---> render redirection to `/redemptions`. 
`window size` = desktop ---> render popup.
**The thing is that the button component checks it only once at the launching of the app without an event listener to window resizing.** 
With this information in mind let's go to the test. First it opens the browser and only **then** resizing the window, hence the button has already rendered with desktop dimensions so instead of redirection it opens the pop-up ---> failure :(
The simple solution is to first set the dimensions and only then to open the browser.

Issue #1844, opened 2020-06-16
Visual bug when viewing proposals created before earning rep


jessebmiller:
When viewing a proposal, this one happened to be boosted, that was created before I had any rep, the text "Must have had reputation in this DAO when the proposal was created" overlaps with the "Votes n votes >" text.

**To Reproduce**
1. Be running fedora linux
2. Use Frefox
3. Use Metamask
4. Open the details of a proposal that was submitted before you had any rep in the DAO

**Expected behavior**
I'd expect the text to wrap within the right half of the space and not overlap with the words on the left in blue

**Screenshot**
![Screenshot_2020-06-16 Alchemy DAOstack](https://user-images.githubusercontent.com/471702/84827071-89a8c000-afe9-11ea-9d49-86999bcc08f8.png)

PR #1872, opened 2020-06-22
fix no rep when created message

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1844

- another case of absolute positioning causing problems.  So I refactored some code
- also looks good in the proposal card popup:

image

- This:

image

becomes:

image

- This: 
image

becomes:

image


dkent600:
> * In the first image I think a little bit more padding is necessary, you can see the words "when" and "was" they are too close too the edge. If not padding maybe smaller font size?

Yeah I agree.  So used to that popup looking ugly that I didn't notice.  I'll add a little padding.

Issue #1858, opened 2020-06-19
Zendesk request for DAO homepage change


EzraWeller:
From zendesk:
```
Hi,

I would like to change the default description on the homepage of my DAO on Alchemy (on Rinkeby).
DAO - https://alchemy-staging-rinkeby.herokuapp.com/dao/0x9003196E314e03daE65f0bF26788befC714D68a1

Could you please modify the description to:
~~~~~
Welcome to The Venus Project DAO, a decentralized organization built on DAOstack.
This is a DAO for the purpose of exploring new governance models and platforms for www.thevenusproject.com
Visit the Proposals page to make a proposal to the DAO or vote on existing proposals.
~~~~~

Thanks
```

orenyodfat:
We still do not have a generic solution for setting dao description. the DAO's header feature is still WIP. 
Till we will have a generic workable solution  , lets have an hard coded registry which map daos addresses to its description text. if such mapping does not exist , the "default" description will be used. lets start with this one.

dkent600:
@EzraWeller Do you think they really want to display "www.thevenusproject.com", or rather "The Venus Project"?  (either way is easily presumed to be a link)

PR #1878, opened 2020-06-23
custom landing page for the Venus Project

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1858

- Also moves all custom DAO data into a separate file.  This file is a tsx file to allow the content to contain react components.
- Has this unresolved question:  https://github.com/daostack/alchemy/issues/1858#issuecomment-648484206.  I settled on "The Venus Project", hot, of course.  @EzraWeller 
- adds presumption that a DAO's address cannot be presumed unique across networks.

### DIFF
--- a/src/customDaoInfo.tsx
+++ b/src/customDaoInfo.tsx
@@ -1,0 +1,30 @@
+import * as React from "react";
+import { Link } from "react-router-dom";
+import { ReactElement } from "react";
+
+interface IDaoCustomInfo {
+  [address: string]: ReactElement | string;
+}
+
+const data: IDaoCustomInfo = {
+  "0x9003196e314e03dae65f0bf26788befc714d68a1": (
+    <>
+      <div>Welcome to The Venus Project DAO, a decentralized organization built on DAOstack.</div>
+      <div>This is a DAO for the purpose of exploring new governance models and platforms for <a href="https://www.thevenusproject.com" target="_blank" rel="noopener noreferrer">The Venus Project</a>.</div>
+      <div>Visit the <Link to={"/dao/0x9003196E314e03daE65f0bF26788befC714D68a1/schemes/"}>Proposals page</Link> to make a proposal to the DAO or vote on existing proposals.</div>
+    </>
+  ),
+  "0xfaf05fedf06cac499b899d6a2052f23ae239b29d": (
+    <>
+      <div>Welcome to the SoS Collective digital co-op.</div>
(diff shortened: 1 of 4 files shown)


src/customDaoInfo.tsx
```diff
@@ -0,0 +1,40 @@
...
+  [network: string]: { [address: string]: ReactElement | string };
+}
+
+const data: IDaoCustomInfo = {
+  ganache: {
+  },
+  rinkeby: {
+    "0x9003196e314e03dae65f0bf26788befc714d68a1": (
+      <>
+        <div>Welcome to The Venus Project DAO, a decentralized organization built on DAOstack.</div>
```
orenyodfat, line 15: does the description need to be with these tags ?
can it be just text?

dkent600: Just text would be extremely limited.

dkent600: If you are want to eliminate react.js components like `<Link`, then you lose important features like being able to cleanly navigate to the Plugin Proposals page.

dkent600: What is the issue?

dkent600: BTW, if a DAO wants just text, then they can.  Tags are not required.

orenyodfat: I think that having a simple description text is cleaner and might be used as dao description in other pages , with different formats.. and even not in alchemy. 
@roienatan @dkent600 ?

roienatan: Depends how much the link for example is important to us. It's also possible to use plain text and add a "GO TO" button or a link near it which directs us to the url.

orenyodfat: The link might be important , though the description can be formatted as dao.json :
{
  "description": "the mother of all daos",
  "url": "www.blala.org"
}
Giving it another thought , we might consider to index that at subgraph level.,as dao metadata,. and dapps(alchemy) could fetch that from there.
This will be easy , and relative fast to implement (in relation to the dao header page feature). 
We will take this approach in alchemy 2.0.

dkent600: I would suggest:

That the needs for enforcing a structured landing page, where DAOstack defines the structure, should be enumerated before implementing.  For example, it might facilitate the user in entering very simple custom data, and might facilitate searching for specific data, such as a website URL, using metadata to refine the search.

And that those needs should be weighed against what believe would be the futility of predicting a structure acceptable to all DAOs.

I think some combination of meta-structure and absolute flexibility would be ideal.

PR #1884, opened 2020-06-24
Show voting params to proposals that were created

roienatan:
Shows voting params to proposals that were created.
For GenericScheme show also contract to call.


src/lib/pluginUtils.ts
```diff
@@ -17,14 +17,47 @@ import * as moment from "moment-timezone";
...
+
+/**
+ * @param {string} pluginName Plugin name
+ * @param {string} pluginToRegisterData Plugin encoded data
+ * @returns {any} An object containing voting params, real plugin name and contract to call if acceptable
+ */
+export const decodePluginToRegisterData = (pluginName: string, pluginToRegisterData: string): any => {
   const WEB3 = require("web3");
   const web3 = new WEB3();
+  const PLUGIN_PARAMS = require("./plugin_params.json");
```
orenyodfat, line 29: does this `require` inside function scope is following our code standard ?
should it be moved to the beginning of the file ?

src/lib/pluginUtils.ts
```diff
@@ -17,14 +17,47 @@ import * as moment from "moment-timezone";
...
   const encodedDataForWeb3 = "0x" + pluginToRegisterData.substring(10);
-  const decodedData = web3.eth.abi.decodeParameters(["address", "address", "uint256[11]", "address", "bytes32", "address", "uint64[3]", "string"], encodedDataForWeb3);
-  return decodedData[7];
+  const decodedData = web3.eth.abi.decodeParameters(PLUGIN_PARAMS[pluginName], encodedDataForWeb3);
```
orenyodfat, line 31: please open an issue to move this decoding to arc.js.

src/components/Proposal/ProposalSummary/ProposalSummaryPluginManager.tsx
```diff
@@ -128,6 +137,23 @@ class ProposalSummary extends React.Component<IProps, IState> {
                           <img className={css.copyButton} src="/assets/images/Icon/Copy-blue.svg" onClick={this.copyToClipboardHandler(proposalState.pluginToRegisterData)} />
                         </td>
                       </tr>
+
+                      { votingParams && (
```
orenyodfat, line 141: is there a need to add check of votingParams and/or decodedData is not `undefined` ?
this can happened if [this condition ](https://github.com/daostack/alchemy/pull/1884/files#diff-9fe300ece25e1d080948a828baef1b9fR59) is not met.

src/lib/pluginUtils.ts
roienatan: Here https://github.com/daostack/alchemy/pull/1865 you to suggested to use it as a local var. You mean only the .json file `require` should be outside?

src/components/Proposal/ProposalSummary/ProposalSummaryPluginManager.tsx
roienatan: Actually there is a checking for that. The line itself (141): 
`votingParams &&` ---> meaning if `votingParams` is defined continue and add more table rows.

src/lib/pluginUtils.ts
orenyodfat: I meant only  for `const web3 = new WEB3(); `
the `require`  should be outside.
Also WEB3 -> Web3

PR #1886, opened 2020-06-24
Update tests to use the robust function to hide cookies window

roienatan:


test/integration/utils.ts
```diff
@@ -91,7 +75,7 @@ export async function gotoDaoPlugins(daoAddress: string): Promise<any> {
  * @param void
  * @returns Promise<void>
  */
-export const acceptCookies = async(): Promise<void> => {
+export const hideCookieAcceptWindow = async(): Promise<void> => {
```
dkent600, line 78: the original code checked `!acceptCookiesButton.error`.  You sure you don't need that?  I wouldn't have put it in unless I found it was needed, though I no longer recall the reason.

roienatan: I checked the tests locally and as you can see all the tests passed. It's not a must but I think the old function is not written in a stable way and can cause problems in the future.

dkent600: I'm not questioning modifying the function, rather whether in doing so we should have deleted the check for `!acceptCookiesButton.error`.  I don't see any reason not to continue checking it.

dkent600: But if you can continue that this works in all three use cases, then I can accept it as is:

1) travis
2) locally, headless
3) locally, not headless

roienatan: I'm not sure what's the purpose of `.error` there but I couldn't find more places that `.error` is used in the tests.
By the way, I don't think `.error` is part of the `Jest.js` framework.

Issue #1888, opened 2020-06-24
Plugin Proposals page should not crash when there is no PluginManager registered to the DAO


dkent600:
Plugin Proposals page crashes when there is no PluginManager registered to the DAO.

I am able to see this on the production site on either of these two pages:

https://alchemy.daostack.io//dao/0x3a2750605eff555a59a77a6a9a42bc089cc962be/schemes/
https://alchemy.daostack.io//dao/0xc6adbeae8f774091e7bfed8a824a548a1a222d6b/schemes/

PR #1889, opened 2020-06-24
don't crash when no PluginManager

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1888

The review app is pointing to mainnet since the repo cases given in the ticket are pointing there.

### DIFF
--- a/src/components/Dao/DaoSchemesPage.tsx
+++ b/src/components/Dao/DaoSchemesPage.tsx
@@ -80,7 +80,7 @@ class DaoSchemesPage extends React.Component<IProps, null> {
     const allKnownSchemes = [...contributionReward, ...knownSchemes];
 
     const schemeManager = data[1];
-    const schemeManagerActive = schemeManager ? getSchemeIsActive(schemeManager) : null;
+    const schemeManagerActive = schemeManager ? getSchemeIsActive(schemeManager) : false;
 
     const schemeCardsHTML = (
       <TransitionGroup>
--- a/src/components/Dao/DaoSchemesPage.tsx
+++ b/src/components/Dao/DaoSchemesPage.tsx
@@ -80,7 +80,7 @@ class DaoSchemesPage extends React.Component<IProps, null> {
     const allKnownSchemes = [...contributionReward, ...knownSchemes];
 
     const schemeManager = data[1];
-    const schemeManagerActive = getSchemeIsActive(schemeManager);
+    const schemeManagerActive = schemeManager ? getSchemeIsActive(schemeManager) : null;
 
     const schemeCardsHTML = (
       <TransitionGroup>
--- a/src/lib/schemeUtils.ts
+++ b/src/lib/schemeUtils.ts
@@ -194,6 +194,11 @@ const schemeActionPropNames = new Map<string, Map<GetSchemeIsActiveActions, stri
   ],
 ]);
 
+/**
+ * Returns true or false indicating whether the scheme is active and thus can accept new proposals.
+ * @param scheme Required parameter that if undefined or null cause this method to throw an exception.
+ * @param action For SchemeRegistrar where we are asking specifically about Add and Remove actions.
+ */
 export function getSchemeIsActive(scheme: ISchemeState, action?: GetSchemeIsActiveActions): boolean {
   let votingMachineParamsPropertyName: string;
   let schemeName = scheme.name ? `${scheme.name[0].toLowerCase()}${scheme.name.slice(1)}` : "";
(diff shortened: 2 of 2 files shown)


src/components/Dao/DaoSchemesPage.tsx
```diff
@@ -80,7 +80,7 @@ class DaoSchemesPage extends React.Component<IProps, null> {
     const allKnownSchemes = [...contributionReward, ...knownSchemes];
 
     const schemeManager = data[1];
-    const schemeManagerActive = getSchemeIsActive(schemeManager);
+    const schemeManagerActive = schemeManager ? getSchemeIsActive(schemeManager) : null;
```
orenyodfat, line 83: need to check scheme validity at getSchemeIsActive .
getSchemeIsActive should not crash alchemy in any case.

dkent600: I added comments for the `getSchemeIsActive` method and a check in `getSchemeIsActive` for `scheme` that throws a more clear exception when `scheme` is not set.

I can find no other incorrect uses of `getSchemeIsActive` in the app than the one I fixed.

src/lib/schemeUtils.ts
```diff
@@ -194,7 +194,17 @@ const schemeActionPropNames = new Map<string, Map<GetSchemeIsActiveActions, stri
...
 
+/**
+ * Returns true or false indicating whether the scheme is active and thus can accept new proposals.
+ * @param scheme Required parameter that if undefined or null will cause this method to throw an exception.
+ * @param action For SchemeRegistrar where we are asking specifically about Add and Remove actions.
+ */
 export function getSchemeIsActive(scheme: ISchemeState, action?: GetSchemeIsActiveActions): boolean {
+
+  if (!scheme) {
+    throw new Error("getSchemeIsActive: scheme parameter is not set");
```
orenyodfat, line 205: does this will crash alchemy ?

src/components/Dao/DaoSchemesPage.tsx
orenyodfat, line 83: schemeManagerActive is active is boolean .. should it get `null` value ?

orenyodfat: - when and how schemeManager can be null/false ? 
- how about getSchemeIsActive should return false if the scheme is null as null scheme is not active ?

src/lib/schemeUtils.ts
orenyodfat: how about it will just return false here as null scheme is not an active one ?

src/components/Dao/DaoSchemesPage.tsx
dkent600: > * when and how schemeManager can be null/false ?

See the ticket this resolves: https://github.com/daostack/alchemy/issues/1888

> * how about getSchemeIsActive should return false if the scheme is null as null scheme is not active ?

That would be misleading, implying to the calling code that it has passed a valid scheme that is legitimately found to be not active, which is not the case.  Aberrant behavior would inevitably follow in the calling code.  The calling code must be responsible for properly handling the case that there is no scheme before calling this function.

src/lib/schemeUtils.ts
dkent600: That would be the wrong thing to do.  See https://github.com/daostack/alchemy/pull/1889#discussion_r452294523

src/components/Dao/DaoSchemesPage.tsx
dkent600: changed to false

dkent600: Throwing an exception on invalid parameters is a very common programming pattern.

src/lib/schemeUtils.ts
orenyodfat: does this error manifested in the ui somehow or just in the logs ?

src/components/Dao/DaoSchemesPage.tsx
orenyodfat: Throwing an exception which is manifested in the UI is a bad thing though

orenyodfat:
how about other calls to getSchemeIsActive such as https://github.com/daostack/alchemy/blob/e62828a098f95862fe243d359c16c2694d3676aa/src/components/Scheme/SchemeContainer.tsx#L126

dkent600:
> how about other calls to getSchemeIsActive such as
> 
> https://github.com/daostack/alchemy/blob/e62828a098f95862fe243d359c16c2694d3676aa/src/components/Scheme/SchemeContainer.tsx#L126

As I mentioned [here](https://github.com/daostack/alchemy/pull/1889#discussion_r447910154) I've checked all the other uses.  They were already doing the right things.

src/lib/schemeUtils.ts
dkent600: I do not see a way in which our code, with this PR, can cause this exception to happen.

src/components/Dao/DaoSchemesPage.tsx
dkent600: Sure, but what you propose isn't the right way to address that.  There are already a jillions ways in which exceptions can occur because good code throws exceptions when unexpected things happen. We strive to program in a way that avoids them, but it is not feasible to eliminate all of them.  One of the things we do to handle the situation is to put up a relatively friendly error page, and in so doing, we are in effect handling all exceptions.

PR #1895, opened 2020-06-25
Disable closing modal when clicking on the modal backdrop (outside)

roienatan:
I also noticed that scrolling behind a new plugin proposal modal is still possible when the modal is open and this is a bad UX so now scrolling outside of a new plugin proposal modal is disabled.

### DIFF
--- a/src/components/Proposal/Create/CreateProposal.scss
+++ b/src/components/Proposal/Create/CreateProposal.scss
@@ -1,10 +1,6 @@
 @import "./react-mde.css";
 :global {
   @import "../../../../node_modules/react-datetime/css/react-datetime";
-
-  body.react-router-modal__modal-open {
-    overflow: hidden;
-  }
 }
 
 .hidden {
--- a/src/components/Proposal/Create/CreateProposal.scss
+++ b/src/components/Proposal/Create/CreateProposal.scss
@@ -1,6 +1,10 @@
 @import "./react-mde.css";
 :global {
   @import "../../../../node_modules/react-datetime/css/react-datetime";
+
+  body.react-router-modal__modal-open {
+    overflow: hidden;
+  }
 }
 
 .hidden {
--- a/src/components/Proposal/Create/CreateProposalPage.tsx
+++ b/src/components/Proposal/Create/CreateProposalPage.tsx
@@ -62,10 +62,6 @@ class CreateProposalPage extends React.Component<IProps, IStateProps> {
     history.push("/dao/" + daoAvatarAddress + "/plugin/" + pluginId);
   }
 
-  private preventDefault = (e: any) => {
-    e.preventDefault();
-  }
-
   public async componentDidMount() {
     document.addEventListener("keydown", this.handleKeyPress, false);
     (document.getElementsByTagName("BODY")[0] as HTMLElement).style.overflow = "hidden";
--- a/src/components/Proposal/Create/CreateProposalPage.tsx
+++ b/src/components/Proposal/Create/CreateProposalPage.tsx
@@ -64,7 +64,6 @@ class CreateProposalPage extends React.Component<IProps, IStateProps> {
 
   public async componentDidMount() {
     document.addEventListener("keydown", this.handleKeyPress, false);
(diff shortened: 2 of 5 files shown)


src/components/Proposal/Create/CreateProposalPage.tsx
```diff
@@ -62,8 +62,13 @@ class CreateProposalPage extends React.Component<IProps, IStateProps> {
     history.push("/dao/" + daoAvatarAddress + "/plugin/" + pluginId);
   }
 
+  private preventDefault = (e: any) => {
```
dkent600: I don't see this being used

roienatan: Yes I noticed this later and removed it in the last push.

src/components/Proposal/Create/CreateProposalPage.tsx
```diff
@@ -62,8 +62,13 @@ class CreateProposalPage extends React.Component<IProps, IStateProps> {
...
+    e.preventDefault();
+  }
+
   public async componentDidMount() {
     document.addEventListener("keydown", this.handleKeyPress, false);
+    (document.getElementsByTagName("BODY")[0] as HTMLElement).style.overflow = "hidden";
```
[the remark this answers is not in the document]
roienatan: I thought about this but for some wired reason `ModalPopup` is not the modal that is used in the modal of the `createProposalPage`, hence changing the `.tsx/.scss` in `ModalPopup` is not affecting `createProposalPage modal`. That's another bad practice that the "generic" modal is not used all across Alchemy.

roienatan:
@dkent600 Regarding your question: _Where is the code that prevents an outside click from closing the popup?_
See my first commit in `AppContainer.tsx`. The deletion of the line there preventing the modal from being closed while clicking outside of the modal.

dkent600:
> @dkent600 Regarding your question: _Where is the code that prevents an outside click from closing the popup?_
> See my first commit in `AppContainer.tsx`. The deletion of the line there preventing the modal from being closed while clicking outside of the modal.

Really just this?  Very counter-intuitive.

image

src/components/Proposal/Create/CreateProposalPage.tsx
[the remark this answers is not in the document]
dkent600: Ah yes, I forgot about where we use `ModalRoute`.  I believe the purpose is that the popup be accessible via defined routes that match a browser URL.

All you have to do is add this to the :global section at the top of CreateProposal.scss

```
  body.react-router-modal__modal-open {
    overflow: hidden;
  }
```

I totally agree that this is not ideal structure.  We might be able to refactor the structure considering the ModalRoute properties here:  https://github.com/davidmfoley/react-router-modal#modalroute, but I believe we should do that in a separate PR since it involves very different use cases for testing.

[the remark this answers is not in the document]
dkent600: We should also look at uses of SimpleMessagePopup, though I don't know if that is in 2.0 yet.

roienatan:
> > @dkent600 Regarding your question: _Where is the code that prevents an outside click from closing the popup?_
> > See my first commit in `AppContainer.tsx`. The deletion of the line there preventing the modal from being closed while clicking outside of the modal.
> 
> Can you please confirm whether is really just this? Very counter-intuitive.
> 
> image

Right, it's just this. I agree it's counter-intuitive.

src/components/Proposal/Create/CreateProposal.scss
```diff
@@ -1,6 +1,10 @@
 @import "./react-mde.css";
 :global {
   @import "../../../../node_modules/react-datetime/css/react-datetime";
+
+  body.react-router-modal__modal-open {
```
dkent600, line 5: Does this not also need to be in ModalPopup.scss, to cover all the other possible popups?

roienatan: Agree, but first I want to locate those popup and see that it's affecting.

roienatan: Okay looks like the above code also affecting ModalPopup.

PR #1903, opened 2020-06-28
Support Torus (issue #1829)

roienatan:


roienatan:
There is a problem with the z-index of the Torus popup. I'm fixing it...

orenyodfat:
see https://github.com/daostack/alchemy/commit/870ec7dbd406cdfd830b348c156e97a04ba47929

orenyodfat:
> There is a problem with the z-index of the Torus popup. I'm fixing it...

see https://github.com/daostack/alchemy/commit/870ec7dbd406cdfd830b348c156e97a04ba47929

roienatan:
Right, I saw it on master branch. It's interesting that locally I had no problem with the z-index.

PR #1917, opened 2020-07-01
Data decoding is now in arc.js

roienatan:


orenyodfat:
need to import new arc.js https://www.npmjs.com/package/@daostack/arc.js/v/2.0.0-experimental.42  :)

PR #1921, opened 2020-07-02
Release 0.10.12

dkent600:


dkent600:
> Seems like we are still leaving gas price at 0 gwei on xdai as well, which is still leading to stalled transactions. Is it possible to fix that, or do we need a new issue for it?

That is not fixed in this release.  There is already a ticket for it.

Issue #1813, opened 2020-06-09
can't change accounts when using ganache


dkent600:
When running locally using ganache, it is not possible to change accounts with MetaMask.  You can change the account in MM, but Alchemy will ignore it.  This is a major problem when trying reproducing certain use cases.

The problem is that `arc.inTesting` is returning true when it shouldn't.

PR #1922, opened 2020-07-02
Merge from dev to 2.0

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1897


.gitignore
```diff
@@ -16,3 +16,4 @@ schema.graphql
 subgraph.yaml
 build/
 *.scss.d.ts
+*.ps1
```
orenyodfat, line 19: what is ps1 ?

orenyodfat:
why removed docs/Alchemy_2_0_goods.md ?

.gitignore
dkent600: powershell scripts

orenyodfat: for what it is needed/use?

src/arc.ts
```diff
@@ -271,7 +270,7 @@ export interface IEnableWalletProviderParams {
 }
 
 function inTesting(): boolean {
-  if (process.env.NODE_ENV === "development" && navigator.webdriver) {
+  if (process.env.NODE_ENV === "development" && (global as any).inAlchemyTests) {
```
orenyodfat, line 273: inAlchemyTests ?

src/components/Dao/DaoHistoryPage.tsx
```diff
@@ -97,8 +93,8 @@ class DaoHistoryPage extends React.Component<IProps, null> {
 
 export default withSubscription({
   wrappedComponent: DaoHistoryPage,
-  loadingComponent: <Loading/>,
-  errorComponent: (props) => <div>{ props.error.message }</div>,
+  loadingComponent: <Loading />,
```
orenyodfat, line 96: does this space is a mistake ?

dkent600: no

src/arc.ts
dkent600: See https://github.com/daostack/alchemy/issues/1813

.gitignore
dkent600: to prevent powerscript scripts from being added to the repository

orenyodfat: what is power scripts ? and why alchemy is using that ? did we had that before ?

dkent600: has nothing to do with Alchemy per se.  I have occasion to use powerscript scripts.

orenyodfat: so please exclude that from .gitignore

orenyodfat:
Screen Shot 2020-07-09 at 14 40 39

ethBalance does not work ?!

dkent600:
> Screen Shot 2020-07-09 at 14 40 39
> 
> ethBalance does not work ?!

Apparently not for that DAO.  You can see in other DAOs is it returning 0.  I would like to check if this behavior is different from what is in the dev-2 branch, but the dev-2 branch won't even let you get into a DAO.

.gitignore
dkent600: .chrome and .vscode are no different in this respect, but they are very desirable.  I'll go ahead and remove .ps1 anyway since it is pretty easy to work-around.  However, it does increase the possibility of people committing files we don't want.

orenyodfat:
> > ethBalance does not work ?!
> 
> Apparently not for that DAO. You can see in other DAOs is it returning 0. I would like to check if this behavior is different from what is in the dev-2 branch, but the dev-2 branch won't even let you get into a DAO.

it is :) 
https://alchemy-2-rinkeby.herokuapp.com/dao/0x365e8c20f57f5acc97a7ed0125d4f4fd41a221a9

dkent600:
> ethBalance does not work ?!

@orenyodfat this was due to a bad infura id.  is 0 now instead of N/A

PR #1923, opened 2020-07-05
xdai wss

roienatan:


src/settings.ts
```diff
@@ -116,7 +116,7 @@ function getWeb3ConnectProviderOptions(network: string) {
           package: Torus,
           options: {
             networkParams: {
-              host: "https://xdai.poanetwork.dev",
+              host: "wss://xdai.poanetwork.dev/wss",
```
orenyodfat, line 119: for torus it should stay https

dkent600: Strange, I wonder how the other providers are able to create websocket connections with https?

orenyodfat:
seems like https will do the work at 2.0 due to https://github.com/daostack/arc.js/blob/client-2-0/src/arc.ts#L345 .. 
(any how , all other networks use https and not wss for now)

PR #1928, opened 2020-07-06
updated http_xdai endpoint

roienatan:


src/subgraph_endpoints.json
```diff
@@ -5,7 +5,7 @@
   "ws_rinkeby": "wss://api.thegraph.com/subgraphs/name/daostack/v8_7_exp_rinkeby",
   "http_kovan": "https://api.thegraph.com/subgraphs/daostack/name/daostack/v8_7_exp_kovan",
   "ws_kovan": "wss://api.thegraph.com/subgraphs/daostack/name/daostack/v8_7_exp_kovan",
-  "http_xdai": "https://api.thegraph.com/subgraphs/daostack/name/daostack/v8_7_exp_xdai",
+  "http_xdai": "https://api.thegraph.com/subgraphs-daostack/name/daostack/v39_4_xdai",
```
orenyodfat, line 8: the subgraph version should stay as is `v8_7_exp_xdai` ..
only the url path is different

orenyodfat: please do the same for all network ..
same as in https://github.com/daostack/alchemy/blob/dev/src/subgraph_endpoints.json#L8

PR #1932, opened 2020-07-06
added plugins UI names

roienatan:
It's not possible just to change the names of the plugins since the names themselves are used in more places not just as display names but as dom objects IDs, keys and more.
Instead, I created a new object called: `PLUGINS_UI_NAMES`. Here we can give the display name to each plugin without affecting the IDs, keys and more.

### DIFF
--- a/src/lib/pluginUtils.ts
+++ b/src/lib/pluginUtils.ts
@@ -46,34 +46,13 @@ export const REQUIRED_PLUGIN_PERMISSIONS: any = {
 
 /** plugins that we know how to interpret  */
 export const PLUGIN_NAMES = {
-  ContributionReward: "ContributionReward",
-  GenericScheme: "GenericScheme",
-  ReputationFromToken: "ReputationFromToken",
-  SchemeRegistrar: "SchemeRegistrar",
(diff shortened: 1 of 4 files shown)


src/lib/pluginUtils.ts
```diff
@@ -66,6 +66,16 @@ export const PROPOSAL_PLUGIN_NAMES = [
   "ContributionRewardExt",
 ];
 
+export const PLUGINS_UI_NAMES = {
```
orenyodfat, line 69: do we do not have these name somewhere else in the code?

roienatan: **Short answer:** no
**Long answer:** 
There is a function: `pluginName(plugin: IPluginState|IContractInfo, fallback?: string)` in `pluginUtils.ts` that returns the display name. This function is used in the `plugin to remove`  tab but there you have the expected params of the function (`IPluginState|IContractInfo`).
In `plugin to add` we don't have these variables unless we make a bigger change in the code there.
In addition I don't think this function does exactly what we need (for example: for `GenericScheme` the result may be `Blockchain Interaction`).
Baside that there is no other place in the code with the display names.

Issue #1930, opened 2020-07-06
2.0 cosmetics


orenyodfat:
- [x] round corners on Proposal Plugins elements

roienatan:
Cosmetics in progress: 
- [ ] Equal margin between Proposal Plugins elements
- [ ] Follow button font family to match whole Alchemy
- [ ] Added create proposal modal bottom margin 
- [ ] Plugin manager information tab - elements are now have round borders and margin was added between each element

dkent600:
Please don't start this until the merge from 1.0  is complete

roienatan:
> Please don't start this until the merge from 1.0 is complete

No worries it's for now all the changes are local.

dkent600:
> > Please don't start this until the merge from 1.0 is complete
> 
> No worries, for now all the changes are local.

That's good, but still, unless you are careful in choosing what to work on, there is a non-zero probability you will be fixing something that is already fixed or whose code has significantly changed.

roienatan:
> > > Please don't start this until the merge from 1.0 is complete
> > 
> > 
> > No worries, for now all the changes are local.
> 
> That's good, but still, unless you are careful in choosing what to work on, there is a non-zero probability you will be fixing something that is already fixed or whose code has significantly changed.

I understand, I wasn't sure how many scss changes there are in the merge.

PR #1937, opened 2020-07-08
UI/UX improvments

roienatan:
See https://github.com/daostack/alchemy/issues/1930


dkent600:
@roienatan I very much hope you do not merge this until after we've merged the 1=>2 merge.

roienatan:
Merged conflicts with dev-2 (after the big merge) are now resolved.

src/components/Daos/DaoCard.tsx
```diff
@@ -34,18 +33,13 @@ const DaoCard = (props: IProps) => {
       <div className={classNames({
         [css.dao]: true,
         [css.daoInactive]: inActive})}>
-        <div className={css.daoTitle}
-          style={{backgroundImage: bgPattern.toDataUrl()}}>
-
+        <div className={css.daoTitle} style={{backgroundColor: bgPattern.color}}>
```
dkent600, line 36: We should not be setting styles inline, only in CSS

src/components/Proposal/Create/CreateProposal.scss
```diff
@@ -15,7 +15,7 @@
   left: 50%;
   transform: translate(-50%, -50%);
   box-shadow: $shadow-3;
-  height: calc(100% - 100px);
+  height: calc(80%);
```
dkent600, line 18: Does this change work consistently on all widths (I would check mobile too even though it is lower priority)?

src/components/Daos/DaoCard.tsx
roienatan: The colors come from a third-party module in the JavaScript and if we want to use it we need to apply it from the JavaScript code. As you can see, also the old code worked in that way.

src/components/Proposal/Create/CreateProposal.scss
roienatan: Why wouldn't it? Not sure what you mean.

dkent600: You have replaced a constant 100px with a variable 20%

dkent600: oh, and sorry I should've said "all heights" not "all widths"

Issue #1942, opened 2020-07-09
Proposal creation form, description field label, should be well-formatted and have a training tooltip


dkent600:
CreateKnownGenericSchemeProposal (like "GenericSchemeDutchX") needs a training tooltip:

![image](https://user-images.githubusercontent.com/1821666/87030489-adf25980-c1af-11ea-90a7-cdfc0ea5c3a7.png)

CreateUnknownGenericSchemeProposal ("Blockchain Interaction") is poorly formatted:

![image](https://user-images.githubusercontent.com/1821666/87030287-6370dd00-c1af-11ea-914e-6b1850776251.png)

dkent600:
These are already fixed in 2.0

PR #1944, opened 2020-07-09
fix description label bugs

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1942

Note this is already fixed in 2.0

### DIFF
--- a/src/components/Proposal/Create/SchemeForms/CreateKnownGenericSchemeProposal.tsx
+++ b/src/components/Proposal/Create/SchemeForms/CreateKnownGenericSchemeProposal.tsx
@@ -286,6 +286,7 @@ class CreateKnownSchemeProposal extends React.Component<IProps, IState> {
 
     const actions = this.state.actions;
     const currentAction = this.state.currentAction;
+    const fnDescription = () => (<span>Short description of the proposal.<ul><li>What are you proposing to do?</li><li>Why is it important?</li><li>How much will it cost the DAO?</li><li>When do you plan to deliver the work?</li></ul></span>);
 
     return (
       <div className={css.containerWithSidebar}>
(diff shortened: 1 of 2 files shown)


src/components/Proposal/Create/SchemeForms/CreateKnownGenericSchemeProposal.tsx
```diff
@@ -286,6 +286,7 @@ class CreateKnownSchemeProposal extends React.Component<IProps, IState> {
 
     const actions = this.state.actions;
     const currentAction = this.state.currentAction;
+    const fnDescription = () => (<span>Short description of the proposal.<ul><li>What are you proposing to do?</li><li>Why is it important?</li><li>How much will it cost the DAO?</li><li>When do you plan to deliver the work?</li></ul></span>);
```
roienatan, line 289: Is there a place we can hold constant text like this?
The code can look cleaner if it'll will be fetched from a constants file for example.
Also, I think we use this text in more places so moving this to a separate file can reduce code duplication.

dkent600: > Is there a place we can hold constant text like this?

I would like to do that too, this is often-duplicated code.  I'd say in proposalUtils.  But not in this PR.

orenyodfat: agree , lets move it to a some lable.json somehow.

dkent600: in a separate PR.  not needed in 1.0

dkent600: in any case, I think putting this in a json file may be overkill, over-engineering unless we can show a need for it.

dkent600: note that fnDescription  needs to be a component...

orenyodfat: moving all texts to specific file  is much cleaner code and well engineered.
will handle that on 2.0.

Issue #1943, opened 2020-07-09
Should be able to navigate to a DAO using ganache


dkent600:
STEPS:

1. in DAOS list page, click on a DAO

RESULT:

Crash

![image](https://user-images.githubusercontent.com/1821666/87030083-1987f700-c1af-11ea-9a8e-c61c99e57239.png)

dkent600:
This is already fixed in 2.0

PR #1945, opened 2020-07-09
fix access to DAOs on ganache

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1943

Note this is already fixed in 2.0

### DIFF
--- a/src/components/Dao/DaoLandingPage.tsx
+++ b/src/components/Dao/DaoLandingPage.tsx
@@ -6,7 +6,7 @@ import Analytics from "lib/analytics";
 import { Link } from "react-router-dom";
 import { BreadcrumbsItem } from "react-breadcrumbs-dynamic";
 import { DiscussionEmbed } from "disqus-react";
-import { showSimpleMessage } from "lib/util";
+import { showSimpleMessage, targetedNetwork } from "lib/util";
 import customDaoInfo from "../../customDaoInfo";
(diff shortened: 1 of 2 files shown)


src/components/Dao/DaoLandingPage.tsx
```diff
@@ -42,7 +42,7 @@ export default class DaoLandingPage extends React.Component<IProps, null> {
 
   public render() {
     const daoState = this.props.daoState;
-    const customData = customDaoInfo[process.env.NETWORK][daoState.id.toLowerCase()];
+    const customData = customDaoInfo[process.env.NETWORK]?.[daoState.id.toLowerCase()];
```
orenyodfat, line 45: how and why this is manifested only in ganache ?

dkent600: happens when "ganache" is called "private".  Your question made me realize that we should be using `targetedNetwork` instead of `process.env.NETWORK`.  So I changed that in this PR, and will also change it in the 1=>2 merge PR.

orenyodfat: so this is manifested on all other networks?

orenyodfat: please do not add that to the merge to 2.0 pr..
you can pr that separately if you wish .

Issue #1837, opened 2020-06-16
Remove use of forceUpdate


dkent600:
Regarding this line:  https://github.com/daostack/alchemy/blob/d2caa82d04138105944bed2b7b667dc500cbeee4/src/layouts/Header.tsx#L133

`forceUpdate` should not be used, it is an anti-pattern.   Moreover, it should not be needed here, just as it is not needed in Alchemy 1.0.  The fact it was added suggests that something is wrong with how account changes were coded in the original merge from 1.0.

This is one of the comments I made on it in the original merge PR:

https://github.com/daostack/alchemy/pull/1646#discussion_r429488912

PR #1946, opened 2020-07-09
eliminate forceUpdate in Header

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1837

I'm not sure why this was not needed in 1.0, but I have a feeling in 1.0 it was more of an intermittent bug, not frequent enough to motivate us to look into it.

### DIFF
--- a/src/layouts/Header.tsx
+++ b/src/layouts/Header.tsx
@@ -32,6 +32,8 @@ interface IStateProps {
   showRedemptionsButton: boolean;
   currentAccountProfile: IProfileState;
   currentAccountAddress: string | null;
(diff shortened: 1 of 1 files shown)


src/layouts/Header.tsx
```diff
@@ -32,6 +32,8 @@ interface IStateProps {
   showRedemptionsButton: boolean;
   currentAccountProfile: IProfileState;
   currentAccountAddress: string | null;
+  currentAccountIsEnabled: boolean;
```
orenyodfat, line 35: this is related to this.forceUpdate?

dkent600: yes

PR #1959, opened 2020-07-14
Updating daocreator experimental package to set redirect link from alchemy

### LINKED ISSUE #1948: Wrong URL on xDAI deployment
Right now, when you use the DAO creator on xDAI, after deployment, the links below go to the website: 

https://alchemy-staging-xdai.herokuapp.com/dao/....

It should be going to https://alchemy-xdai.daostack.io/dao/... instead

This is the case for both of these links: http://prntscr.com/tfis10

cbrzn:
I added an example on how you should set the URL via props. Now you can, based on the `targetedNetwork`, define where you want to redirect the user when they create a new DAO.

Note: The props just needs the domain (like in the example I did), the rest is handled by daocreator, you can check the code here: https://github.com/dOrgTech/DAOcreator/blob/issue-509-dynamic_link_on_deployment/packages/ui_v2/src/components/commonV2/dao/Migrator/DeployButton.tsx#L452

### DIFF
--- a/src/components/DaoCreator/index.tsx
+++ b/src/components/DaoCreator/index.tsx
@@ -48,7 +48,7 @@ class DaoCreator extends React.Component<IProps> {
             return await getWeb3Provider();
           }}
           noDAOstackLogo
-          redirectURL="https://alchemy-xdai.daostack.io"
+          redirectURL={process.env.BASE_URL}
         />
       </React.Suspense>
     );
--- a/src/components/DaoCreator/index.tsx
+++ b/src/components/DaoCreator/index.tsx
@@ -48,6 +48,7 @@ class DaoCreator extends React.Component<IProps> {
             return await getWeb3Provider();
           }}
           noDAOstackLogo
+          redirectURL="https://alchemy-xdai.daostack.io"
         />
       </React.Suspense>
     );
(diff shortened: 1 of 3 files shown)


src/components/DaoCreator/index.tsx
```diff
@@ -48,6 +48,7 @@ class DaoCreator extends React.Component<IProps> {
             return await getWeb3Provider();
           }}
           noDAOstackLogo
+          redirectURL="https://alchemy-xdai.daostack.io"
```
dkent600, line 51: This does not need to use `targetedNetwork`, can just use `${window.location.protocol}//${window.location.host}` or `process.env.BASE_URL` or some derivation of `RouteComponentProps.location`.  `process.env.BASE_URL` seems easiest.  Whatever the choice, I think the change should be done in this PR.

cbrzn: You call sir! @dkent600

dkent600: @cbrzn Cool.  Did you test this?  If yes and it works, I'll approve the PR

cbrzn: Yup yup @dkent600 - Would if awesome if you can test locally too :-D. I will mark this conv as resolved :+1:

orenyodfat:
I tested that . looks ok.

dkent600:
Hi @cbrzn Looks like there are merge conflicts

cbrzn:
hola! @dkent600 fixed :+1:

PR #1965, opened 2020-07-15
 poll per 60 seconds for eth/token balances

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1963

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -609,15 +609,7 @@ export function ethBalance(address: Address): Observable<BN> {
 
   const arc = getArc();
 
-  if (targetedNetwork() !== "xdai") {
-    return arc.ethBalance(address);
-  }
-
   /**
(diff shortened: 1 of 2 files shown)


src/lib/util.ts
```diff
@@ -593,7 +593,97 @@ export function initializeUtils(options: IInitializeOptions) {
...
+
+const ethBalanceObservedAccounts: IObservedAccounts = {};
+let ethBalancePollingInterval: any | undefined = undefined;
+
 export function ethBalance(address: Address): Observable<BN> {
+
   const arc = getArc();
-  return arc.ethBalance(address);
+
+  if (targetedNetwork() !== "xdai") {
```
orenyodfat, line 612: we want the same for all network ..no ?

dkent600: oops, yes

dkent600: fixed

Issue #1964, opened 2020-07-15
Locate places that we're pulling the dao balance


roienatan:
See issue https://github.com/daostack/alchemy/issues/1960 and it's PR https://github.com/daostack/alchemy/pull/1962
Instead of `ethBalance(...)` should be `arc.dao(...).ethBalance()`

PR #1974, opened 2020-07-16
Pulling ethBalanceDao

roienatan:
resolves: https://github.com/daostack/alchemy/issues/1964

### DIFF
--- a/src/components/Proposal/ProposalData.tsx
+++ b/src/components/Proposal/ProposalData.tsx
@@ -1,6 +1,6 @@
 import { Address, AnyProposal, IProposalState, IDAOState, IMemberState, IRewardState, Reward, Stake, Vote, Proposal, Member, IContributionRewardProposalState } from "@daostack/arc.js";
 import { getArc } from "arc";
-import { ethErrorHandler, ethBalance } from "lib/util";
+import { ethErrorHandler, ethBalanceDao } from "lib/util";
 
 import BN = require("bn.js");
 import withSubscription, { ISubscriptionProps } from "components/Shared/withSubscription";
@@ -163,7 +163,7 @@ export default withSubscription({
         // TODO: also need the member state for the proposal proposer and beneficiary
         //      but since we need the proposal state first to get those addresses we will need to
         //      update the client query to load them inline
-        concat(of(new BN("0")), await ethBalance(daoState.address)
+        concat(of(new BN("0")), (await ethBalanceDao(daoState.address))
           .pipe(ethErrorHandler())),
         arc.GENToken().balanceOf(currentAccountAddress)
           .pipe(ethErrorHandler()),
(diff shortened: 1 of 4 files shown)


roienatan:
I don't know why github doesn't show Travis here but I can see the job in Travis itself.
image

src/components/Proposal/ProposalData.tsx
```diff
@@ -163,7 +163,7 @@ export default withSubscription({
         // TODO: also need the member state for the proposal proposer and beneficiary
         //      but since we need the proposal state first to get those addresses we will need to
         //      update the client query to load them inline
-        concat(of(new BN("0")), await ethBalance(daoState.address)
+        concat(of(new BN("0")), (await ethBalanceDao(daoState.address))
```
orenyodfat, line 166: why proposalData page need to pull for the dao balance ?

src/components/Redemptions/RedemptionsMenu.tsx
```diff
@@ -198,7 +198,7 @@ const SubscribedMenuItemContent = withSubscription({
     const { currentAccountAddress, proposal } = props;
     const arc = getArc();
     const dao = arc.dao(proposal.coreState.dao.id);
-    const daoEthBalance = concat(of(new BN("0")), await ethBalance(proposal.coreState.dao.id)).pipe(ethErrorHandler());
+    const daoEthBalance = concat(of(new BN("0")), await ethBalanceDao(proposal.coreState.dao.id)).pipe(ethErrorHandler());
```
orenyodfat, line 201: why the redemption page need to pole for eth balance ? should it just read that to see the dao balance and calculate if the redemption is possible.?

src/components/Proposal/ProposalData.tsx
dkent600: the question is out of scope of this PR, but the answer is to feed the `ActionButton` component which uses it for the redeem button.

dkent600: we poll in order to keep the value up-to-date

src/components/Redemptions/RedemptionsMenu.tsx
dkent600: the question is out of scope of this PR, but the answer is we poll in order to keep the value up-to-date

src/components/Proposal/ProposalData.tsx
orenyodfat: I think it can avoided here.
When and if the use press the redeem button it should read the balance.

dkent600: > I think it can avoided here.
> When and if the use press the redeem button it should read the balance.

That is not within the scope of this PR

orenyodfat:
handled by #1987

PR #1975, opened 2020-07-16
fix computation of GEN available for staking

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1969https://github.com/daostack/alchemy/issues/1969

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -390,7 +390,6 @@ export function getGpRewards(reward: IRewardState, daoBalances: { [key: string]:
    * note the following assume that the GenesisProtocol is using GEN for staking
    */
   if (reward.tokensForStaker.gt(new BN(0))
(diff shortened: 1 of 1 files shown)


dkent600:
> The `redeem` popup are now enable though pressing the `redeem proposal` button doing noting :(

I don't understand why we should not see the redeem popup when pressing the redeem button.

orenyodfat:
> > The `redeem` popup are now enable though pressing the `redeem proposal` button doing noting :(
> 
> I don't understand why we should not see the redeem popup when pressing the redeem button. What do you mean "doing nothing"?

nothing happened if pressing the `redeem proposal` button (it should open metamask)

dkent600:
> nothing happened if pressing the redeem proposal button (it should open metamask)

I will investigate whether this behavior is related to this PR, but not until Monday.

PR #1977, opened 2020-07-17
expose alchemy version number

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1900https://github.com/daostack/alchemy/issues/1900

* puts the version number in a new tooltip over Alchemy icon in the header
* removes the existing TrainingTooltip that was obsolete and incorrect
* obtains the version number using https from github in the package.json of the master-2 branch
  * note that unique patch versions might not be reflected in the master branch
  * if desired, this could be made to pull the version from dev-2 for staging/review apps
* adds instructions on using the version in the github bug template
* should we do this for Alchemy 1.0 too?  @orenyodfat

### DIFF
--- a/src/layouts/Header.tsx
+++ b/src/layouts/Header.tsx
@@ -24,7 +24,6 @@ import { Address, IDAOState } from "@daostack/arc.js";
 import { ETHDENVER_OPTIMIZATION } from "../settings";
 import * as css from "./App.scss";
 import ProviderConfigButton from "layouts/ProviderConfigButton";
-import axios from "axios";
 import Tooltip from "rc-tooltip";
 
 interface IExternalProps extends RouteComponentProps<any> {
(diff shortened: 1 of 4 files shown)


src/layouts/Header.tsx
```diff
@@ -108,6 +117,8 @@ class Header extends React.Component<IProps, null> {
         this.props.disableTrainingTooltipsShowAll();
       };
     }
+
+    this.setState({ alchemyVersion: (await axios("https://raw.githubusercontent.com/daostack/alchemy/master-2/package.json"))?.data?.version ?? "Not found" });
```
orenyodfat, line 121: lets avoid adding another dependency (GitHub) .  
take the version during build time from package.json.

dkent600: Done

Issue #1595, opened 2020-04-15
Proposal form should save progress


EzraWeller:
We've gotten a number of stories from people frustrated that they lost their progress on a proposal when they navigated away from the proposal form for whatever reason. Can we come up with a minimal way to save progress via local storage?

dkent600:
Done, in 2.0

PR #1979, opened 2020-07-17
conserve modal form entries

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1595

* stores form values in sessionStorage (so they are scoped to the browser session)
* values are stored when the form closes for any reason, including Submit
* creates a service for use by all the proposal/submission-create modal popup components:  `FormModalService.tsx`.  It is designed to work with any modal component that has a single form and they want to persist and hydrate the form's values in sessionStorage, hydrate the form from url querystring params, or copy to the clipboard the form values as a url


roienatan:
In a situation when a user accidentally closes the modal or for some reason the proposal does not pass (lack of funds for example) that's a very good UX to save the params. 
The only thing that I doubt about is when a proposal passes and one wants to create another proposal he sees the old values of the previous proposal - that's a little bit unnatural and this is the only disadvantage here.
The thing is that we can't know beforehand when a proposal passes (for example if the contract reverts) to delete the old values. So it's kind of a conflict here but I still think there are more advantages than disadvantages. 
@orenyodfat What do you say?

orenyodfat:
lets add a `clear` button , which bring it to its default values, at the modal ..for user who want to start from scratch .

src/components/Proposal/Create/PluginForms/ResetFormButton.scss
```diff
@@ -0,0 +1,24 @@
...
+  color: $white;
+
+  &:hover {
+    cursor: pointer;
+  }
+
+  :disabled {
+    background-color: $disabled-button-color;
+    border: $disabled-button-border;
+    color: rgba(75, 75, 75, 1);
```
roienatan, line 22: Let's use global variable.

src/components/Proposal/Create/PluginForms/ResetFormButton.scss
```diff
@@ -0,0 +1,24 @@
...
+  height: 30px;
+  border: none;
+  display: inline-block;
+  margin: 0 5px;
+  font-size: 13px;
+  padding: 0 25px 2px 25px;
+  font-family: $body-font;
+  font-weight: bold;
+  transition: all 0.25s ease;
+  background-color: $green-button-color;
```
roienatan, line 11: I think a red color for a 'clear' button makes more sense, what do you think?
Or maybe transparent fill but with a red border.

src/components/Proposal/Create/PluginForms/ResetFormButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+      onClick={(): void => {
+      /**
+       * timeout for the case where clicking the button causes a focus change
+       * that triggers validation, which causes this operation not to work
+       */
+        setTimeout(() => {
+          props.resetToDefaults();
+        }, 0);
+      }}>
+    Clear entries
```
roienatan, line 20: For new development let's use `i18next`. Just add this text to `src/assets/locales/en/translation.json` and use it as the rest of the texts there. Please let me know if something is not understood there.

src/components/Proposal/Create/PluginForms/ResetFormButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+import * as css from "./ResetFormButton.scss";
+
+export default (props: {
+  resetToDefaults: () => void,
+  isSubmitting: boolean,
+}): JSX.Element => {
+  return (
+    <button className={css.resetFormButton} type="button" disabled={props.isSubmitting}
+    // eslint-disable-next-line react/jsx-no-bind
+      onClick={(): void => {
```
roienatan, line 11: Is it possible to call this function not inline? If you can declare it in the class scope and just call it. Better for readability and no need to disable eslint manually.

src/components/Proposal/Create/PluginForms/CreateKnownGenericPluginProposal.tsx
```diff
@@ -64,23 +67,68 @@ interface IState {
...
+        case "string":
+          defaultValues[field.name] = "";
+          break;
+        case "bool":
+          defaultValues[field.name] = 0;
+          break;
+        case "address[]":
+          defaultValues[field.name] = [""];
+          break;
+      }
```
roienatan, line 105: Is it not complaining of missing `default` keyword?

dkent600:
@roienatan Thanks for the comments.  I forgot that this was not longer a draft.  I am working on a bug that I found in the UsreSearchField.  It is causing the input field to not work in all circumstances and tests to fail. In other words, the Clear button functionality you requested is not yet complete.  So please hold off...I'm setting this back to draft.

src/components/Proposal/Create/PluginForms/ResetFormButton.scss
dkent600: Note this is not new code for this PR, and I wasn't planning to spend more time refactoring that css (is arguable whether I should have touched it at all), but I'll go ahead and make this change anyway.

src/components/Proposal/Create/PluginForms/ResetFormButton.scss
dkent600: Why red?

src/components/Proposal/Create/PluginForms/CreateKnownGenericPluginProposal.tsx
dkent600: It is not complaining.  This is old code that I would hesitate to mess with.

dkent600: This line looks suspect to me:

```
defaultValues[field.name] = [""]
```

seems like should more likely be:

```
defaultValues[field.name] = []
```

src/components/Proposal/Create/PluginForms/ResetFormButton.tsx
dkent600: Done

src/components/Proposal/Create/PluginForms/ResetFormButton.tsx
dkent600: done

src/components/Proposal/Create/PluginForms/ResetFormButton.scss
dkent600: done

src/components/Proposal/Create/PluginForms/ResetFormButton.scss
roienatan: Usually red color refers to delete/clear/reset while green color refers to add/execute.

dkent600:
> lets add a `clear` button , which bring it to its default values, at the modal ..for user who want to start from scratch .

Done

src/components/Proposal/Create/PluginForms/ResetFormButton.scss
dkent600: Done:

image

package.json
```diff
@@ -114,7 +114,7 @@
     "query-string": "^6.2.0",
     "rc-tooltip": "^3.7.3",
     "react": "^16.8.0",
-    "react-autosuggest": "^9.4.3",
+    "react-autosuggest": "^10.0.2",
```
roienatan, line 117: Is it related?

dkent600: This was because the control was not working right and I wanted to ensure the problem would not be resolved by upgrading (it wasn't).  Since in general we support keeping our packages up-to-date, I left this in.

PR #1994, opened 2020-07-21
Update daocreator package 1.0

cbrzn:
Daocreator 1.0 stack redirects to base url


orenyodfat:
@cbrzn  is that for 1.0 ?

cbrzn:
@orenyodfat that's correct sir

Issue #1992, opened 2020-07-21
Link to competition should appear without having to refresh the page


dkent600:
Here: http://localhost:3000/dao/0xba09811d02631d1b03c21d1369cfcba42348738e/proposal/0xa665681686d402832d19d962ce821d723b8cdac9ef40e36f6158bd34d9b59b85

Upon normal entry to the page from the plugin proposals page, no link:

![image](https://user-images.githubusercontent.com/1821666/88047231-4f1dd000-cb1f-11ea-8b7e-03a853b9145f.png)

After refreshing the page the link is correctly visible (ignore the weird rewarder contract name alias):

![image](https://user-images.githubusercontent.com/1821666/88047271-5e9d1900-cb1f-11ea-9ebf-4acf655a4e34.png)

dkent600:
Merged

PR #1997, opened 2020-07-23
fix Competion link visibility on proposal details

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1992


dkent600:
> For readability, please add a space between the plugin name and the address.

This is an alias.  We take all aliases as-is, unaltered.

Issue #1993, opened 2020-07-21
2.0: Cannot redeem !


orenyodfat:
see https://github.com/daostack/alchemy/pull/1975#pullrequestreview-451040891

- maybe only for none contributionReward proposal. (maybe not)

roienatan:
There are two unrelated problems to solve this issue:
- Analytics fails.Need to check that `contributionRewards` defined.
- There is no redemption function in arc.js for SchemeFactory and SchemeRegistrar.

This issue depends on this https://github.com/daostack/arc.js/issues/516

PR #1998, opened 2020-07-23
Fix Analytics failure ; Fix redeem issue

roienatan:
Resolves: https://github.com/daostack/alchemy/issues/1993

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
   "dependencies": {
     "3box": "1.17.1",
     "@burner-wallet/burner-connect-provider": "^0.1.1",
-    "@daostack/arc.js": "^2.0.0-experimental.48",
+    "@daostack/arc.js": "2.0.0-experimental.48",
     "@dorgtech/daocreator-ui-experimental": "1.1.10",
     "@fortawesome/fontawesome-svg-core": "^1.2.10",
     "@fortawesome/free-brands-svg-icons": "^5.6.1",
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
   "dependencies": {
     "3box": "1.17.1",
     "@burner-wallet/burner-connect-provider": "^0.1.1",
-    "@daostack/arc.js": "^2.0.0-experimental.47",
+    "@daostack/arc.js": "^2.0.0-experimental.48",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -88,6 +88,7 @@
     "@sentry/browser": "^5.0.8",
     "@toruslabs/torus-embed": "^1.3.0",
     "@walletconnect/web3-provider": "^1.0.0-beta.37",
+    "apollo-link-retry": "^2.2.16",
```
dkent600, line 91: Why add this?

roienatan: Actually it's not necessary. For some reason when targeting arc.js locally to Alchemy (in order to test) I was asked to install that dependency, otherwise it won't compile. I forgot to remove this, will do.

src/components/Proposal/ActionButton.tsx
```diff
@@ -309,20 +309,22 @@ class ActionButton extends React.Component<IProps, IState> {
...
 
-    Analytics.track("Redeem", {
+    Analytics.track("Redeem", Object.assign({
       "DAO Address": daoState.address,
       "DAO Name": daoState.name,
       "Proposal Hash": proposal.coreState.id,
       "Proposal Title": proposal.coreState.title,
       "Plugin Address": proposal.coreState.plugin.entity.coreState.address,
       "Plugin Name": proposal.coreState.plugin.entity.coreState.name,
+      "GEN for staking": fromWei((gpRewards.daoBountyForStaker as BN).add(gpRewards.tokensForStaker)),
```
orenyodfat, line 319: why this moved ?

roienatan: No special reason. I think it's just more readable when the condition is in the end of the Object.assign.

package.json
```diff
@@ -79,7 +79,7 @@
   "dependencies": {
     "3box": "1.17.1",
     "@burner-wallet/burner-connect-provider": "^0.1.1",
-    "@daostack/arc.js": "^2.0.0-experimental.47",
+    "@daostack/arc.js": "^2.0.0-experimental.48",
```
orenyodfat, line 82: please remove ^

PR #2000, opened 2020-07-23
remove obsolete work-around

dkent600:


dkent600:
> Can you please explain why the old code is not relevant? Why don't we need the filter?

Sorry, I didn't explain that in the description because it is given in the deleted code comments:

image

.

Issue #1984, opened 2020-07-20
proposal countdown should display properly


dkent600:
See this competition proposal, here:  https://alchemy-2-rinkeby.herokuapp.com/dao/0xba09811d02631d1b03c21d1369cfcba42348738e/plugin/0xe2a368db127496077fa594ed069256373b1801413dcdb5c882c2b29a7b6d6d7d

![image](https://user-images.githubusercontent.com/1821666/87934942-1b2ea600-ca5e-11ea-8bdf-6b88199cdc6f.png)

![image](https://user-images.githubusercontent.com/1821666/87935205-a5770a00-ca5e-11ea-9512-de08b6d03dbf.png)

dkent600:
This is a problem with arc.js:  https://github.com/daostack/arc.js/issues/518

dkent600:
Merged

PR #2002, opened 2020-07-24
handle all arc.js datetime types

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1984

* Adds `safeMoment`.
* a couple other tweaks to attempt to ensure we're safe from any changes to arc.js wrt datetime typing
  * where Competition computes its status with respect to its workflow
  * where the vote buttons compute whether the proposal was created before the account had rep in the DAO

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -652,6 +652,7 @@ export function safeMoment(dateSpecifier: moment.Moment | Date | number | string
     case "string":
       return moment(dateSpecifier);
     case "number":
+      // then should be a count of seconds in UNIX epoch
       return moment.unix(dateSpecifier);
     default:
       throw new Error(`safeMoment: unknown type: ${typeof dateSpecifier}`);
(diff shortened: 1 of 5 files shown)


dkent600:
> So if we solve this issue from arc.js this will not be necessary?

I'm not sure at the moment, would need to research it a bit.  May depend on which data type is chosen.  May not hurt to keep the change, just to be robust.

This line in particular should be checked:

image

src/lib/util.ts
```diff
@@ -301,7 +301,7 @@ export async function getNetworkId(web3Provider?: Web3Provider): Promise<string>
       return network.chainId.toString();
     } else {
       const promise = new Promise<string>((resolve, reject) => {
-        web3Provider.send("net_version", (error, response) => {
+        web3Provider.send("net_version", (error: any, response: any) => {
```
roienatan, line 304: Is it related?

dkent600: it is just the usual typescript lint warning improvements, an ongoing improvement...does not affect execution of the code.

roienatan: ok, yes I assume also the `boolean` in `proposalHelpers.ts`

Issue #1978, opened 2020-07-17
should complain when connecting to the wrong chain


dkent600:
STEPS:

1. in Metamask, connect to the wrong network
2. click Connect in the app

RESULT:

there is no notification that the app has connected to the wrong network

EXPECT:

a notification

I've seen this in 2.0 for sure.  Not sure about 1.0.

dkent600:
Merged

PR #2005, opened 2020-07-25
fix computation of networkId

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1978

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -280,19 +280,12 @@ export function sleep(milliseconds: number): Promise<void> {
  * @param web3Provider
  */
 export async function getNetworkId(web3Provider?: Web3Provider): Promise<string> {
-  let arc: Arc;
+  if (web3Provider) {
 
-  try {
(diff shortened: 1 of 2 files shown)


src/lib/util.ts
```diff
@@ -312,11 +305,24 @@ export async function getNetworkId(web3Provider?: Web3Provider): Promise<string>
...
-    const web3 = (window as any).web3;
-    return (await (web3.eth.net ? web3.eth.net.getId() : promisify(web3.version.getNetwork)())).toString();
   } else {
-    throw new Error("getNetworkId: unable to find web3");
+    let arc: Arc;
+
+    try {
+      arc = getArc();
+    } catch (ex) {
+      // Do nothing
```
roienatan, line 314: Why not `throw` or `console.log`?

dkent600: It is a normal occurrence that there may be no arc.  Just don't care.  We check for the case of no arc in the next line.

Issue #2003, opened 2020-07-24
can't connect to fortmatic


dkent600:
Only tried on rinkeby.  Trying to get the networkId in `util.getNetworkId(web3Provider)`:

`web3Provider.send("net_version", (error, response)...`

on the provider as returned by Web3Modal.  But I'm getting the following error, and, being unhandled, the promise is not invoking the reject callback.

```
Uncaught DOMException: Failed to execute 'postMessage' on 'Window': (error, response) => {
                        if (error) {
                            reject(error);
     ...<omitted>... } could not be cloned.
```

Result is that no error is reported by Alchemy, and can't connect to fortmatic.

dkent600:
Hmm, same with Portmatic.  May be our own error.

dkent600:
OK, Portmatic, if it works at all only works on mainnet where 2.0 currently has no presence.

I have tested with Metamask, Formatic, Torus and WalletConnect (Gnosis Safe)

dkent600:
Fixed in 2.0

PR #2011, opened 2020-07-27
fix non-MM connections, upgrade provider packages

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/2003

* changes how we obtain the chainId for all non-Metamask providers
* upgrades all supported provider packages
* removes webpack from dependencies and bumps its version in devDependencies (addresses travis build issue)


dkent600:
OK, Portmatic, if it works at all only works on mainnet where 2.0 currently has no presence.

I have tested with Metamask, Formatic, Torus and WalletConnect (Gnosis Safe)

PR #2018, opened 2020-07-29
refactor and fix issues relating 3box operations

dkent600:
* Refactors 3box in prep for adding threads.
* Addresses issues such as when switching metamask accounts.  Primary improvement is probably to cast account address to lower case to enable account comparisons and accounts used as keys, to resolve properly.
* adds typings definitions to help self-document the hard-to-understand code in profilesReducer
* refactors some of profilesReducer to clarify and robustify the code
* modifies AccountProfile UI logic to only go into edit mode once the account is fully connected and we are sure that the user can edit the profile.  Particularly at issue is when the metamask and alchemy-cached accounts differ.  Before allowing the user to edit. must be connected to metamask to assure this is not the case.
* handles case where logging out and there is a big console error from our 3Box code and who knows what else happens when the `currentAccountAddress` goes to null.
* not fixed: can have an issue [described here](https://github.com/daostack/alchemy/issues/2059)
* does not use the latest 3Box [due to this](https://github.com/3box/3box-js/issues/891), but does use the next prior version


dkent600:
Waiting for this from 3Box:  https://github.com/3box/3box-js/pull/874

dkent600:
Waiting for this from 3Box@1.21.0:  https://github.com/3box/3box-js/issues/889

dkent600:
Blocked on this: https://github.com/3box/3box-js/issues/890

src/actions/profilesActions.ts
```diff
@@ -155,71 +244,56 @@ export function updateProfile(accountAddress: string, name: string, description:
...
+        payload: { type, id, isFollowing },
+      } as FollowItemAction);
 
-    if (follows[type].includes(id)) {
-      follows[type] = arrayRemove(follows[type], id);
-      isFollowing = false;
-    } else {
-      follows[type].push(id);
+    } catch (e) {
+      dispatch(showNotification(NotificationStatus.Failure, `Failed to connect to 3box: ${e.message}`));
```
orenyodfat, line 292: is this can be moved to the string place ?

src/assets/locales/en/translation.json
```diff
@@ -28,5 +28,9 @@
     "ETH Balance": "ETH Balance",
     "Members": "Members",
     "Plugin activation time": "The plugin will be activated at",
+    "3BoxLoginSuccess": "Logged in to 3Box",
+    "3BoxProfileSuccess": "Profile data saved to 3Box",
+    "Following": "Now following",
+    "UnFollowing": "No longer following"
```
orenyodfat, line 34: `'` missing

dkent600: Fixed.  Dealing with constant merge conflicts in this file is becoming a pain.  Wondering if we can figure out a way of editing this file that will minimize that.

src/actions/profilesActions.ts
dkent600: That is old code.  I have not tasked myself in this PR with moving every single string into i18.

Issue #2030, opened 2020-08-02
Simple UX/UI improvments


roienatan:
- Modal size to be responsive to content (eliminate white void)
- Loading spinner color to match Alchemy Earth styles
- Thicker scroll bar ; lighter color on hover
- Search input no border
- Thinner DAOs cards (3 in a row)
- Create DAO button and email alerts to be in the same vertical line of the DAOs cards
- A space is added between proposal counters for better readability

PR #2031, opened 2020-08-02
Simple UX/UI improvments

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2030

- Modal size to be responsive to content (eliminate white void)
- Loading spinner color to match Alchemy Earth styles
- Thicker scroll bar ; lighter color on hover
- Search input no border
- Thinner DAOs cards (3 in a row)
- Create DAO button and email alerts to be in the same vertical line of the DAOs cards
- A space is added between proposal counters for better readability


src/components/Dao/ProposalPluginCard.tsx
```diff
@@ -54,7 +54,7 @@ const ProposalPluginCard = (props: IProps) => {
           </TrainingTooltip> : headerHtml
         }
         <div>
-          <b>{pluginState.numberOfBoostedProposals}</b> <span>Boosted</span> <b>{pluginState.numberOfPreBoostedProposals}</b> <span>Pending Boosting</span> <b>{pluginState.numberOfQueuedProposals}</b> <span>Regular</span>
+          <b>{pluginState.numberOfBoostedProposals}</b> <span>Boosted   </span> <b>{pluginState.numberOfPreBoostedProposals}</b> <span>Pending Boosting   </span> <b>{pluginState.numberOfQueuedProposals}</b> <span>Regular</span>
```
dkent600, line 57: good opportunity to replace the `<b>` with css

src/components/Proposal/Create/CreateProposal.scss
```diff
@@ -388,7 +387,7 @@
   content: "";
   clear: both;
   overflow: auto;
-  height: calc(100% - 40px);
+  max-height: calc(100vh - 200px);
```
dkent600, line 390: how does 200px look on mobile?  Would a % be better?

src/layouts/App.scss
```diff
@@ -36,13 +36,13 @@ button:focus {
 
 /* Handle */
 ::-webkit-scrollbar-thumb {
-  background: #888;
+  background: #555;
```
dkent600, line 39: scss color variable?

src/components/Dao/ProposalPluginCard.tsx
roienatan: But still it'll be necessary to wrap it `<div>` or something since only the number is bold.

dkent600: of course

src/components/Proposal/Create/CreateProposal.scss
roienatan: I agree that generally `%` is preferable but in this case, probably because the way it's built, it's not look like expected with `%`. Regarding the mobile it looks good, I've done a small modification
now.

dkent600: > it's not look like expected with %

I don't understand what you're saying

roienatan: It just does not works good with `%`, that's the reason I use `vh` there.

Issue #2044, opened 2020-08-06
Disable "+ New Proposal" button when a plugin is not active yet


roienatan:
Currently when a plugin is not active nothing happens when clicking on "+ New Proposal" button. This is a bad UX - should disable the button and/or give a message: "The plugin is not active yet".

dkent600:
My preference in these situations is to give the message.  Or second, if disabled, provide a tooltip explaining why.

roienatan:
> My preference in these situations is to give the message. Or second, if disabled, provide a tooltip explaining why.

That's what I thought :)

PR #2045, opened 2020-08-06
New proposal button is now disabled when a plugin is not activated yet

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2044

When training tooltip is enables it looks like this now:
Screen Shot 2020-08-06 at 13 57 06


orenyodfat:
> I would do a regular tooltip, so it is always available to the user (or best, a message right on the page). This isn't really a question of training. It is a context-sensitive state that needs to be more proactively explained to the user when it happens.

the tooltip is annoying . we might remove that at all.

roienatan:
> > I would do a regular tooltip, so it is always available to the user (or best, a message right on the page). This isn't really a question of training. It is a context-sensitive state that needs to be more proactively explained to the user when it happens.
> 
> the tooltip is annoying . we might remove that at all.

@orenyodfat  @dkent600 What do you guys think about a regular tooltip (like the one you have when you hover above Alchemy logo)? Or just adding the text from the left of the button.

roienatan:
Please let me know what you guys think:
Screen Shot 2020-08-06 at 16 16 53

dkent600:
> > I would do a regular tooltip, so it is always available to the user (or best, a message right on the page). This isn't really a question of training. It is a context-sensitive state that needs to be more proactively explained to the user when it happens.
> 
> the tooltip is annoying . we might remove that at all.

I would only be there when the button is disabled.

Issue #2065, opened 2020-08-14
no symbol of DXD at Funding and Voting Power proposal


orenyodfat:
https://alchemy.daostack.io/dao/0x519b70055af55a007110b4ff99b0ea33071c720a/proposal/0x92b57aff2571091eefd616ca185f2136232060bf15fefb29611290127eb7bc93

<img width="783" alt="Screen Shot 2020-08-14 at 13 59 14" src="https://user-images.githubusercontent.com/5994398/90242824-60b07a00-de36-11ea-94b3-3b7f9726b6a5.png">

there is a `?` instead of DXD

dkent600:
Not repro.  I see the proposal:

![image](https://user-images.githubusercontent.com/1821666/90243328-b5370380-ddfc-11ea-8faf-6b7a0c1e5404.png)

orenyodfat:
" could we fix this and additionall add a hover on the Symbol that displays the token address on external token requests? "

dkent600:
fixed by #2068

PR #2068, opened 2020-08-14
Fix dxd token display

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/2065

Same fix applies to NEC

I'm not wild about this because it doesn't address the fact that someone has to remember/document that these tokens must always be submitted in lowercase.

### DIFF
--- a/data/tokens.json
+++ b/data/tokens.json
@@ -48,7 +48,7 @@
         "name": "Tether USD",
         "symbol": "USDT"
       },
-      "0xCc80C051057B774cD75067Dc48f8987C4Eb97A5e": {
+      "0xcc80c051057b774cd75067dc48f8987c4eb97a5e": {
         "decimals": 18,
         "name": "Nectar",
         "symbol": "NEC"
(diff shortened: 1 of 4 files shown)


data/tokens.json
```diff
@@ -48,7 +48,7 @@
         "name": "Tether USD",
         "symbol": "USDT"
       },
-      "0xCc80C051057B774cD75067Dc48f8987C4Eb97A5e": {
+      "0xcc80c051057b774cd75067dc48f8987c4eb97a5e": {
```
orenyodfat, line 51: why not alway do `toLowerCase` when reading that ?

orenyodfat:
here https://github.com/daostack/alchemy/blob/d1b660ab931e99eaa3b3607497bc206b138274c3/src/lib/util.ts#L235 
and 
https://github.com/daostack/alchemy/blob/d1b660ab931e99eaa3b3607497bc206b138274c3/src/lib/util.ts#L239
and
https://github.com/daostack/alchemy/blob/d1b660ab931e99eaa3b3607497bc206b138274c3/src/lib/util.ts#L244
it is shift to lowercase so where is the issue ?

data/tokens.json
dkent600: believe me I tried to do something like that, but those are property _names_, not values.  So requires that whole data structure change.

Issue #2038, opened 2020-08-03
Plugin Manager Create Proposal actions menu should have a right margin


dkent600:
![image](https://user-images.githubusercontent.com/1821666/89223590-0b0ddf00-d5a5-11ea-96a3-df5091f39c16.png)

PR #2071, opened 2020-08-15
Plugin manager modal UI improvments

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2038

The only change that was made is wrapping the `<div className={pluginManagerFormClass}></div>` with `<div className={css.contentWrapper}></div>`, then I had to indent everything inside, that's way GitHub shows so many differences.


roienatan:
> travis

Updated arc.js, test env and subgraph.

Issue #2087, opened 2020-08-18
Remove required field in NFT scheme


arsena21:
See this thread: https://daostack.slack.com/archives/GUM9N26HX/p1597269328002500

It should be possible to leave the "Data" field in the proposal empty. 

This is for the NFT Management scheme for arc2.0

PR #2091, opened 2020-08-30
added default value and placeholder

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2087

@arsena21


src/genericPluginRegistry/plugins/NFTManager.json
```diff
@@ -30,7 +30,8 @@
           {
             "label": "Data",
             "name": "data",
-            "placeholder": ""
+            "placeholder": "Arbitrary data to transfer (0x0)",
+            "defaultValue": "0x0"
```
dkent600, line 34: It would be better if the default value were "" instead of "0x0"

roienatan: As discussed with Oren - decided to leave like this.

orenyodfat (approved):

src/components/Proposal/Create/PluginForms/CreateKnownGenericPluginProposal.tsx
```diff
@@ -380,7 +383,7 @@ class CreateKnownPluginProposal extends React.Component<IProps, IState> {
                 }
 
                 if (field.type.includes("bytes")) {
-                  if (!isHexString(value)) {
+                  if (value && !isHexString(value)) {
```
dkent600, line 386: I think it may be wrong to expect this to be a hex string.  the data type is bytes.  I seem to recall the contract saying it ben be any data.  Are you sure this cannot in fact be *any* string value?  This is what you are implying in the placeholder text.

src/genericPluginRegistry/plugins/NFTManager.json
```diff
@@ -30,7 +30,9 @@
           {
             "label": "Data",
             "name": "data",
-            "placeholder": ""
+            "placeholder": "Arbitrary data to transfer (0x0)",
```
dkent600, line 33: Can this be:  "Arbitrary data to transfer (0x0 means no data)"

src/components/Proposal/Create/PluginForms/CreateKnownGenericPluginProposal.tsx
orenyodfat: it is hex string

PR #2093, opened 2020-08-31
Updated chrome driver for tests

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2092

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,6 @@ before_install:
 addons:
   chrome: stable
 
-before_script:
-  - LATEST_CHROMEDRIVER_VERSION=`curl -s "https://chromedriver.storage.googleapis.com/LATEST_RELEASE"`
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -14,6 +14,11 @@ before_install:
 addons:
   chrome: stable
 
+before_script:
+  - LATEST_CHROMEDRIVER_VERSION=`curl -s "https://chromedriver.storage.googleapis.com/LATEST_RELEASE"`
```
dkent600, line 18: Where is LATEST_CHROMEDRIVER_VERSION used?

orenyodfat (approved):

.travis.yml
roienatan: In the next line :)

dkent600: How does it sync with the changes in wdio.conf.js?

roienatan: For now, I removed it since I'm not sure if it works properly. Will check it next time.
Thanks!

Issue #2057, opened 2020-08-12
Voting should automatically update UI


dkent600:
This is an example of a proposal card on which I have voted and the vote was mined, but my vote is not reflected in the GUI and won't be until I manually refresh the screen.  For what it is worth, it is sitting in the Pending queue, but I don't know if that makes a difference.

![image](https://user-images.githubusercontent.com/1821666/90016460-04076080-dc78-11ea-88c2-d3f238f06966.png)

PR #2102, opened 2020-09-04
Automatically refresh voting status

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2057 for 2.0


dkent600 (approved):
I assume you confirmed this works in rinkeby  ;-)

roienatan:
> I assume you confirmed this works in rinkeby ;-)

After investigating, it's solves the issue in ganache but not in Rinkeby.
Will be investigating the issue in Rinkeby in separate PR.

Issue #190, opened 2018-10-26
SearchBar 


GreyEcologist:
a way for users to search proposals by their title

user story:
- user goes to the alchemy page of a DAO
- a search bar is available on the top part of the page
- a placeholder "search a proposal" text 
- user enters search words 
- user clicks a "search" button to the right on the search bar
- relevant proposals from both active and history are returned

jellegerbrandy:
Clsoing this due to inactivity

Theylon:
Hey @orenyodfat @roienatan @EzraWeller This has been brought up by DXdao as a potential useful feature. 
It's a low hanging fruit, can this be prioritized?

dkent600:
This is unclear as to on which page the search filter should appear.  I think it makes most sense, and is most easily coded, to put it on the proposal history page.   Importantly:  Proposals on that page span all schemes.

However, the request here is asking to include in the search results proposals that are still "active".  The history page currently includes `IProposalStage.ExpiredInQueue, IProposalStage.Executed, IProposalStage.Queued`.  I have no idea why it doesn't display all stages.

Should we make the History page display all proposals instead of filtering out several stages as it does?

@orenyodfat @EzraWeller @Theylon

dkent600:
Reopening for the 2.0 PR

Issue #2103, opened 2020-09-04
make DAOs search input execute on Enter and blur


dkent600:
Currently the DAOs search input executes a DAOs query and re-render of the page on every keypress when the input value is longer than two characters.  This is often going to cause a lot of unnecessary queries and renders.

So I would like to propose that we do one of two things:

* execute on enter and blur (loss of focus), or
* add a search button to the right of the input field, and an 'x' in the input field to clear it

PR #2104, opened 2020-09-04
add proposal filter feature (1.0)

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/190 for 1.0

Currently:

- Implements the feature on the DAO History page
- Changes the DAO History page to show _all_ proposals, not just IProposalStage.ExpiredInQueue, IProposalStage.Executed, IProposalStage.Queued and closingAt_lte: now
- adds DaoHistoryPage.scss
- consolidates the proposals query into one place (was in two)
- Behavior of search input is different from the DAOs search.  Executes on Enter and blur (see https://github.com/daostack/alchemy/issues/2103)
- is entirely case-sensitive
- does not search Description field.  Subgraph's GraphQL does not support 'or' across fields in a query.  This this would even require executing two queries, or would require that the user select in the UI what field they want to search by.  I would prefer to do that as a separate ticket and PR.  Same for searching through tags.  @Theylon
- there is a spinny icon that shows on long queries, but not when clearing the filter.

Heads up:
- when the filtered list is showing, the values will not automatically update on proposal state changes.  The DAOs list filter has the same behavior.


package.json
```diff
@@ -68,8 +68,8 @@
     "start-prod": "cross-env NODE_ENV=production SHOW_ALL_DAOS=false NETWORK=main node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js  --config webpack.dev.config.js",
     "start-staging-kovan": "cross-env NODE_ENV=production SHOW_ALL_DAOS=true NETWORK=kovan node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --config webpack.dev.config.js",
     "start-staging-rinkeby": "cross-env NODE_ENV=production SHOW_ALL_DAOS=true NETWORK=rinkeby node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --config webpack.dev.config.js",
+    "start-staging-xdai": "cross-env NODE_ENV=production SHOW_ALL_DAOS=true NETWORK=xdai node --max_old_space_size=4096 ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --config webpack.dev.config.js",
```
orenyodfat, line 71: is this related to this pr ?

dkent600: no, just a little harmless correction

src/components/Dao/DaoHistoryPage.tsx
```diff
@@ -34,24 +82,41 @@ class DaoHistoryPage extends React.Component<IProps, null> {
     });
   }
 
+  onSearchChange = async (e: any) => {
```
roienatan, line 85: Is there a reason why this function is `async`?

roienatan:
Is there a reason that the search is case-sensitive? Usually it's more convenient that it's not case-sensitive.

src/components/Dao/DaoHistoryPage.tsx
dkent600: This is an artifact of a previous change, but since that has been deleted, we can remove the async.

dkent600:
> Is there a reason that the search is case-sensitive? Usually it's more convenient that it's not case-sensitive.

Yes, thegraph doesn't support case insensitive searches.

roienatan:
There are some indentations errors by Travis, please fix.

dkent600:
> There are some indentations errors by Travis, please fix.

This is tricky because for some reason vscode insists on indenting in this way that eslint is rejecting.  I need to figure out why, so won't won't get to this until Monday.  Normally vscode matches eslint behavior perfectly.

orenyodfat (changes requested):
@dkent600 there a lint issues

roienatan (approved):

orenyodfat (approved):

Issue #2101, opened 2020-09-04
DAO search input should not look weird when selecting autocomplete value


dkent600:
This is in a chromium browser:

1. go to the DAOs list page
1. type into the search input something that causes the autocomplete dropdown to appear
1. if there are more than one items in the dropdown, arrow down to select one.

RESULT:
![image](https://user-images.githubusercontent.com/1821666/92241846-e56a4300-ee8c-11ea-9e57-529534ed47b8.png)

EXPECT:
![image](https://user-images.githubusercontent.com/1821666/92241958-121e5a80-ee8d-11ea-8575-138d58f7a54a.png)

dkent600:
Works for me (coding a proposal search input in 1.0):

```
  .searchBox {
    margin-bottom: 20px;
    width: 425px;
    border-radius: 20px;
    height: 30px;
    background: url("/assets/images/Icon/search.svg") $white no-repeat left 6px center;
    background-size: 18px;
    padding-left: 30px;

    input {
      border: none;
      font-family: "AccordBold", sans-serif;
      height: 30px;
      width: 406px;
      font-size: 15px;
      color: #4f6176;
    }

    input:focus {
      outline: none;
    }

    button {
      height: 26px;
      display: inline-block;
      position: relative;
      top: 2px;
      border: none;
      outline: none;
      background-color: transparent;

      &:hover {
        cursor: pointer;
      }
    }
  }
```

PR #2127, opened 2020-09-12
Fix bad ui when using autocomplete input in chrome

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2101

### DIFF
--- a/src/layouts/App.scss
+++ b/src/layouts/App.scss
@@ -24,6 +24,10 @@ button:focus {
   outline: none;
 }
 
+input:-webkit-autofill {
+  transition: all 5000s ease-in-out 0s;
+}
+
 /* width */
 ::-webkit-scrollbar {
   width: 10px;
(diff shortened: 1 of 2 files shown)


src/layouts/App.scss
```diff
@@ -24,6 +24,10 @@ button:focus {
   outline: none;
 }
 
+input:-webkit-autofill {
```
dkent600, line 27: How will making this effectively global affect other inputs where we don't to alter the behavior?

dkent600 (changes requested):

src/layouts/App.scss
roienatan: If we decide to disable the autocomplete background there I it's suppose to be across all Alchemy consistently.

dkent600: No I don't think so.  The other cases across Alchemy do not have the custom background that the Search inputs have.  Thus I suggest that the default behavior is desirable in those cases, no reason to change it.

dkent600 (approved):

Issue #2126, opened 2020-09-11
Subgraph subscriptions in general do not appear to be working in either 1.0 or 2.0


dkent600:
Subgraph subscriptions in general do not appear to be working in either 1.0 or 2.0.  Voting, staking, changes in reputation, new proposals...nothing is updating.

Important point:  The subscriptions work in ganache, but not in rinkeby or kovan or mainnet.

Watching the websocket messages in the browser debugger indicates that the messages are simply not arriving when they should in rinkeby, but are arriving perfectly in ganache.  This suggests that the problem is not with the code in Alchemy, but something wrong with the subgraph or the connection we are using with subgraph.

dkent600:
@orenyodfat Need your input on this

roienatan:
As discussed with Jannis from The Graph, they having issues with the subscriptions. Using polling instead is recommended  until further notice.
https://github.com/daostack/arc.js/issues/543 is required.

roienatan:
Done for 2.0, Open to 1.0

Issue #2049, opened 2020-08-10
Proposal count is not updated


dkent600:
Rinkeby, when the last pending proposal leaves the pending queue, this often happens:.  The count should be zero and the fetching message should not be showing:

![image](https://user-images.githubusercontent.com/1821666/89811720-0e591b80-db0d-11ea-9bf9-2d9f836fe493.png)

roienatan:
Same issue happens when creating a proposal, you can see it on the UI but the number is updated only after a refresh. After an investigation I realise that `pluginState` prop is not updated automatically hence the component does not re-render and this is causing those issues.

PR #2140, opened 2020-09-17
Update arc.js 56 ; use polling

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2126 - for 2.0
resolves: https://github.com/daostack/alchemy/issues/2049

- **Creating proposals**, **voting** and **redeeming** are reflected in the UI automatically
- **Staking** and **number of Regular Proposals, Pending Boosting and Boosted Proposals** are still not reflected in the UI (need to refresh the page to see)

**UPDATE:** proposals counters are now fixed.
**UPDATE:** staking data is now updated automatically

### DIFF
--- a/src/components/Plugin/ContributionRewardExtRewarders/Competition/utils.ts
+++ b/src/components/Plugin/ContributionRewardExtRewarders/Competition/utils.ts
@@ -198,10 +198,3 @@ export const getSubmissionVoterHasVoted = (submissionId: string, voterAddress: s
   return getSubmissionVotes(submissionId, voterAddress, subscribe)
     .pipe(map((votes: Array<CompetitionVote>) => !!votes.length));
 };
-
-// export const primeCacheForSubmissionsAndVotes = (): Observable<any> => {
-//   return combineLatest(
-//     CompetitionSuggestion.search(getArc(), {}, { polling: true, fetchAllData: true }),
-//     CompetitionVote.search(getArc(), {}, { polling: true, fetchAllData: true })
-//   );
-// };
--- a/src/components/Plugin/ContributionRewardExtRewarders/Competition/utils.ts
+++ b/src/components/Plugin/ContributionRewardExtRewarders/Competition/utils.ts
@@ -201,7 +201,7 @@ export const getSubmissionVoterHasVoted = (submissionId: string, voterAddress: s
 
 // export const primeCacheForSubmissionsAndVotes = (): Observable<any> => {
 //   return combineLatest(
-//     CompetitionSuggestion.search(getArc(), {}, { subscribe: true, fetchAllData: true }),
-//     CompetitionVote.search(getArc(), {}, { subscribe: true, fetchAllData: true })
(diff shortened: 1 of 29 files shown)


roienatan:
@dkent600 The issue with the proposal count (https://github.com/daostack/alchemy/issues/2049) happens because `pluginState` in not updated. I'm not sure it's related to the polling.

dkent600:
> @dkent600 The issue with the proposal count (#2049) happens because `pluginState` in not updated. I'm not sure it's related to the polling.

If Alchemy is supplying subscribe:true when it requests the proposalState, then the state should be updated by polling.

roienatan:
> > @dkent600 The issue with the proposal count (#2049) happens because `pluginState` in not updated. I'm not sure it's related to the polling.
> 
> If Alchemy is supplying subscribe:true when it requests the proposalState, then the state should be updated by polling.

I say this because I tried to replace **all** `subscribe: true` to `polling: true` and still the counters didn't update. That's why I might think that there is a problem in Alchemy code itself (which for now I didn't find) and that's why I asked you if you recall that those counters were properly working before.

orenyodfat (changes requested):
The following lines need to be updated : 
https://github.com/daostack/alchemy/blob/update-arcjs-56/src/components/Account/AccountProfilePage.tsx#L358

https://github.com/daostack/alchemy/blob/update-arcjs-56/src/components/Plugin/ContributionRewardExtRewarders/Competition/List.tsx#L140

orenyodfat:
> > > @dkent600 The issue with the proposal count (#2049) happens because `pluginState` in not updated. I'm not sure it's related to the polling.
> > 
> > 
> > If Alchemy is supplying subscribe:true when it requests the proposalState, then the state should be updated by polling.
> 
> I say this because I tried to replace **all** `subscribe: true` to `polling: true` and still the counters didn't update. That's why I might think that there is a problem in Alchemy code itself (which for now I didn't find) and that's why I asked you if you recall that those counters were properly working before.

how about updating this as well ? https://github.com/daostack/alchemy/blob/update-arcjs-56/src/components/Dao/ProposalPluginCard.tsx#L103 ?

src/components/Plugin/ContributionRewardExtRewarders/Competition/utils.ts
```diff
@@ -201,7 +201,7 @@ export const getSubmissionVoterHasVoted = (submissionId: string, voterAddress: s
 
 // export const primeCacheForSubmissionsAndVotes = (): Observable<any> => {
 //   return combineLatest(
-//     CompetitionSuggestion.search(getArc(), {}, { subscribe: true, fetchAllData: true }),
-//     CompetitionVote.search(getArc(), {}, { subscribe: true, fetchAllData: true })
+//     CompetitionSuggestion.search(getArc(), {}, { polling: true, fetchAllData: true }),
```
orenyodfat, line 204: lets remove these comments lines ?

orenyodfat (changes requested):
lets set the polling time to 30 seconds and have it on a single global var

Issue #2134, opened 2020-09-13
Competition proposal form takes more than a full screen alchemy-2


Theylon:
The competition proposal form takes more than a full screen on alchemy-2, there is no scrollbar, and the only way to submit the proposal is zooming out in the browser

Try to create a competition on this Kovan DAO: https://alchemy-2-kovan.herokuapp.com/dao/0x2fca504da74b3e0c6699b3f6789fa3bc7e3ecdd5/plugin/0x873b490c0189078089abaf10a0fdc824e72926c51d6b2a8059289999dfc74afa/proposals/create/

![image](https://user-images.githubusercontent.com/27952472/93012920-bea6bd80-f5ac-11ea-8f29-1d06ce1f6839.png)

PR #2141, opened 2020-09-18
Fix no side bar modal scrolling

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2134

### DIFF
--- a/src/components/Proposal/Create/CreateProposal.scss
+++ b/src/components/Proposal/Create/CreateProposal.scss
@@ -381,7 +381,7 @@
 .containerNoSidebar {
   width: 562px;
   overflow-y: auto;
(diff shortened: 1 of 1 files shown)


dkent600 (approved):

dkent600:
I would make sure this is ok on mobile.  Note that instead of `calc(100vh - 200px)` you can do: `80vh` which is more flexible.

roienatan:
> @roienatan  I would make sure this is ok on mobile.  Note that instead of `calc(100vh - 200px)` you can do: `80vh` which is more flexible.

That's the same as the modal with sidebar.

orenyodfat (approved):

Issue #1553, opened 2020-04-01
3box comment threads


orenyodfat:
replace Discuss with 3box msging threads .
https://docs.3box.io/build/web-apps/messaging

"3box could be better because its decentralized and uses the same profile as everywhere in the app"

Re already Discuss msgs ,one option will be to leave a link to Discuss so users can reach out to see old discussions.
@EzraWeller your opinion please.

tibetsprague:
@EzraWeller i can do this one next, but lets make a judgement call on whether to link to old Disqus conversation or just move on as a reset

EzraWeller:
I can think of at least one example where people were using the Disqus on the DAO wall to link to onboarding docs for the DAO and that kind of thing. Losing this would only be annoying, though, and wouldn't actually lose that DAO any actual work.

If it's going to add significant work to keep the Disqus conversation histories, I don't think we should do it.

tibetsprague:
@EzraWeller we could always copy those comments over from Disqus to 3box :)

dkent600:
STEPS:

1. start not connected
1. go to Gustatory Engineering
1. connect
1. click in new comment input field and notice you can see your account's name
1. enter a comment
1. wait
1. see your new comment appear (along with others, which is another bug report)
1. go to Genesis Alpha
1. click in new comment input field

RESULT:
Notice you cannot see your account's name, only the address, even though nothing has changed with respect to the login.

dkent600:
STEPS:

1. start not connected
2. go to Gustatory Engineering

RESULT:  you cannot see your comments.  

3. connect

RESULT:  you still cannot see your comments.  Your won't be able to see any comments unless you try to enter a new one.

dkent600:
Now there is a problem compiling 3Box@1.21.0:  https://github.com/3box/3box-js/issues/889

dkent600:
We are going to start from scratch using 3Box threads.

dkent600:
blocked by: https://github.com/3box/3box-js/issues/891

Theylon:
This feature should be added to arc 1

PR #2146, opened 2020-09-23
3box threads

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/1553

Note 3box does not have out-of-the-3box support for editing or replying to comments.  We can implement those ourselves, but as a separate PR.


roienatan:
1. After submitting a comment the content inside the text area should disappear (need to think about it since as I understood there is no option to edit comments after submitting)
2. I would change the trash color to bright red
3. No need the "Please confirm" tooltip when hovering the trash icon
4. Feels like the Submit button should placed in the right (like the rest of the submit/ok buttons across the project) and maybe green color like the Join Conversation button
5. A grey thin line between each comment would be nice (see attached image)
Screen Shot 2020-10-02 at 15 43 08
6. The orange question mark is unnecessary there. I know it's part of the text component - maybe need to add prop to the text component which we can control whether to show the orange question mark or not.

7. The Yes and No buttons in the delete modal are not styled as the rest of the buttons in Alchemy (I think they have unnecessary border?)
8.  Try placing the comment date under the name - I think it'll look better.

package.json
```diff
@@ -122,6 +123,7 @@
     "i18next": "^19.6.3",
     "immutability-helper": "2.6.4",
     "interweave": "^11.1.0",
+    "levelup": "^4.4.0",
```
roienatan, line 126: What's this is used for?

src/settings.ts
```diff
@@ -3,6 +3,7 @@ export const ETHDENVER_OPTIMIZATION = true;
 // if this is true, we do get the contractInfos from a locally stored file in ./data instead of from the subgraph
 export const USE_CONTRACTINFOS_CACHE = false;
 export const GRAPH_POLL_INTERVAL = 30000;
+export const THREEBOXTHREADSMODERATOR = "0xd50fc49ff389558d23a76Cf246dA147FF53D8Df8";
```
roienatan, line 6: Is it some private key? If so it should be as an ENV_VAR

package.json
dkent600: work-around for a 3box bug.  before submitting this for review (note it is still a draft) I plan to check whether they have fixed the bug

src/settings.ts
dkent600: no, that is the address of the moderator, is meant to be public

orenyodfat:
remove the warning.. there is no need for alchemy 2

dkent600:
> 1. After submitting a comment the content inside the text area should disappear (need to think about it since as I understood there is no option to edit comments after submitting)

For me this is not a big deal either way.  I chose the current behavior because 1) I am never comfortable deleting a user's data unless there is a really good reason, and 2) we don't have a way to edit comments once submitted.  So to change a comment one has to delete and resubmit

> 2. I would change the trash color to bright red

Why?  Red doesn't look very good in this color palette, and is a color that has a tendency to create tension.  Since we prompt to confirm the deletion, I see no need for anything alarming.

> 3. No need the "Please confirm" tooltip when hovering the trash icon

I changed it to "Delete this comment...", which I think is needed so people know what will happen if they click it.  The "..." is standard in computer UX to indicate that a further prompt will follow.

> 4. Feels like the Submit button should placed in the right (like the rest of the submit/ok buttons across the project) and maybe green color like the Join Conversation button

I don't agree with either suggestion, especially the button position.  1) this isn't a modal popup, and the width of the page is significantly great, 2) the user's eyes, typing from left to right, are always going to be biased to viewing the left side, 3) the grey button color matches the grey palette of the markdown component, thus helping to user to comfortably and intuitively parse the markdown component and the button as a functional group amidst the surrounding clutter.

> 5. A grey thin line between each comment would be nice

I thought of that, however, Discus doesn't do it, google search doesn't do it, I feel that there are already plenty of visual cues enabling the user to comfortably parse the page without adding clutter to it.

> 6. The orange question mark is unnecessary there. I guess it's part of the text component - maybe need to add prop to the text component which we can control whether to show the orange question mark or not.

This help button is our standard where ever we have the ability to compose markdown.  I would not remove or change it unless we remove or change it everywhere, and that should be as a separate PR. 

> 7. The Yes and No buttons in the delete modal are not styled as the rest of the buttons in Alchemy (I think they have unnecessary border?)

I only see a problem with the No button, so I removed the border.  Otherwise these buttons appear the same way whenever we have a SimpleMessagePopup (which until now has only been with an OK button).  If we want to improve it really ought to be a separate PR.

> 8.  Try placing the comment date under the name - I think it looks better.

Done:
image

package.json
dkent600: I have a query in with 3box to discover whether this work-around is still necessary

orenyodfat (approved):

src/components/Shared/ThreeBoxThreads.tsx
```diff
@@ -0,0 +1,233 @@
...
+    const daoState = this.props.daoState;
+
+    return (
+      <div className={css.container}>
+        <div className={css.startDiscussionButton}>
+          {!this.state.joinedThread && (this.state.threeboxPosts !== null) ?
+            <Tooltip placement="top" trigger={["hover"]} overlay={i18next.t(this.state.threeboxPosts?.length ? "Enable3BoxInteractions" : "CreateFirst3BoxPost")}>
+              <a onClick={this.startDiscussion}>{i18next.t(this.state.threeboxPosts?.length ? "JoinTheConversation" : "StartAConversation")}{this.state.joiningThread ? <img className={css.loading} src="/assets/images/Icon/buttonLoadingWhite.gif" /> : ""}</a>
+            </Tooltip>
+            : ""
```
roienatan, line 186: Can you please not use the pattern of empty quotation marks? Instead you can use `&&`, for example: 
`{ this.state.joinedThread && <div>your content</div> } `

I find this empty quotation marks pattern confusing and I'm trying to eliminate this across Alchemy.
I would use the ternary operation only when I have something to show if the other condition met, otherwise `&&` is much more cleaner.

Please change on the other places also.

src/components/Shared/ThreeBoxThreads.scss
```diff
@@ -0,0 +1,131 @@
...
+        .message {
+          overflow: hidden;
+          white-space: pre-wrap;
+          word-break: break-word;
+          img {
+            max-width: 100%;
+          }
+
+          a {
+            color: #0071ff;
```
roienatan, line 94: Global variable for color, I guess we already use this color in the globals.

roienatan (changes requested):

src/components/Shared/ThreeBoxThreads.tsx
dkent600: frankly I find your logic much more confusing because I have to understand how the boolean expression is evaluated left to right coupled with the fact that the boolean value of the <div> is totally irrelevant and that the final result of the whole expression is never actually evaluated, suggesting to me that some linters might complain and compilers/optimizers might motivated to remove the entire block.

dkent600: Further, asking me to change this logic risks creating bugs forcing me to test everything all over again.  Would have been better to asked for this sort of thing in advance.

src/components/Shared/ThreeBoxThreads.scss
dkent600: ok, I updated 26 files from color variables I've in this PR.  No way I'm testing all that though...

src/components/Shared/ThreeBoxThreads.tsx
dkent600: On further reflection I realize that contrary to what I asserted above, the final result of the whole expression actually *is* evaluated.  But it that doesn't much alter the fact that the expression has taken me a while to fully grok, a lot longer than the terniary, so I'm still not very compelled to agree that this is a simpler way of doing this.

dkent600: Maybe I'm just dense.  Bottom line is I'd prefer not have to retest all this code for this late request.

dkent600: BTW, react sucks

dkent600: If you beg me I'll do it

dkent600: OK OK I did it in the one file, but I'll go on record as saying that for me the new logic is less intuitive/clear than terniaries.

And that react sucks in any case.

PR #2151, opened 2020-09-29
introduce polling

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/2126 for 1.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
   "dependencies": {
     "3box": "1.17.1",
     "@burner-wallet/burner-connect-provider": "^0.1.1",
-    "@daostack/arc.js": "0.2.73",
+    "@daostack/arc.js": "^0.2.74",
     "@dorgtech/daocreator-ui": "^1.0.13",
     "@fortawesome/fontawesome-svg-core": "^1.2.10",
     "@fortawesome/free-brands-svg-icons": "^5.6.1",
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
   "dependencies": {
     "3box": "1.17.1",
     "@burner-wallet/burner-connect-provider": "^0.1.1",
-    "@daostack/arc.js": "^0.2.74",
+    "@daostack/arc.js": "0.2.74",
     "@dorgtech/daocreator-ui": "^1.0.13",
     "@fortawesome/fontawesome-svg-core": "^1.2.10",
     "@fortawesome/free-brands-svg-icons": "^5.6.1",
--- a/src/components/Dao/ProposalSchemeCard.tsx
+++ b/src/components/Dao/ProposalSchemeCard.tsx
@@ -4,7 +4,7 @@ import VoteGraph from "components/Proposal/Voting/VoteGraph";
 import ProposalCountdown from "components/Shared/ProposalCountdown";
 import Loading from "components/Shared/Loading";
 import withSubscription, { ISubscriptionProps } from "components/Shared/withSubscription";
-import { humanProposalTitle } from "lib/util";
+import { humanProposalTitle, standardPolling } from "lib/util";
 import { schemeName } from "lib/schemeUtils";
 import * as React from "react";
 import { Link } from "react-router-dom";
@@ -54,7 +54,7 @@ const ProposalSchemeCard = (props: IProps) => {
         <div>
           <b>{schemeState.numberOfBoostedProposals}</b> <span>Boosted</span> <b>{schemeState.numberOfPreBoostedProposals}</b> <span>Pending Boosting</span> <b>{schemeState.numberOfQueuedProposals}</b> <span>Regular</span>
         </div>
-        {proposals.length === 0 ?
+        {numProposals === 0 ?
           <div className={css.loading}>
             <img src="/assets/images/meditate.svg" />
             <div>
@@ -65,7 +65,7 @@ const ProposalSchemeCard = (props: IProps) => {
         }
       </Link>
 
-      {proposals.length > 0 ?
+      {numProposals > 0 ?
         <div>
           {proposalsHTML}
           <div className={css.numProposals}>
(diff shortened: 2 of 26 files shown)


package.json
```diff
@@ -79,7 +79,7 @@
   "dependencies": {
     "3box": "1.17.1",
     "@burner-wallet/burner-connect-provider": "^0.1.1",
-    "@daostack/arc.js": "0.2.73",
+    "@daostack/arc.js": "^0.2.74",
```
orenyodfat, line 82: remove the `^ ` ?

src/components/Dao/ProposalSchemeCard.tsx
```diff
@@ -54,7 +54,7 @@ const ProposalSchemeCard = (props: IProps) => {
         <div>
           <b>{schemeState.numberOfBoostedProposals}</b> <span>Boosted</span> <b>{schemeState.numberOfPreBoostedProposals}</b> <span>Pending Boosting</span> <b>{schemeState.numberOfQueuedProposals}</b> <span>Regular</span>
         </div>
-        {proposals.length === 0 ?
+        {numProposals === 0 ?
```
orenyodfat, line 57: how this is related ?

src/components/Dao/ProposalSchemeCard.tsx
```diff
@@ -91,17 +91,14 @@ export default withSubscription({
     const arc = getArc();
     const dao = arc.dao(props.dao.address);
     return combineLatest(
-      props.scheme.state(),
+      props.scheme.state(standardPolling()),
```
orenyodfat, line 94: why adding standardPolling here ?

src/components/Dao/ProposalSchemeCard.tsx
```diff
@@ -91,17 +91,14 @@ export default withSubscription({
...
         scheme:  props.scheme.id,
         // eslint-disable-next-line @typescript-eslint/camelcase
         stage_in: [IProposalStage.Boosted, IProposalStage.QuietEndingPeriod],
       },
       orderBy: "boostedAt",
-      }, {
-        fetchAllData: true,
-        subscribe: true, // subscribe to updates of the proposals. We can replace this once https://github.com/daostack/subgraph/issues/326 is done
-      }) // the list of boosted proposals
+      }, { fetchAllData: true }) // the list of boosted proposals
```
orenyodfat, line 101: no need polling here ?

src/components/Scheme/SchemeContainer.tsx
```diff
@@ -207,7 +207,7 @@ class SchemeContainer extends React.Component<IProps, IState> {
                         href="#!"
                         onClick={this.handleEditPlugin}
                       >
-                    Edit Plugin
+                        Edit Plugin
```
orenyodfat, line 210: why adding these spaces ?

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
```diff
@@ -105,7 +111,9 @@ export interface ICreateSubmissionOptions {
   tags: Array<string>;
 }
 
-export const createCompetitionSubmission = (proposalId: string, options: ICreateSubmissionOptions ): ThunkAction<any, IRootState, null> => {
+const standardPolling = (poll: boolean, fetchAllData = false) => { return { polling: poll, pollInterval: GRAPH_POLL_INTERVAL, fetchAllData }; };
```
orenyodfat, line 114: does this not defined somewhere else ?m lib/util ?

orenyodfat (changes requested):

package.json
dkent600: ah yes, thanks for catching that

src/components/Dao/ProposalSchemeCard.tsx
dkent600: Reviewing this I find I fixed it wrong.  So I've undone this change but added a different change on line 101 of this file.  The problem is that the list of boosted proposals was not going to be responsive to new boosted or unboosted proposals.

src/components/Dao/ProposalSchemeCard.tsx
dkent600: because this scheme state would not update otherwise, and we need that or else the scheme proposal count properties will not be responsive to changes as proposals move through their lifecycle.

src/components/Dao/ProposalSchemeCard.tsx
dkent600: Ha, I just added that (see previous comment above)

src/components/Scheme/ContributionRewardExtRewarders/Competition/utils.ts
dkent600: this is a different version of the function that is only needed in this one module.

src/components/Scheme/SchemeContainer.tsx
dkent600: that is a mystery.  I was able to restore them and neither lint nor my editor is complaining.

src/lib/util.ts
```diff
@@ -702,3 +703,6 @@ export function safeMoment(dateSpecifier: moment.Moment | Date | number | string
       throw new Error(`safeMoment: unknown type: ${typeof dateSpecifier}`);
   }
 }
+
+export const standardPolling = (fetchAllData = false) =>
```
roienatan, line 707: With this function you lose the ability to give different poll intervals for different polls, right?

orenyodfat (approved):

src/lib/util.ts
dkent600: you don't lose it, you just have to write code to do it, which would be simple.  Currently that functionality is not required.

roienatan (approved):

Issue #2183, opened 2020-10-11
Show different message when no write methods found in ABI


roienatan:
When creating a unknown generic plugin proposal, if we found an ABI but no write methods we should give different error message and not "ABI not found..."

PR #2184, opened 2020-10-11
Different message when no write methods found in ABI

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2183


src/assets/locales/en/translation.json
```diff
@@ -52,6 +52,7 @@
     "Validate Digits" : "Must contain only digits",
     "Call to Contract": "This proposal will call the following ",
     "No ABI": "No ABI found for target contract, please verify the ",
+    "No Write Methods": "No write methods found for target contract, please verify the ",
```
orenyodfat, line 55: if no "write" methods are found so verify the contract will not help.
I think it should be only 
`No write methods found for target contract`

orenyodfat (approved):

PR #2187, opened 2020-10-13
Support generic multi-call scheme - 2.0

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2186

Also:
- [x] Update the single call generic scheme to be consistent with the multi-call (removed SelectSearch component)
- [x] Use same validation code to reduce duplicated code
- [x] Fix issue that user data is not saved when you close the proposal modal and go back.
- [x] More minor code improvements


orenyodfat:
duplicate at https://github.com/daostack/alchemy_2/pull/4

Issue #2178, opened 2020-10-07
link v1->v2 and v2-v1


orenyodfat:
https://www.figma.com/file/mFrV7mIFXFmxjW4C1qariX/Untitled?node-id=0%3A1

PR #2188, opened 2020-10-13
add new sidebar links - 1.0

dkent600:
Adds the links spec'd here: https://github.com/daostack/alchemy/issues/2178, for 1.0

bumps package version to 1.0.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alchemy-client",
-  "version": "0.10.17",
+  "version": "1.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "alchemy-client",
-  "version": "0.10.17",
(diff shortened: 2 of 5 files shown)


orenyodfat (changes requested):
- [ ] the upper bar as it appears in the figma does not showed up in the review app 
  
Screen Shot 2020-10-14 at 12 30 39

- [ ] lets use env vars for the links so it can be controlled . please note the v1.xdai.alchemy.do still does not work ( for now it should be https://alchemy-xdai.herokuapp.com/ )

dkent600:
> * [ ]  the upper bar as it appears in the figma does not showed up in the review app

that will be part of a second PR

dkent600:
> * [ ]  lets use env vars for the links so it can be controlled . please note the v1.xdai.alchemy.do still does not work ( for now it should be https://alchemy-xdai.herokuapp.com/ )

@orenyodfat Use env vars far what?  Defining which URLs to use?

orenyodfat (approved):

PR #2189, opened 2020-10-13
add new sidebar links - 2.0

dkent600:
Adds the links spec'd here: https://github.com/daostack/alchemy/issues/2178, for 2.0

Bumps package version to 2.0.0

### DIFF
--- a/webpack.base.config.js
+++ b/webpack.base.config.js
@@ -122,7 +122,7 @@ module.exports = {
       PACKAGE_VERSION: JSON.stringify(require("./package.json").version)
     }),
     new webpack.EnvironmentPlugin({
-      ALCHEMY_URL_V1_XDAI: "https://alchemy-xdai.herokuapp.com/ ", // will be: https://v1.xdai.alchemy.do
+      ALCHEMY_URL_V1_XDAI: "https://v1.xdai.alchemy.do",
       ALCHEMY_URL_V1_MAINNET: "https://v1.alchemy.do",
       ALCHEMY_URL_V2_XDAI: "https://xdai.alchemy.do",
       ALCHEMY_URL_V2_MAINNET: "https://alchemy.do",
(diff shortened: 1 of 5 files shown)


roienatan (changes requested):
Please add some space between "switch to mainnet" and "DAOstack", it looks a bit ugly especially when hovering with the mouse.
Screen Shot 2020-10-14 at 15 52 35

dkent600:
> Please add some space between "switch to mainnet" and "DAOstack", it looks a bit ugly especially when hovering with the mouse.

The problem is that it is wrapping, and it probably shouldn't be.    That wasn't happening for me.  Can you please describe how you reproduced it?

roienatan:
> > Please add some space between "switch to mainnet" and "DAOstack", it looks a bit ugly especially when hovering with the mouse.
> 
> @roienatan The problem is that it is wrapping, and it probably shouldn't be. That wasn't happening for me. Can you please describe how you reproduced it?

I just launched the review app :)
I see, it's related to the screen size, when I zoomed out it was not wrapping and when when I zoomed even more out it's wrapping which is really weird.

webpack.base.config.js
```diff
@@ -122,7 +122,7 @@ module.exports = {
       PACKAGE_VERSION: JSON.stringify(require("./package.json").version)
     }),
     new webpack.EnvironmentPlugin({
-      ALCHEMY_URL_V1_XDAI: "https://alchemy-xdai.herokuapp.com/ ", // will be: https://v1.xdai.alchemy.do
+      ALCHEMY_URL_V1_XDAI: "https://v1.xdai.alchemy.do",
```
orenyodfat, line 125: this can be override by the env var ?

dkent600: > this can be override by the env var ?

Yes

orenyodfat (approved):

roienatan (approved):

PR #2192, opened 2020-10-15
lowercase tokens, proposalState to RedemptionsTip

dkent600:
bugs found while investigating the PrimeDAO redeeming bug

### DIFF
--- a/data/tokens.json
+++ b/data/tokens.json
@@ -33,7 +33,7 @@
         "name": "Paxos Standard",
         "symbol": "PAX"
       },
-      "0x0000000000085d4780B73119b644AE5ecd22b376": {
+      "0x0000000000085d4780b73119b644ae5ecd22b376": {
         "decimals": 18,
         "name": "TrueUSD",
         "symbol": "TUSD"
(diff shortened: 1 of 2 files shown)


roienatan:
So it fixed the issue? I wonder why it's not happened while redeeming other items...

dkent600:
> I wonder why it didn't happen while redeeming other items...

No this doesn't fix the Redeem problem.  That appears to be happening in the contracts somewhere.  In the course of investigating we found these two other issues in Alchemy.  One causes the token name not to be displayed.  The other causes the tooltip to not appear over the Redeem button.

roienatan:
> > I wonder why it didn't happen while redeeming other items...
> 
> 
> 
> No this doesn't fix the Redeem problem.  That appears to be happening in the contracts somewhere.  In the course of investigating we found these two other issues in Alchemy.  One causes the token name not to be displayed.  The other causes the tooltip to not appear over the Redeem button.

I think it appears on other items.
What issue the change in ActionButton solves?

dkent600:
> What issue the change in ActionButton solves?

The tooltip not to appear over the Redeem button.

roienatan (approved):

PR #2197, opened 2020-10-16
Support generic multi-call scheme - 1.0

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2186 for 1.0


src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
```diff
@@ -0,0 +1,502 @@
...
+      contractsToCall: contractsToCall,
+      callsData: callsData,
+      values: values,
+      url: formValues.url,
+      dao: this.props.daoAvatarAddress,
+      scheme: this.props.scheme.address,
+      tags: this.state.tags,
+    };
+
+    setSubmitting(false);
```
orenyodfat, line 134: what this one is doing ?

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
```diff
@@ -0,0 +1,502 @@
...
+ * @param {string} contractToCall
+ */
+  private verifyContract = async (contractToCall: string) => {
+    const addContractStatus: IAddContractStatus = {
+      error: "",
+      message: "",
+    };
+
+    if (contractToCall === "") {
+      addContractStatus.error = "";
```
orenyodfat, line 169: in that case no need error msg?

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
```diff
@@ -0,0 +1,502 @@
...
+      addContractStatus.message = "";
+    } else if (!isAddress(contractToCall)) {
+      addContractStatus.error = "NOT_VALID_ADDRESS";
+      addContractStatus.message = "Please enter a valid address";
+    } else if (this.state.whitelistedContracts.includes(contractToCall) || this.state.userContracts.includes(contractToCall)) {
```
orenyodfat, line 174: why this check ?

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
```diff
@@ -0,0 +1,502 @@
...
+      this.setState({ loading: true });
+      const abiData = await getABIByContract(contractToCall);
+      const abiMethods = extractABIMethods(abiData);
+      if (abiMethods.length > 0) {
+        this.state.userContracts.push(contractToCall);
+        addContractStatus.error = "";
+        addContractStatus.message = "Contract added successfully!";
+      } else {
+        addContractStatus.error = "ABI_DATA_ERROR";
+        addContractStatus.message = abiData.length === 0 ? "No ABI found for target contract, please verify the " : "No write methods found for target ";
```
orenyodfat, line 187: is it suppose to be `"No write ABI found...` ?

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
```diff
@@ -0,0 +1,502 @@
...
+      error: "",
+      message: "",
+    };
+
+    if (contractToCall === "") {
+      addContractStatus.error = "";
+      addContractStatus.message = "";
+    } else if (!isAddress(contractToCall)) {
+      addContractStatus.error = "NOT_VALID_ADDRESS";
+      addContractStatus.message = "Please enter a valid address";
```
orenyodfat, line 173: it might be nicer to have `setContractStatus(error, msg)` and call that on the different cases

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
```diff
@@ -0,0 +1,502 @@
...
+      if (abiMethods.length > 0) {
+        this.state.userContracts.push(contractToCall);
+        addContractStatus.error = "";
+        addContractStatus.message = "Contract added successfully!";
+      } else {
+        addContractStatus.error = "ABI_DATA_ERROR";
+        addContractStatus.message = abiData.length === 0 ? "No ABI found for target contract, please verify the " : "No write methods found for target ";
+      }
+    }
+    this.setState({ loading: false, addContractStatus: addContractStatus });
```
orenyodfat, line 190: so the state is set even there is an error?

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
```diff
@@ -0,0 +1,502 @@
...
+      addContractStatus.message = "";
+    } else if (!isAddress(contractToCall)) {
+      addContractStatus.error = "NOT_VALID_ADDRESS";
+      addContractStatus.message = "Please enter a valid address";
+    } else if (this.state.whitelistedContracts.includes(contractToCall) || this.state.userContracts.includes(contractToCall)) {
+      addContractStatus.error = "CONTRACT_EXIST";
+      addContractStatus.message = "Contract already exist!";
+    } else {
+      this.setState({ loading: true });
+      const abiData = await getABIByContract(contractToCall);
```
orenyodfat, line 179: is this not called twice ? one here for verify and other time at https://github.com/daostack/alchemy/pull/2197/files#diff-ea7dd2c9cc24fa59c55016881a6d0bddcc07ee4acf441f88989a3c32ac604fd3R196

src/components/Proposal/Create/SchemeForms/Validators.ts
```diff
@@ -0,0 +1,49 @@
...
+import { isAddress } from "lib/util";
+
+/**
+ * Given a value returns error message in case value is less than 0 or no value provided
+ * @param {any} value
+ */
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const requireValue = (value: any): string => {
+  let error;
+  if (value === "") {
```
orenyodfat, line 11: value is string or number ?

src/components/Proposal/Create/SchemeForms/Validators.ts
```diff
@@ -0,0 +1,49 @@
+import { isHexStrict } from "web3-utils";
+import { isAddress } from "lib/util";
+
+/**
+ * Given a value returns error message in case value is less than 0 or no value provided
```
orenyodfat, line 5: return error string,

src/components/Proposal/Create/SchemeForms/Validators.ts
```diff
@@ -0,0 +1,49 @@
...
+        if (!isHexStrict(value)) {
+          error = "Must be a hex value";
+        }
+        break;
+      case type.includes("uint"):
+        if (/^\d+$/.test(value) === false) {
+          error = "Must contain only digits";
+        }
+        break;
+    }
```
orenyodfat, line 46: add default to switch

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: This is Forimk function to inform the form that we're no longer in a submitting process. This is a standard in all our Formik forms.

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: No, this field is not really part of the form, it's just to add a new contract to the select picker. It's okay if it's empty.

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: Not necessary anymore, it was relevant when we gave the option to add custom contract even if we have a whitelist. Will update, thanks.

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
orenyodfat: so why checking that at all ?

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: It's necessary because here we call it to check if this contract is verified and has write methods.
In line 196 we call each time we change a contract. I don't want to call all the list in advance because when you have a whitelist the user may not use all the contracts in the whitelist.

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: Yes, the error state is used to display what kind of error occured (Not valid address, contract exist already in the list or contract has not valid ABI data)

src/components/Proposal/Create/SchemeForms/Validators.ts
roienatan: I'm not sure I understand, anyway this is the way Formik handles errors in a from.

src/components/Proposal/Create/SchemeForms/Validators.ts
roienatan: This is kind of a hack. For some reason if there is no value supplied Formik considers the empty value as string, and when the user enters a value of course it's a number.

src/components/Proposal/Create/SchemeForms/Validators.ts
orenyodfat: add documentation for the return value for this function

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: When the user tries to add a new custom contract we need to check:

- Valid address
- The contract not already exist in the list (that he does not add the same contract twice)
- That the contract has a valid ABI data (verified with write methods)

If one of the above conditions does not met, there is not reason to add this contract to the select picker.

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
orenyodfat: what do you mean by `each time we change a contract.` ? you mean each time user enter a new call ?

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: Ahh because if an error already occurred and the user cleans the input so we want to remove the last error.

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: Agree.

roienatan: In a second thought, that's a very specific case and the helper function will only be used here, but I'll try to do something nicer.

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: No, each time the user selects different contract in the select picker.

src/components/Proposal/Create/SchemeForms/Validators.ts
roienatan: In this case it's not necessary to give a default value since if no case is true in the switch statement, an `undefined` will be returned (line 50). I can add default case: `error = undefined` but it will do nothing actually (because it's already returned in line 50).
I can eliminate line 50 and then `return error;` in each case instead of the `break` statement and `return error;` in the first `if` but it will give no optimization or something.

src/components/Proposal/Create/SchemeForms/Validators.ts
```diff
@@ -0,0 +1,51 @@
...
+  else {
+    switch (true) {
+      case type.includes("address"):
+        if (!isAddress(value)) {
+          error = "Please enter a valid address";
+        }
+        break;
+      case type.includes("byte"):
+        if (!isHexStrict(value)) {
+          error = "Must be a hex value";
```
orenyodfat, line 40: `Must be an hex`

src/components/Proposal/Create/SchemeForms/Validators.ts
```diff
@@ -0,0 +1,51 @@
...
+ * @param {string} value
+ * @returns {undefined | string} undefined if there is no error, otherwise a string representing the error message.
+ */
+export const validateParam = (type: string, value: string): string => {
+  let error;
+  if (!value && type !== "bool") {
+    error = "Required";
+  }
+  else {
+    switch (true) {
```
orenyodfat, line 32: does the switch here is really require ?
how about replacing that to if-else statement  ?

roienatan: Why not? I think it's more elegant in this case. I use `true` since this is the way to use dynamic values in the `case` statements.

src/components/Proposal/Create/SchemeForms/Validators.ts
```diff
@@ -0,0 +1,51 @@
+import { isHexStrict } from "web3-utils";
+import { isAddress } from "lib/util";
+
+/**
+ * Given a value returns error message in case value is less than 0 or no value provided
+ * @param {any} value
+ */
+// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
+export const requireValue = (value: any): string => {
```
orenyodfat, line 9: please document that this function will return `undefined` if all is ok

src/components/Proposal/Create/SchemeForms/Validators.ts
```diff
@@ -0,0 +1,51 @@
...
+  else {
+    switch (true) {
+      case type.includes("address"):
+        if (!isAddress(value)) {
+          error = "Please enter a valid address";
+        }
+        break;
+      case type.includes("byte"):
+        if (!isHexStrict(value)) {
+          error = "Must be an hex";
```
orenyodfat, line 40: `Must be an hex value`

orenyodfat (approved):
lets make tests pass though

PR #2230, opened 2020-11-01
Avoid crashing Alchemy when there is no network or The Graph servers are overloaded - 2.0

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2231 for 2.0

### DIFF
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -124,7 +124,7 @@ export async function initializeArc(provider?: Web3Provider): Promise<boolean> {
     // https://www.apollographql.com/docs/link/links/retry/
     const retryLink = new RetryLink({
       attempts: (count) => {
-        if (count === 10) {
+        if (count !== 10) {
           return false;
         }
         return true;
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -124,12 +124,10 @@ export async function initializeArc(provider?: Web3Provider): Promise<boolean> {
     // https://www.apollographql.com/docs/link/links/retry/
     const retryLink = new RetryLink({
       attempts: (count) => {
-        if (count !== 10) {
-          return false;
-        }
-        return true;
+        return (count !== 10);
       },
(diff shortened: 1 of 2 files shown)


orenyodfat (changes requested):
it does not seems to continue polling after the network error

roienatan:
> it does not seems to continue polling after the network error

Looks like React Apollo Client does not resume polling after a network error, see this:
https://github.com/apollographql/apollo-client/issues/1099
I need to investigate and see if there is an option to resume the polling. The fix might be in arc.js if so.

src/arc.ts
```diff
@@ -124,21 +123,14 @@ export async function initializeArc(provider?: Web3Provider): Promise<boolean> {
...
-        max: 5,
-        retryIf: (error, _operation) => {
-          // eslint-disable-next-line no-console
-          console.error("error occurred fetching data, retrying...");
-          // eslint-disable-next-line no-console
-          console.log(error);
-          return !!error;
-        },
+      attempts: (count) => {
+        if (count === 10) {
```
orenyodfat, line 127: this lines can be changed to : 
`return (count !== 10);`

src/arc.ts
```diff
@@ -124,21 +123,14 @@ export async function initializeArc(provider?: Web3Provider): Promise<boolean> {
...
+        }
+        return true;
       },
-      delay: {
-        initial: 500, // this is the initial time after the first retry
-        // next retries )up to max) will be exponential (i..e after 2*iniitial, etc)
-        jitter: true,
-        max: Infinity,
+      delay: () => {
+        return Math.floor(Math.random() * (30000 - 5000 + 1) + 5000);
```
orenyodfat, line 133: lets add a comment here : 
" this will give a random delay between retries between the range of 5 to 30 seconds"

orenyodfat (approved):

PR #2253, opened 2020-11-10
Release 1.0.2

roienatan:


EzraWeller:
@roienatan I'm getting invalid format for ['0xE11BA2b4D45Eaed5996Cd0823791E0C93114882d','0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1'] on a field that says it takes address[] in the multicall -- am I screwing that up somehow?

roienatan:
> @roienatan I'm getting invalid format for ['0xE11BA2b4D45Eaed5996Cd0823791E0C93114882d','0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1'] on a field that says it takes address[] in the multicall -- am I screwing that up somehow?

Please use double quotes

EzraWeller:
Another small thing: it seems like `approve` calls need the spender to also be whitelisted. Is that something we were already aware of?

orenyodfat:
> Another small thing: it seems like `approve` calls need the spender to also be whitelisted. Is that something we were already aware of? We might want to make it easier to copy the whitelisted addresses to the clipboard to facilitate UX for this. With the non-whitelisted version, I assume this isn't an issue. @roienatan

it is a matter of the scheme constraint it is using.
for now , we decided to go with a simple only whitelisted or not at all constraint.
Yes, on the original dxdaoConstraint we also require the spender to be whitelisted but we are not using that.
If there will be such a constraint ,which require the spender to be whitelisted,  it will be forced by the contract ..so no proposal can be done unless the spender is white listed.

EzraWeller:
> > Another small thing: it seems like `approve` calls need the spender to also be whitelisted. Is that something we were already aware of? We might want to make it easier to copy the whitelisted addresses to the clipboard to facilitate UX for this. With the non-whitelisted version, I assume this isn't an issue. @roienatan
> 
> it is a matter of the scheme constraint it is using.
> for now , we decided to go with a simple only whitelisted or not at all constraint.
> Yes, on the original dxdaoConstraint we also require the spender to be whitelisted but we are not using that.
> If there will be such a constraint ,which require the spender to be whitelisted, it will be forced by the contract ..so no proposal can be done unless the spender is white listed.

Yes, I get the reason why, but we might want to make it easier for the user to copy the whitelisted addresses so they can use them as method inputs. Can we add that @roienatan ?

roienatan:
> > > Another small thing: it seems like `approve` calls need the spender to also be whitelisted. Is that something we were already aware of? We might want to make it easier to copy the whitelisted addresses to the clipboard to facilitate UX for this. With the non-whitelisted version, I assume this isn't an issue. @roienatan
> > 
> > 
> > it is a matter of the scheme constraint it is using.
> > for now , we decided to go with a simple only whitelisted or not at all constraint.
> > Yes, on the original dxdaoConstraint we also require the spender to be whitelisted but we are not using that.
> > If there will be such a constraint ,which require the spender to be whitelisted, it will be forced by the contract ..so no proposal can be done unless the spender is white listed.
> 
> Yes, I get the reason why, but we might want to make it easier for the user to copy the whitelisted addresses so they can use them as method inputs. Can we add that @roienatan ?

I can think about something. Do we want it for the current release or it's not that urgent?

EzraWeller:
So I just passed a multicall proposal, and I don't see how to execute it:
Screen Shot 2020-11-18 at 9 00 14 AM
It says I can redeem rep, and there's a note on this page that says it has already been executed. In history, it says the status is passed.

@orenyodfat @roienatan

EzraWeller:
Okay, I'm not seeing issues with anything other than the multicall, so I'll approve this, but can we move the above issues I raised with the multicall to the next release? Let me know the preferred way to keep that organized.

EzraWeller (dismissed):

Razzwan (approved):

roienatan:
> So I just passed a multicall proposal, and I don't see how to execute it:
> Screen Shot 2020-11-18 at 9 00 14 AM
> It says I can redeem rep, and there's a note on this page that says it has already been executed. In history, it says the status is passed.
> 
> @orenyodfat @roienatan

@EzraWeller So, what's the issue? It's passed since it has absolute majority.

PR #2254, opened 2020-11-10
Support method arrays params in generic multi-call

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2255

### DIFF
--- a/src/components/Proposal/Create/SchemeForms/ABIService.ts
+++ b/src/components/Proposal/Create/SchemeForms/ABIService.ts
@@ -4,6 +4,7 @@ const Web3 = require("web3");
 import axios from "axios";
 import { targetedNetwork } from "lib/util";
 const InputDataDecoder = require("ethereum-input-data-decoder");
+import { isArrayParameter } from "./Validators";
 
 export interface IAllowedAbiItem extends AbiItem {
   name: string;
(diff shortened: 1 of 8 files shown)


orenyodfat:
- the numbering should be per call , not per contract.it should be :
# 1
contract(0x...)

not 

Screen Shot 2020-11-11 at 10 05 21

orenyodfat:
I think that adding the param names in the function prototype will give better UI.
so instead of

Screen Shot 2020-11-11 at 10 07 44

it should be

`method approve(address _spender,uint _value,bytes _data)`

orenyodfat:
the ETH value on the new proposal should be displayed only after the user choose a contact along with the list of Abis

roienatan:
> * the numbering should be per call , not per contract.it should be :
>   `# 1`
>   contract(0x...)
> 
> not
> 
> Screen Shot 2020-11-11 at 10 05 21

It's quite the same since each contract is necessarily a call, but okay.

roienatan:
> * the numbering should be per call , not per contract.it should be :
>   `# 1`
>   contract(0x...)
> 
> not
> 
> Screen Shot 2020-11-11 at 10 05 21

Done.

roienatan:
> I think that adding the param names in the function prototype will give better UI.
> so instead of
> 
> Screen Shot 2020-11-11 at 10 07 44
> 
> it should be
> 
> `method approve(address _spender,uint _value,bytes _data)`

Done.

roienatan:
> the ETH value on the new proposal should be displayed only after the user choose a contact along with the list of Abis

As discussed, we will wait with that.

orenyodfat (approved):

PR #2256, opened 2020-11-11
fix throwing an error and try to subscirbe in 5sec. later

Razzwan:


src/components/Shared/withSubscription.tsx
```diff
@@ -116,7 +116,8 @@ const withSubscription = <Props extends ISubscriptionProps<ObservableType>, Obse
           // eslint-disable-next-line no-console
           console.error(getDisplayName(wrappedComponent), "Error in subscription", error);
           // this will go to the error page
-          this.setState(() => { throw error; });
+          this.subscription.unsubscribe();
+          setTimeout(this.setupSubscription.bind(this, observable), 5000);
```
orenyodfat, line 120: I think the time out here should be a random one between 2500 to 7500.  to avoid multiple  clients to access the service in the same time (can overload the service)

orenyodfat: also we need some additional logic here as ignoring error for long time should give false information to the user.
something like :  throw after 10 retry attempts.

PR #2257, opened 2020-11-12
Ignore network and GraphQL errors

roienatan:
Due to the way Apollo Client works, when a GraphQL or a network error occurs it fails and terminates the observable including the polling, then Alchemy crashes. This is the simplest and cleanest solution to address this issue without changing anything in arc.js.
This is a known bug in Apollo Client `watchQuery` when polling, and looks like the only way to avoid this is to recreate the polling when such an error occurs.

### DIFF
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -152,7 +152,7 @@ export async function initializeArc(provider?: any): Promise<boolean> {
       },
     });
 
-    arcSettings.retryLink = retryLink;
+    arcSettings.graphqlRetryLink = retryLink;
 
     // if there is no existing arc, we create a new one
     if ((window as any).arc) {
--- a/src/arc.ts
+++ b/src/arc.ts
@@ -1,7 +1,6 @@
 import { NotificationStatus } from "reducers/notifications";
 import { getNetworkId, getNetworkName, targetedNetwork } from "./lib/util";
 import { settings, USE_CONTRACTINFOS_CACHE } from "./settings";
-import { RetryLink } from "apollo-link-retry";
 import { Address, Arc } from "@daostack/arc.js";
 import Web3Modal, { getProviderInfo, IProviderInfo } from "web3modal";
 import { Observable } from "rxjs";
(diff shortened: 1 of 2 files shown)


src/arc.ts
```diff
@@ -134,16 +134,25 @@ export async function initializeArc(provider?: any): Promise<boolean> {
...
+      delay: {
+        initial: 500, // this is the initial time after the first retry
+        // next retries )up to max) will be exponential (i..e after 2*iniitial, etc)
+        jitter: true,
+        max: Infinity,
       },
     });
 
-    arcSettings.graphqlRetryLink = retryLink;
+    arcSettings.retryLink = retryLink;
```
orenyodfat, line 155: this should be `arcSettings.graphqlRetryLink `. or do not set retryLink at all as arc.js expect `graphqlRetryLink`

src/components/Shared/withSubscription.tsx
```diff
@@ -116,7 +117,16 @@ const withSubscription = <Props extends ISubscriptionProps<ObservableType>, Obse
...
           console.error(getDisplayName(wrappedComponent), "Error in subscription", error);
           // this will go to the error page
-          this.setState(() => { throw error; });
+          /**
+           * The below condition is a workaround to avoid crashing Alchemy when a GraphQL error or a Network error occurs.
+           * This is due to the way Apollo Client works when such an error occurs - it fails and terminates the observable including the polling.
+           */
+          if (error.message.includes("GraphQL") || error.message.includes("Network")) {
+            this.subscription.unsubscribe();
+            setTimeout(this.setupSubscription.bind(this, observable), GRAPH_POLL_INTERVAL);
```
orenyodfat, line 126: https://github.com/daostack/alchemy/pull/2256#discussion_r522907629

src/arc.ts
```diff
@@ -134,12 +134,21 @@ export async function initializeArc(provider?: any): Promise<boolean> {
 
     // https://www.apollographql.com/docs/link/links/retry/
     const retryLink = new RetryLink({
```
orenyodfat, line 136: I think we can remove the retry link @roienatan

roienatan: Why?

orenyodfat: because the subscription already cover that in general.

roienatan: Note that `withSubscription` only handles Network and GraphQL errors (for now). If we're sure that those are the only possible errors so `retryLink` can be safely removed.

orenyodfat: other errors should be throw for now..so lets removed the retryLink.

orenyodfat (approved):

PR #2264, opened 2020-11-16
Support multiple networks

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2261
resolves: https://github.com/daostack/alchemy/issues/2282


src/lib/util.ts
```diff
@@ -270,6 +270,27 @@ export const getNetworkByProvider = (provider: any): Networks => {
   return "main";
 };
 
+/**
+ *
+ * @param dao
+ */
+export const getNetworkByDAO = (DAO: any): Networks => {
+  const arcs = getArcs();
```
orenyodfat, line 278: here you can do `getDaos`

src/lib/util.ts
```diff
@@ -270,6 +270,27 @@ export const getNetworkByProvider = (provider: any): Networks => {
...
+  for (const network in arcs) {
+    const arc = arcs[network];
+    try {
```
orenyodfat, line 281: no need try/catch here..

src/lib/util.ts
orenyodfat: see e807ebe

EzraWeller:
A few things I'd like to see addressed with this release, ideally:

---
After searching "DXdao," I see copies of the same DAO:

Screen Shot 2020-11-27 at 9 05 08 AM

---

From the DAOs page, if you are connected to xDai and try to unfollow a mainnet DAO, it will log you out and give you the "please connect to mainnet" alert. Can we keep the alert but not make it log you out?

---

DAOcreator does not work if you are connected to xDai. Console error:

Screen Shot 2020-11-27 at 10 46 36 AM

(Probably an issue for us to ask Cesar to deal with?)

---

No bubble around the network name please:

Screen Shot 2020-11-27 at 10 54 08 AM

---

Adjust the account information to show only the current network and current network's assets:
Instead of :
```
'Holdings - main
{ assets }
Holdings - xdai
{ assets }
Provider
{provider}'
```
Let's have:
```
'Holdings
{assets}
Provider
{provider}
Network
{network}'
```
Screen Shot 2020-11-27 at 2 36 05 PM

orenyodfat:
> After searching "DXdao," I see copies of the same DAO:
> 
> Screen Shot 2020-11-27 at 9 05 08 AM

fixed with 2afc3f7

orenyodfat:
> Adjust the account information to show only the current network and current network's assets:
what wrong to show holding from all networks ? @EzraWeller

EzraWeller:
> > Adjust the account information to show only the current network and current network's assets:
> 
> what wrong to show holding from all networks ? @EzraWeller

It makes it even less clear which network you are actually connected to and I don't think it adds anything -- a lot of the time at least one of the two says `N/A` for me anyway? Are we trying to store that information in cookies or something?

orenyodfat:
> > > Adjust the account information to show only the current network and current network's assets:
> > 
> > 
> > what wrong to show holding from all networks ? @EzraWeller
> 
> It makes it even less clear which network you are actually connected to and I don't think it adds anything -- a lot of the time at least one of the two says `N/A` for me anyway? Are we trying to store that information in cookies or something?

the N/A is because the infura account which is set in the review app . It is nothing. will fix soon

EzraWeller:
Okay, a last set of small feedbacks, @roienatan ! I will also push DXdao to try this out and give me any feedback this week. If all is well, we release at end of week or beginning of next?

---
I don't care which we use, but can we make these three sections' formatting match each other?

Screen Shot 2020-11-30 at 3 12 49 PM

---

When I switch networks in Metamask, my account information (above) does not reflect the change until I log out and log in again. Is it possible to make this react to Metamask (and other providers) switching networks dynamically?

--- 

The DAO dropdown should always be third in the breadcrumb, right after "All DAOs". We should never see this:

Screen Shot 2020-11-30 at 3 22 15 PM

roienatan:
@EzraWeller Regarding `Breadcrum`, as far as I see it appears to be a problem. The elements on the Breadcrum were not designed to be a `select` tag. The thing is that the Breadcrum order is determined automatically according the path, as it more deeper it will be on the right side. Now I could not find a way to disable the redirection when you click on the `select` tag, and the only way to disable it is to direct you the current page (it's kind of a hack) hence the current page is always the deepest and that's why it's always going to the right. It's quite difficult to explain but I hope it's clear, I'm still trying to figure out a way to bypass this behaviour.

roienatan:
@EzraWeller Regarding the network switching in MetaMask, I now realise that this issue is exist in Alchemy before, probably we didn't notice this because we never supported two and more networks (for example if you go to Alchemy on xdai, connect and then switch MetaMask to Rinkeby you'll see that the account menu still shows xdai and xgen).
Obviously now this is becoming a more important issue. I think the only way Alchemy can know when the user is switching network on MetaMask is by polling, just like we poll every 2 seconds to check if the user changed account. Maybe MetaMask provides a way to notify the app when the user changes account/network? it's something that need to be investigated, if so we can eliminate the polling also for the account change.

src/arc.ts
```diff
@@ -110,18 +116,41 @@ export function providerHasConfigUi(provider?: any): any | undefined {
   return provider && provider.isTorus;
 }
 
+async function getAllDaos(arc: Arc): Promise<any> {
```
Razzwan, line 119: Please, specify the retunring type here like:
```ts
async function getAllDaos(arc: Arc): Promise<Array<{id: string; name: string;}>> {
  // body
}
```

src/arc.ts
```diff
@@ -132,34 +161,45 @@ export async function initializeArc(provider?: any): Promise<boolean> {
     const readonly = typeof provider === "string";
 
     // if there is no existing arc, we create a new one
-    if ((window as any).arc) {
-      arc = (window as any).arc;
+    if ((window as any).arcs[network]) {
```
Razzwan, line 164: we can also specify this type globaly like this:
```ts
type IArc = any;
declare global {
    interface Window { arc: IArc; }
}
```
[more details here](https://stackoverflow.com/questions/12709074/how-do-you-explicitly-set-a-new-property-on-window-in-typescript)

src/arc.ts
```diff
@@ -132,34 +161,45 @@ export async function initializeArc(provider?: any): Promise<boolean> {
...
-        contractInfos = await arc.fetchContractInfos();
-      } catch (err) {
-        // eslint-disable-next-line no-console
-        console.error(`Error fetching contractinfos: ${err.message}`);
+
+    try {
+      contractInfos = await arc.fetchContractInfos();
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(`Error fetching contractinfos: ${err.message}`);
```
Razzwan, line 178: We should rather do something here. Let's discuss what should happen here

src/arc.ts
```diff
@@ -132,34 +161,45 @@ export async function initializeArc(provider?: any): Promise<boolean> {
...
+    }
+
+    const daos = await getAllDaos(arc);
+
+    if (daos !== undefined) {
+      const daosMap = {} as any;
```
Razzwan, line 184: Please, specify correct type here:
```ts
const daosMap: {[key in string]: string} = {};
```

src/arc.ts
```diff
@@ -132,34 +161,45 @@ export async function initializeArc(provider?: any): Promise<boolean> {
...
+      for (const dao of daos) {
+        daosMap[dao.id] = dao.name;
       }
+      (window as any).daos[network] = daosMap;
```
Razzwan, line 188: The same thing with type of window.daos. We can specify it like this:
```ts
type IDaos = any;
declare global {
    interface Window { daos: IDaos; }
}
```
for more details see above

src/arc.ts
```diff
@@ -195,27 +235,27 @@ export async function initializeArc(provider?: any): Promise<boolean> {
...
       if (networkName === "unknown (100)") {
         // we are fine, mayby
         return;
       }
     }
     // eslint-disable-next-line no-console
-    console.error(`connected to the wrong network, should be ${expectedNetworkName} (instead of "${networkName}")`);
-    throw new Error(`Please connect your wallet provider to ${expectedNetworkName}`);
+    console.error(`connected to the wrong network, should be ${expectedNetworkNames} (instead of "${networkName}")`);
+    throw new Error(`Please connect your wallet provider to ${(expectedNetworkNames as Array<any>).join(" or ")}`);
```
Razzwan, line 258: You are trhowing new error like this here:
```ts
throw new Error('');
```
but previously it was done like this:
```ts
throw Error('');
```
I recommend to use just only first variant everywhere to keep codebase consistent and easy to reead. And yes, I know, that both variants works the same, but code consistency is very important for such huge project

src/components/Dao/DaoContainer.tsx
```diff
@@ -54,13 +60,43 @@ const mapDispatchToProps = {
   showNotification,
 };
 
-class DaoContainer extends React.Component<IProps, null> {
+class DaoContainer extends React.Component<IProps, IState> {
+  constructor(props: any) {
```
Razzwan, line 64: props type here is IProps:
```ts
constructor(props: IProps) {}
```

src/components/Dao/DaoContainer.tsx
```diff
@@ -54,13 +60,43 @@ const mapDispatchToProps = {
...
+    };
+  }
   public daoSubscription: any;
   public subscription: Subscription;
 
   public async componentDidMount() {
     // TODO: use this once 3box fixes Box.getProfiles
     //this.props.getProfilesForAddresses(this.props.data[1].map((member) => member.staticState.address));
+    const { followingDaosAddresses } = this.props;
+    const memberQuery = gql` query memberDaos {
```
Razzwan, line 77: Why we are not using redux actions for getting information from the server in order to be able reuse the data which was already reseived in other place?

src/components/Daos/DaosPage.tsx
```diff
@@ -113,12 +120,40 @@ class DaosPage extends React.Component<IProps, IState> {
...
+    }
+
+    // Combine all sub-arrays into one array.
+    yourDAOs = [].concat(...yourDAOs);
+    otherDAOs = [].concat(...otherDAOs);
+
     // Always show DAOs that the current user is a member of or follows first
-    const yourDAOs = data[1].concat(data[2]).filter(d => d.staticState.name.toLowerCase().includes(search)).sort((a, b) => a.staticState.name.localeCompare(b.staticState.name));
+    yourDAOs = yourDAOs.filter(d => d.staticState.name.toLowerCase().includes(search)).sort((a, b) => a.staticState.name.localeCompare(b.staticState.name));
+    const tempMap = {} as any;
```
Razzwan, line 145: Please, specify correct type here:
```ts
const tempMap: {[key in string]: boolean} = {};
```

src/components/Daos/DaosPage.tsx
```diff
@@ -131,7 +166,7 @@ class DaosPage extends React.Component<IProps, IState> {
     // eslint-disable-next-line no-extra-boolean-cast
     if (process.env.SHOW_ALL_DAOS === "true") {
       // on staging we show all daos (registered or not)
-      otherDAOs = otherDAOs.filter((d: DAO) => !yourDAOAddresses.includes(d.id) && d.staticState.name.toLowerCase().includes(search));
+      otherDAOs = otherDAOs.filter((d: DAO) => !yourDAOAddresses.includes(d.id) && d.staticState !== undefined && d.staticState.name.toLowerCase().includes(search));
```
Razzwan, line 169: You can simplify this to:
```ts
otherDAOs = otherDAOs.filter((d: DAO) => !yourDAOAddresses.includes(d.id) && d.staticState?.name.toLowerCase().includes(search));
```

src/components/Daos/DaosPage.tsx
```diff
@@ -245,15 +278,21 @@ const createSubscriptionObservable = (props: IStateProps, data: SubscriptionData
...
+
+
+
+  const arcs = getArcs();
+  const daosData = [];
+
+  for (const network in arcs) {
+    const arc = arcs[network];
+    daosData.push(arc.daos({ orderBy: "name", orderDirection: "asc", first: PAGE_SIZE, skip: data ? data[0].length : 0 }, standardPolling(true)));
+    // eslint-disable-next-line @typescript-eslint/camelcase
```
Razzwan, line 290: I suggest to specify this setting globally if we using is such way

src/components/Daos/DaosPage.tsx
```diff
@@ -245,15 +278,21 @@ const createSubscriptionObservable = (props: IStateProps, data: SubscriptionData
...
+    daosData.push(followingDAOs.length ? arc.daos({ where: { id_in: followingDAOs }, orderBy: "name", orderDirection: "asc" }, standardPolling(true)) : of([]));
```
Razzwan, line 291: Why do we need push `of([])` to the list of daosData? Can we do like this:
```ts
if (followingDAOs.length) {
  daosData.push(
    arc.daos({
      where: { id_in: followingDAOs },
      orderBy: "name",
      orderDirection: "asc"
    },
    standardPolling(true))
  );
}
```

Razzwan (changes requested):

package.json
```diff
@@ -82,7 +83,7 @@
     "3box": "1.20.2",
     "@burner-wallet/burner-connect-provider": "^0.1.1",
     "@daostack/arc.js": "0.2.76",
-    "@dorgtech/daocreator-ui": "^1.0.13",
+    "@dorgtech/daocreator-ui": "^1.0.14",
```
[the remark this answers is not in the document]
roienatan: Actually I'm not sure why it's written with `^` from the first place. Is there a reason or I can eliminate? @orenyodfat

src/App.tsx
```diff
@@ -42,42 +43,47 @@ export class App extends React.Component<{}, {
...
+          this.setState({ retryingArc: true });
+          // eslint-disable-next-line no-console
+          numberOfAttempts += 1;
+          // retry
+          if (numberOfAttempts >= totalNumberOfAttempts) {
+            const msg = "Could not connect to the network; please retry later...";
+            this.setState({ error: msg});
+            throw Error(msg);
+          }
+          // eslint-disable-next-line no-console
```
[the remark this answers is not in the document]
roienatan: Where?

src/arc.ts
```diff
@@ -110,18 +116,41 @@ export function providerHasConfigUi(provider?: any): any | undefined {
...
+    const query = gql`
+    query AllDaos {
+      daos(first: ${MAX_BATCH_QUERY} skip: ${skip * MAX_BATCH_QUERY}) {
+        id
+        name
+      }
+    }
+  `;
+    const response = await arc.sendQuery(query, {});
+    daos = response.data.daos as any[];
```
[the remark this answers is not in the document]
roienatan: Actually this code supposed to be in `arc.js` and not here, but for now I've added an interface for that.

src/arc.ts
roienatan: I've added an interface and it looks like this `async function getAllDaos(arc: Arc): Promise<Array<IDAO>>`

src/arc.ts
roienatan: What can we do if not throwing an error?

src/arc.ts
roienatan: This is not related to this PR, we can do a separate code improvements. Please let's focus on changes related to this PR (otherwise we have many many things to fix in Alchemy code 😁)
You already commented so I'll make the change. So I didn't get you, you want to go with or without the `new`?

src/components/Dao/DaoContainer.tsx
roienatan: I totally agree with you! In many places we could reuse data if it was proper handled in the store from the first place.
In this specific case the only place we use this query is in the `DaosPage`, until now, where it's needed also here. The thing is that `DaoContainer` is not a descendant of `DaosPage` so it can't be passed through props. The only (and right) way to eliminate this code duplication is to fetch this query higher up in the hierarchy and save it in Redux store (or use React Context). It's necessary since if one goes straight to a DAO url, `DaosPage` will never be mounted hence the query will never be executed. This can be done in this PR or a separate one.

src/components/Daos/DaosPage.tsx
roienatan: What setting exactly?

package.json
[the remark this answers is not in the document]
orenyodfat: please remove `^`

src/arc.ts
Razzwan: You just use different approaches withing this exact PR

src/components/Daos/DaosPage.tsx
Razzwan: `@typescript-eslint/camelcase`

src/components/Dao/DaoContainer.tsx
Razzwan: OK, if you wish, we can do it as a part of this PR

Razzwan (approved):

orenyodfat (approved):
lets merge that and pr the daocreator update when it will be. ready https://github.com/dOrgTech/DAOcreator/issues/525#issuecomment-737346057.

PR #2265, opened 2020-11-16
Fix wrong share link

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2263

### DIFF
--- a/src/components/Proposal/ProposalDetailsPage.tsx
+++ b/src/components/Proposal/ProposalDetailsPage.tsx
@@ -5,7 +5,7 @@ import AccountProfileName from "components/Account/AccountProfileName";
 import ProposalCountdown from "components/Shared/ProposalCountdown";
 import FollowButton from "components/Shared/FollowButton";
 import { DiscussionEmbed } from "disqus-react";
-import { humanProposalTitle, ensureHttps, formatFriendlyDateForLocalTimezone, safeMoment, targetedNetwork } from "lib/util";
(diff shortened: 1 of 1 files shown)


src/components/Proposal/ProposalDetailsPage.tsx
```diff
@@ -336,7 +336,7 @@ class ProposalDetailsPage extends React.Component<IProps, IState> {
         {this.state.showShareModal ?
           <SocialShareModal
             closeHandler={this.closeShareModal}
-            url={`https://alchemy.daostack.io/dao/${daoState.address}/proposal/${proposal.id}`}
+            url={`https://${targetedNetwork() !== "main" ? `${targetedNetwork()}.` : ""}v1.alchemy.do/dao/${daoState.address}/proposal/${proposal.id}`}
```
orenyodfat, line 339: How about using env var BASE_URL here ?

orenyodfat (approved):

src/components/Proposal/ProposalDetailsPage.tsx
roienatan: We have a `BASE_URL` in `webpack.prod.config.js` but since we're adding text in the middle it's not usable.

orenyodfat: not sure why. ?
you can have a  util function `getBaseURL()`

roienatan: Let me try something 👍

orenyodfat (approved):

Issue #2247, opened 2020-11-09
Support an array of decimals in genericSchemeRegistry


roienatan:

dkent600:
CurveLabs needs this ASAP (as in, Nov 30).  Any status @roienatan ?

dkent600:
To clarify:  we want to be ale to specify "decimals", like `"decimals": 18` to be able to convert each number value in the array to WEI.

roienatan:
> CurveLabs/PrimeDAO need this ASAP (as in, Nov 30).  Any status @roienatan ?

Yes sorry I'm busy with more issues lately, I’ll try to see how to prioritize this.

dkent600:
@roienatan I'm going to have a go at this.

PR #2289, opened 2020-11-26
transform GenericScheme array item values just like field values

dkent600:
Resolves: https://github.com/daostack/alchemy/issues/2247

@orenyodfat  This adds the ability to transform items of an array the same way we transform regular field items  (I am referring to the configurable fields of a `GenericScheme` create-proposal form).

But this has the potential to break schemes having incorrectly-configured json.

The only example I can find:

This line:  https://github.com/daostack/alchemy/blob/5c5750474b617f40d8a0587cba78464769989c87/src/genericSchemeRegistry/schemes/TimelockManager.json#L142

... would currently be a no-op and not yield an array of WEI.  But with my new code, it would result in an array of items converted to WEI.

This is the only place I see that could have a problem with this code change.  The question is:  In the TimelockManager case, would the code change be improving their situation, or breaking it?


src/genericSchemeRegistry/index.ts
```diff
@@ -96,40 +96,57 @@ export class ActionField {
...
+     * Note that if this is an array item, the field's one `decimals` value applies
+     * to all items in the array.
+     */
+    else if (this.decimals) {
+      return (new BN(value as string).mul(new BN(10).pow(new BN(this.decimals)))).toString();
     }
+    else {
+      switch (this.transformation) {
+        case "namehash": {
+          return namehash.hash(value);
```
orenyodfat, line 119: why these additional spaces ?

dkent600: I'm not sure what spaces you are referring to here.  I know there were some added elsewhere in the file, I don't know whether that was my editor's preference or lint.  But they seem reasonable for readability.

orenyodfat: It might be your editor .can you try to remove it please ? see https://github.com/daostack/alchemy/pull/2289/files#diff-3f9aa40293ee5ff0e8a044c6f87937328712efc794fabbf134cb807019b09b5bR33. also

dkent600: ok I removed the whitespace changes

roienatan (approved):

orenyodfat:
@dkent600 please fix conflicts

dkent600:
> @dkent600 please fix conflicts

there aren't any conflicts. I hope we can merge before any appear  :-)

orenyodfat:
> > @dkent600 please fix conflicts
> 
> @orenyodfat there aren't any conflicts. I hope we can merge before any appear :-)

there are conflicts

Issue #1811, opened 2020-06-08
notifications with long messages look bad


dkent600:
![image](https://user-images.githubusercontent.com/1821666/84046678-6726ff00-a978-11ea-979a-efcfb181379d.png)

The 'X' button should always be by itself in the upper right of the window.  The window padding should always be the same on both right and left.

It can look even worse when there is a long unbroken string like a long hash which can greatly 
 overflow the notification window to the right.  Should be clipped off, or big _maybe_ the window should stretch horizontally to accommodate.

orenyodfat:
@vaske we accidentally merged your pr (and then revert it back);
please re submit it.

roienatan:
Sometimes it even goes out the modal:
![Screen Shot 2021-01-11 at 12 02 00](https://user-images.githubusercontent.com/34843014/104167808-a5e8d000-5405-11eb-9555-5756a7c58de3.png)

PR #2291, opened 2020-11-30
#1811 - fixed styling of notifications message modal

vaske:
This PR fix issue with positioning of 'X' on notifications; per #1811


roienatan:
I think as an ongoing process of using a single palette of colors let's try to use colors from here as scss variables (add if necessary) `src/assets/styles/global-variables.scss`

PR #2296, opened 2020-12-02
New proposal window with select

Razzwan:


orenyodfat:
@Razzwan can you please add more information re this pr ?
- why it is needed ? what does it solve?

Razzwan:
> @Razzwan can you please add more information re this pr ?
> 
> * why it is needed ? what does it solve?

It is not finished yet. Work in progress. Status of Pull Request is Draft. Will let you know once I am finished. Separate file is required to be able independently develop component without data context

src/components/Dao/DaoContainer.tsx
```diff
@@ -98,6 +102,28 @@ class DaoContainer extends React.Component<IProps, IState> {
     this.setState({ memberDaos: memberDaos });
   }
 
+  public handleNewProposal = async (schemeId?: string): Promise<void> => {
+    const { showNotification, daoAvatarAddress } = this.props;
+
+    if (!await enableWalletProvider({ showNotification }, getNetworkByDAOAddress(daoAvatarAddress))) { return; }
+
+    if (typeof schemeId === "string") {
```
orenyodfat, line 110: is there a case where schemeId is not a "string". ?
can this condition be `if (schemeId)` ?

Razzwan: Yes, there is a case, when schemeId is event. so we have to check, that type is exactly "string", or change the place, where we are using this function as event handler.

roienatan:
I guess we need a way to distinguish between plugins when we have the same type more than once? Maybe by ID? @orenyodfat @EzraWeller @Razzwan 
Screen Shot 2020-12-10 at 9 06 23

src/components/Dao/DaoContainer.tsx
roienatan: Can you explain/give an example?

roienatan:
No form is displayed, and also I guess it should be Competition title.
See this DAO to reproduce: https://alchemy-proposal-window-bp8lgv.herokuapp.com/dao/0x95f74bce31c67865d75c9038d6b23671ad5a0bda
Screen Shot 2020-12-10 at 9 11 28

roienatan (changes requested):
Why this file is here? `src/components/Proposal/ProposalHistoryRow.tsx`
Please delete this file, it was removed completely since the new proposals page has introduced.

src/components/Proposal/Create/CreateProposalPage.tsx
```diff
@@ -53,14 +49,17 @@ class CreateProposalPage extends React.Component<IProps, IStateProps> {
     };
   }
 
-  public handleClose = (e: any) => {
-    e.preventDefault();
+  public handleClose = (e?: any) => {
+    if (e?.preventDefault) {
```
roienatan, line 53: Why this `if` statement is necessary?

src/components/Proposal/Create/SelectProposal/SelectProposal.scss
```diff
@@ -0,0 +1,16 @@
+.body {
+  width: 100%;
+}
+
+.wrapper {
+  padding: 20px 30px 0;
+}
+
+.label {
+  color: #506176;
```
roienatan, line 10: Please use colors from `src/assets/styles/global-variables.scss` as a global variable.

roienatan (changes requested):

orenyodfat:
> I guess we need a way to distinguish between plugins when we have the same type more than once? Maybe by ID? @orenyodfat @EzraWeller @Razzwan
> Screen Shot 2020-12-10 at 9 06 23

no meed to distinguish . there is only a single type of plugin for each plugin . duplicate in the select is not needed though and need to be removed.

src/components/Dao/DaoContainer.tsx
Razzwan: [Here is the place where we are using this method as a event handler](https://github.com/daostack/alchemy/pull/2296/files/c460e9df9d144b1d8ed4889e329cb2ca89a847db#diff-8b2c028141c5d45e15a6a8b82d4edcdd55005613d480a0ce3ec57641b1237501R80)

[And this is a second place, where we are using this method as a function with string param](https://github.com/daostack/alchemy/pull/2296/files/c460e9df9d144b1d8ed4889e329cb2ca89a847db)

orenyodfat: are you sure the links above are correct ?

Razzwan: @orenyodfat fixed second link. Right now both are correct. If you don't see hightlight, please, try to little bit scroll up and you will see hightlighted line.

src/components/Proposal/Create/CreateProposalPage.tsx
Razzwan: Because e is not always exists. When we close window programmaticaly (instead of handle event) we are using this function without param. Any other suggestions?

Razzwan:
> No form is displayed, and also I guess it should be Competition title.
> See this DAO to reproduce: https://alchemy-proposal-window-bp8lgv.herokuapp.com/dao/0x95f74bce31c67865d75c9038d6b23671ad5a0bda
> Screen Shot 2020-12-10 at 9 11 28

Nice catch. Will take a look

src/components/Proposal/Create/SelectProposal/SelectProposal.scss
Razzwan: Done

Razzwan:
> > I guess we need a way to distinguish between plugins when we have the same type more than once? Maybe by ID? @orenyodfat @EzraWeller @Razzwan
> > Screen Shot 2020-12-10 at 9 06 23
> 
> yes, adding the address to the select along the name is good idea ( e.g `GenericScheme 0xabc...123`) @Razzwan

Added plugin id to select menu list

Razzwan:
> Why this file is here? `src/components/Proposal/ProposalHistoryRow.tsx`
> Please delete this file, it was removed completely since the new proposals page has introduced.

Removed. I forgot remove it after merge conflicts.

src/components/Proposal/Create/CreateProposalPage.tsx
Razzwan: @roienatan does it acceptable from you point of view?

orenyodfat:
> > > I guess we need a way to distinguish between plugins when we have the same type more than once? Maybe by ID? @orenyodfat @EzraWeller @Razzwan
> > > Screen Shot 2020-12-10 at 9 06 23
> > 
> > 
> > yes, adding the address to the select along the name is good idea ( e.g `GenericScheme 0xabc...123`) @Razzwan
> 
> Added plugin id to select menu list

1. please add space between the plugin name and the address .
2. no need to show all address just the 4 first  chars of it.   0xabcd... . you can use https://github.com/daostack/alchemy/blob/dev/src/lib/util.ts#L68  
Screen Shot 2020-12-10 at 21 26 11
@Razzwan

Razzwan:
> > > > I guess we need a way to distinguish between plugins when we have the same type more than once? Maybe by ID? @orenyodfat @EzraWeller @Razzwan
> > > > Screen Shot 2020-12-10 at 9 06 23
> > > 
> > > 
> > > yes, adding the address to the select along the name is good idea ( e.g `GenericScheme 0xabc...123`) @Razzwan
> > 
> > 
> > Added plugin id to select menu list
> 
> 1. please add space between the plugin name and the address .
> 2. no need to show all address just the 4 first  chars of it.   0xabcd... . you can use https://github.com/daostack/alchemy/blob/dev/src/lib/util.ts#L68
> 
> Screen Shot 2020-12-10 at 21 26 11
> 
> @Razzwan

Done 
image

Razzwan:
> No form is displayed, and also I guess it should be Competition title.
> See this DAO to reproduce: https://alchemy-proposal-window-bp8lgv.herokuapp.com/dao/0x95f74bce31c67865d75c9038d6b23671ad5a0bda
> Screen Shot 2020-12-10 at 9 11 28

Implemented in last commin

src/components/Proposal/Create/CreateProposalPage.tsx
roienatan: It's okay I just wonder since it never caused a problem (at least from what I witnessed) probably it's necessary because of this new feature.

roienatan:
There is a small issue I've found: when you switch to a scheme from the same type the form is not updated.
Steps to reproduce:

1. Go to https://alchemy-proposal-window-bp8lgv.herokuapp.com/dao/0x95f74bce31c67865d75c9038d6b23671ad5a0bda
2. Click on create a proposal.
3. Choose `GenericSchemeMultiCall 0xacdc32b3d5f22c7f043d7dc3e4096fd22852cb9f5bc4d6f8235c9be70b4d8b8b`
Check the field "Contract Address" and you will see it has 3 contracts. Same for `GenericSchemeMultiCall  0x26333cf21051f3c1cd1151f7da9c251a29662f88ceb1a52e47cd1b454debb665` but it's not correct since `0xacdc32b3d5f22c7f043d7dc3e4096fd22852cb9f5bc4d6f8235c9be70b4d8b8b` has only 1 contract which is `0x000...00`

You can verify it by going to the proposal modal via the plugins page:
https://alchemy-proposal-window-bp8lgv.herokuapp.com/dao/0x95f74bce31c67865d75c9038d6b23671ad5a0bda/scheme/0xacdc32b3d5f22c7f043d7dc3e4096fd22852cb9f5bc4d6f8235c9be70b4d8b8b/proposals/create/

and 

https://alchemy-proposal-window-bp8lgv.herokuapp.com/dao/0x95f74bce31c67865d75c9038d6b23671ad5a0bda/scheme/0x26333cf21051f3c1cd1151f7da9c251a29662f88ceb1a52e47cd1b454debb665/proposals/create/

roienatan:
Let's give the select container some bottom padding and box shadow so it looks nicer when you scroll, and instead of this:

Screen Shot 2020-12-11 at 19 09 51

It will look something like this:

Screen Shot 2020-12-11 at 19 13 01

roienatan:
When pressing on the new proposal button there are two loaders.

image

roienatan:
> > > > > I guess we need a way to distinguish between plugins when we have the same type more than once? Maybe by ID? @orenyodfat @EzraWeller @Razzwan
> > > > > Screen Shot 2020-12-10 at 9 06 23
> > > > 
> > > > 
> > > > yes, adding the address to the select along the name is good idea ( e.g `GenericScheme 0xabc...123`) @Razzwan
> > > 
> > > 
> > > Added plugin id to select menu list
> > 
> > 
> > 
> > 1. please add space between the plugin name and the address .
> > 2. no need to show all address just the 4 first  chars of it.   0xabcd... . you can use https://github.com/daostack/alchemy/blob/dev/src/lib/util.ts#L68
> > 
> > Screen Shot 2020-12-10 at 21 26 11
> > @Razzwan
> 
> Done
> image

Let's try to be consistent with other places in Alchemy and show the first 4 and last 4 characters of the scheme address. There is a function that does it for you, see `schemeNameAndAddress` in `src\lib\schemeUtils.ts`.
Also, I would put the scheme address inside a brackets like this `(0xacdc32b3d5f22c7f043d7dc3e4096fd22852cb9f5bc4d6f8235c9be70b4d8b8b)`

src/components/Proposal/Create/CreateProposalPage.tsx
Razzwan: @roienatan yes, exactly

orenyodfat:
still seeing the loader 

Screen Recording 2020-12-13 at 9 22 00
Screen Recording 2020-12-13 at 9 22 00

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
```diff
@@ -135,6 +135,14 @@ class CreateGenericMultiCallScheme extends React.Component<IProps, IStateProps>
     };
   }
 
+  componentDidUpdate(prevProps: Readonly<IProps>) {
```
roienatan, line 138: That's a specific solution for `multiCallGenericScheme`.
A generic solution is better, that will re-render the form when we switch.

Razzwan: How does it work?

src/assets/styles/global-variables.scss
```diff
@@ -23,6 +23,7 @@ $accent-3-halftone: rgba(255, 0, 72, 0.5);
 $accent-4: rgba(46, 168, 122, 1);
 $accent-4-halftone: rgba(46, 168, 122, 0.5);
 $accent-4-transparent: rgba(46, 168, 122, 0);
+$gray-label: #506176;
```
roienatan, line 26: Instead of adding new color, let's just use an existing grey one, for example `$gray-1`

src/components/Proposal/Create/SchemeForms/CreateGenericMultiCallProposal.tsx
roienatan: What do you mean? 
Maybe I'm wrong but as I see this, the problem is when you switch in the select to a different scheme (from the same type) the form is not re-rendered. If it's the problem so it should be fixed generically maybe here `SelectProposal.tsx`.
Because this is very specific to `multiCallGenericScheme`.

orenyodfat (approved):
approve , though lets fix #2324 just after that.

roienatan (approved):

PR #2301, opened 2020-12-03
Brand new Proposals Page

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2279
Figma: https://www.notion.so/daostack/Product-Roadmap-0062c4863b684509b1aa9e7757ed22af?p=389fe3a94bb54249b5965c6d29b08b34


src/components/Dao/DaoProposalsPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+
+const proposalsQuery = (dao: IDAOState, skip: number, titleSearch?: string): Observable<Array<Proposal>> => {
+  const filter: any = {
+  };
+
+  let orderDirection: OrderDirection = "asc";
+
+  if (!titleSearch) {
+    if (skip % PAGE_SIZE === 0) {
```
Razzwan, line 34: Could you please explain this logic? Looks like a potential problem

src/components/Dao/DaoProposalsPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+    } else {
+      filter["title_contains"] = titleSearch;
+    }
+  }
+
+  return dao.dao.proposals({
+    where: filter,
+    orderBy: "closingAt",
+    orderDirection: orderDirection,
+    first: titleSearch ? undefined : PAGE_SIZE, // TEMPORARY UNTIL WE PASS "titleSearch" in line 143
```
Razzwan, line 55: Why this string? Looks quite wired. Could you please explain this temporary logic?

src/components/Dao/DaoProposalsPage.tsx
```diff
@@ -0,0 +1,152 @@
...
+    <div className={css.wrapper}>
+      <BreadcrumbsItem to={`/dao/${daoState.address}`}>Proposals</BreadcrumbsItem>
+      <div className={css.topBarWrapper}>
+        <div className={css.top}>
+          <h1 className={css.title}>Proposals</h1>
+          <div
+            // Temporary until https://github.com/daostack/alchemy/issues/2292 is ready.
+            style={{ pointerEvents: "none", backgroundColor: "gray"}}
+            className={css.createProposalButton}
+            data-test-id="createProposal">
```
Razzwan, line 97: Why do we need this?

Razzwan (approved):

docs/fetchingandsubscribingdata.md
```diff
@@ -14,7 +14,7 @@ Executing these four steps for each React component is quite expensive (just thi
 So the approach we are taking in Alchemy is the following.
 
 *to avoid the cost of step 2.:*
-Fetch the data that is used by many subcomponents in a single query, before those components are rendered. When those subcomponents are rendered, they will be able to get their data directly  from the cache, and so will avoid a server request in step 2. An example of that pattern is found in the `DaoHistoryPage.tsx`
+Fetch the data that is used by many subcomponents in a single query, before those components are rendered. When those subcomponents are rendered, they will be able to get their data directly  from the cache, and so will avoid a server request in step 2.
```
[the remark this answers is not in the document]
roienatan: DaoHistoryPage is deprecated and is removed from Alchemy. The new proposals page includes also the history proposals.

src/components/Dao/DaoProposalsPage.tsx
roienatan: You mean the grey button? It's temporary until your PR will be ready and we can open the new modal from this button.

src/components/Dao/DaoProposalsPage.tsx
roienatan: `skip` is the size of the total data (proposals) that we currently show on the UI (the proposals that we've received so far).
if `skip % PAGE_SIZE !== 0` it means that we have received less proposals then the page size in the last query, meaning we finished to fetch all the proposals with `closingAt` greater than `0` and we can change the filter to fetch the rest of the proposals. I do agree that it might be a problem if the last chank of proposals will be at size of `PAGE_SIZE`. It's kind of a hack to determine when we received all the data, and it should be improved to be more robust. @orenyodfat @Razzwan

src/components/Dao/DaoProposalsPage.tsx
roienatan: This is a temporary hack until I'll figure out a way to pass the `titleSearch` string into here, since it's necessary

PR #2322, opened 2020-12-10
whitelist PrimeDAO Token 

dkent600:


orenyodfat:
only rinkeby ?
why all these commits ?

dkent600:
frack, thank you for catching rinkeby versus mainnet.  I moved it to mainnet

dkent600:
as for the commits, no idea, but there is only one file change, that is what matters, yes?

orenyodfat (approved):

Issue #2326, opened 2020-12-13
use subgraph v41_3 


orenyodfat:
for enable sorting proposal by closingAt field

PR #2328, opened 2020-12-13
Sort Proposals

roienatan:
resolves: https://github.com/daostack/alchemy/issues/2326

Sort proposals by the next order: executables, passing and falling by `closingAt` in ascending order, and then failed and executed proposals.

### DIFF
--- a/test/integration/proposal-genericSchemeEnsRegistry.e2e.ts
+++ b/test/integration/proposal-genericSchemeEnsRegistry.e2e.ts
@@ -1,6 +1,6 @@
 import * as uuid from "uuid";
 import { first } from "rxjs/operators";
-import { getArc, gotoDaoSchemes } from "./utils";
+import { getArc, gotoDaoSchemes, hideCookieAcceptWindow } from "./utils";
 
 describe("Proposals ENS Registry", () => {
   let daoAddress: string;
@@ -16,6 +16,7 @@ describe("Proposals ENS Registry", () => {
 
   it("Create a Generic Scheme ENS Registry proposal and check that the data is submitted correctly", async () => {
     await gotoDaoSchemes(daoAddress);
+    await hideCookieAcceptWindow();
 
     const ensTitle = await $("h2=GenericSchemeENSRegistry");
     await ensTitle.waitForExist();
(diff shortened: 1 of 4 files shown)


orenyodfat:
filter executed proposal here https://github.com/daostack/alchemy/blob/dev/src/components/Redemptions/RedemptionsPage.tsx#L250

test/integration/proposal-genericSchemeEnsRegistry.e2e.ts
```diff
@@ -16,6 +16,7 @@ describe("Proposals ENS Registry", () => {
 
   it("Create a Generic Scheme ENS Registry proposal and check that the data is submitted correctly", async () => {
     await gotoDaoSchemes(daoAddress);
+    await hideCookieAcceptWindow();
```
orenyodfat, line 19: why is this needed ?

roienatan: For some reason this test failed on Travis and locally because the cookies window blocked it. Anyway it's harmless it just hides the cookies prompt.

roienatan:
> filter executed proposal here https://github.com/daostack/alchemy/blob/dev/src/components/Redemptions/RedemptionsPage.tsx#L250

I think we should leave it as it is, with this change the redeem test fails because it has nothing to redeem. It already fetches the proposals with this filter `accountsWithUnclaimedRewards_contains`. I also verified it on a proposal on Rinkeby, with this filter I have nothing to redeem but without it I can see what I can redeem. Let me know what you think.

Razzwan (approved):

test/integration/proposal-genericSchemeEnsRegistry.e2e.ts
Razzwan: @roienatan it can be a problem for other tests as well? Can you show the video, how does it looks like?

Razzwan (approved):

test/integration/proposal-genericSchemeEnsRegistry.e2e.ts
roienatan: As I saw it's not consistent. Will explain you when we talk.

PR #2336, opened 2020-12-15
feat: hide add proposal button if empty schemes array

Razzwan:


roienatan (changes requested):
Here for example (https://alchemy-hide-add-propos-fauxpj.herokuapp.com/dao/0x6a041608fa575168957e53c6b6db9a76771fe011), the plugin is not supported and the New Proposal button is stuck.
I think we should not consider an unsupported scheme in the our count. Please suggest @orenyodfat

Razzwan:
> Here for example (https://alchemy-hide-add-propos-fauxpj.herokuapp.com/dao/0x6a041608fa575168957e53c6b6db9a76771fe011), the plugin is not supported and the New Proposal button is stuck.
> I think we should not consider an unsupported scheme in the our count. Please suggest @orenyodfat

Agreed, will doing

roienatan:
@Razzwan It's showing `0` instead of empty space, please fix.
image

Razzwan:
> @Razzwan It's showing `0` instead of empty space, please fix.
> image

Fixed

roienatan (approved):

PR #2340, opened 2020-12-15
Release 1.0.4

roienatan:


Razzwan (changes requested):
Please, add this PR to current Release: https://github.com/daostack/alchemy/pull/2341

roienatan:
There is a small issue that sometimes the form is not getting updated when you change the scheme in the Select. This will be fixed by @Razzwan and we'll update the PR. All other functionalities can be reviewed.

Razzwan (dismissed):

EzraWeller:
Issues
Ideally, we can address all of these in this release. If any are particularly thorny/giving you trouble, please let me know!
---
### 1)
Disqus not loading on proposal pages:
Screen Shot 2020-12-15 at 3 08 14 PM

### 2)
Pending boost proposals should be displayed in the new proposal list by their _regular queue_ voting time, not their pending boost time, i.e. the time they will _resolve_ is the key time for ordering the list. The first proposal on the list below is an example:
Screen Shot 2020-12-15 at 3 15 39 PM

### 3)
Small CSS problem. I think the account pop up is showing up behind the left side bar here:
Screen Shot 2020-12-15 at 3 20 00 PM

### 4)
Can we make the proposal search bar not case sensitive? Don't think we want case sensitivity here.
Screen Shot 2020-12-15 at 3 37 32 PM

### 5)
a. My top preference for the search bar would be for it to update as you type and not require hitting enter. If this isn't possible do to queries taking a long time:
b. My second preference would be to separate it into a search box and a "Search" button that triggers the query as well as hitting return/enter. I think this is the standard pattern for searches like this.
Let's also change the prefilled text to just `Search by title or proposer address`.
Screen Shot 2020-12-15 at 3 38 32 PM

### 6)
The height of the proposal form is different for `Funding and Voting Power` than other proposal types right now. Can we make all the proposals forms stay the same height? (we can let width change when the sidebar is needed as it is now)
Screen Shot 2020-12-15 at 3 43 12 PM
Screen Shot 2020-12-15 at 3 43 27 PM

### 7)
Two small proposal form behavior changes.
1. Allow the description box to be extendable by dragging the corner. This is a standard pattern, so if you aren't familiar with it, please talk to me before starting implementation so we can make sure we're on the same page.
2. Only close the proposal form if the user clicks the "x" in the top right corner of the form, rather than if they click anywhere outside the form. Too many people complaining that they are accidentally closing the form and losing progress!

### 8)
Sorry to say it, but we should remove the Alchemy 2 announcement bars:
Screen Shot 2020-12-15 at 3 56 01 PM
Screen Shot 2020-12-15 at 3 56 16 PM

### 9)
Non-boosted proposals with less than 50% "Yes" votes are _not_ passing and shouldn't read as passing: they should read as failing. This is because they will fail unless boosted or reaching absolute majority. The two top proposals below are examples (they have only ~5% yes votes). Let me know if that doesn't make sense!
Screen Shot 2020-12-15 at 4 05 59 PM

### 10)
Trying to connect to Alchemy using Wallet Connect from the Metamask mobile app (and also the Gnosis Safe legacy mobile app, as reported by DXdao but no long testable), is giving this error:
Screen Shot 2020-12-15 at 4 20 13 PM

### 11)
stil testing

roienatan:
@EzraWeller Regarding number 6, if we will make all proposal modals the same hight, you might see big white void for a proposal with short content. Currently the height is adapting be the content.

EzraWeller:
> @EzraWeller Regarding number 6, if we will make all proposal modals the same hight, you might see big white void for a proposal with short content. Currently the height is adapting by the content.

Can't we just extend the content width to the fit the box? If this is a ton of work, then we can skip this one.

roienatan:
> > @EzraWeller Regarding number 6, if we will make all proposal modals the same hight, you might see big white void for a proposal with short content. Currently the height is adapting by the content.
> 
> Can't we just extend the content width to the fit the box? If this is a ton of work, then we can skip this one.

So we're talking about the width or the hight?

EzraWeller:
> > > @EzraWeller Regarding number 6, if we will make all proposal modals the same hight, you might see big white void for a proposal with short content. Currently the height is adapting by the content.
> > 
> > 
> > Can't we just extend the content width to the fit the box? If this is a ton of work, then we can skip this one.
> 
> So we're talking about the width or the hight?

Oh, right, sorry. We're talking about height. It's fine to have some white space or extra scroll, imo.

roienatan:
> > > > @EzraWeller Regarding number 6, if we will make all proposal modals the same hight, you might see big white void for a proposal with short content. Currently the height is adapting by the content.
> > > 
> > > 
> > > Can't we just extend the content width to the fit the box? If this is a ton of work, then we can skip this one.
> > 
> > 
> > So we're talking about the width or the hight?
> 
> Oh, right, sorry. We're talking about height. It's fine to have some white space or extra scroll, imo. White space should always be at the bottom, right?

Yes.

roienatan:
Resolved now: (3) (7.2) (8) (13) (9)

EzraWeller:
So unfortunately, wallet connect still seems not to be working. I can connect with it successfully, but transactions are not appearing on my phone when I try to propose, vote, or stake on mainnet. I'm also not able to successfully connect to xDai with wallet connect (on the main release deployment), though that could possibly be something wrong with my mobile metamask's connection to xDai.

orenyodfat:
> So unfortunately, wallet connect still seems not to be working. I can connect with it successfully, but transactions are not appearing on my phone when I try to propose, vote, or stake on mainnet.
> 
> I'm also not able to successfully connect to xDai (on the main release deployment) or rinkeby with wallet connect at all.
> 
> Second issue: the main proposals page for DAOs switches very slowly when you use the drop down to switch DAOs. The slow loading isn't a huge problem, but because it doesn't indicate it's loading, it's easy to think that the switch failed. Perhaps we can put some kind of "loading" message instead? In general, loading times are seeming quite long. Maybe we should leave this for now and see if we can do some optimization while I'm gone for the holidays?
> 
> Last thing I've found: breadcrumb seems to be fairly broken.
> Screen Shot 2020-12-22 at 8 26 22 PM

thanks @EzraWeller 
1. wallet connect issue cannot be reproduced.
2. loader was added to the `daos switch`. - we will try to optimise the timing though not for this release
3. the breadcrumb cannot be reproduced. is there any steps for doing that ?

EzraWeller:
> > So unfortunately, wallet connect still seems not to be working. I can connect with it successfully, but transactions are not appearing on my phone when I try to propose, vote, or stake on mainnet.
> > I'm also not able to successfully connect to xDai (on the main release deployment) or rinkeby with wallet connect at all.
> > Second issue: the main proposals page for DAOs switches very slowly when you use the drop down to switch DAOs. The slow loading isn't a huge problem, but because it doesn't indicate it's loading, it's easy to think that the switch failed. Perhaps we can put some kind of "loading" message instead? In general, loading times are seeming quite long. Maybe we should leave this for now and see if we can do some optimization while I'm gone for the holidays?
> > Last thing I've found: breadcrumb seems to be fairly broken.
> > Screen Shot 2020-12-22 at 8 26 22 PM
> 
> thanks @EzraWeller
> 
> 1. wallet connect issue cannot be reproduced.
> 2. loader was added to the `daos switch`. - we will try to optimise the timing though not for this release
> 3. the breadcrumb cannot be reproduced. is there any steps for doing that ?

Okay, both 1 and 3 seem to be fixed for me this morning. Still some strange behavior sometimes -- when a proposal opens in a new tab, got some strange behavior around the web3 connection -- but I think this is releasable.

EzraWeller (dismissed):

orenyodfat (approved):

PR #2350, opened 2020-12-17
fix: fix height of proposal form

Razzwan:


roienatan (changes requested):
Looks bad. 
Screen Shot 2020-12-17 at 15 54 47

Razzwan:
> Looks bad.
> Screen Shot 2020-12-17 at 15 54 47

What is your suggestion?

roienatan:
> > Looks bad.
> > Screen Shot 2020-12-17 at 15 54 47
> 
> What is your suggestion?

We can't use this, here I can not even see the title of the modal.
This is really not urgent issue, currently the modal size is changed by it's content and it looks good. Also according to Ezra it's not important issue now (see his updated comment in the release)

PR #2353, opened 2020-12-17
Preparation for logical structure

Razzwan:
This is the first step **Prepare redux store structure** of implementation smart proposal serch from this steps

### 1. Prepare redux store structure
2. Create reducer for storing proposals in the redux store
3. Add browser cache to be able restore data from browser cache first
4. Save proposals list to redux store and subscribe redux store to graphQL endpoints instead of subscribing each separate component
5. Add lodash library to be able implement smart search by proposals using local storage first
6. Test everything


src/components/Proposal/Create/CreateProposal.scss
```diff
@@ -38,9 +38,9 @@ fieldset {
 }
```
orenyodfat, line 38: does this update is related to this pr ?

orenyodfat (approved):

PR #2359, opened 2020-12-20
update web3 to 1.3.1

### LINKED ISSUE #2354: wallet-connect does not work
some instruction here : https://walletconnect.org

orenyodfat:
fix #2354

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -382,6 +382,10 @@ export async function getNetworkName(id?: string): Promise<Networks> {
     id = await getNetworkId();
   }
 
+  if (typeof id !== "string") {
+    id = "" + id;
+  }
+
   switch (id) {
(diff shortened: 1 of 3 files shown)


src/lib/util.ts
```diff
@@ -382,6 +382,10 @@ export async function getNetworkName(id?: string): Promise<Networks> {
     id = await getNetworkId();
   }
 
+  if (typeof id !== "string") {
```
roienatan, line 385: Why this is necessary?

orenyodfat: because there are cases where the id is not a string . which comes from https://github.com/daostack/alchemy/blob/dev/src/lib/util.ts#L280 .  when connecting to WalletConnect

roienatan (approved):

PR #2381, opened 2021-01-04
Support CL4R

roienatan:
resolves: #2386
resolves:  #1811
resolves:  #2385


orenyodfat:
please use https://github.com/daostack/arc.js/releases/tag/0.2.83

orenyodfat:
information tab : 
- [ ] token : add name of token near the address
- [ ] batch time : should be in days units 
- [ ] constA should be displayed in  a readable number  format e.g ( 85000 REP) 
- [ ] constB should be displayed in readable decimal format e.g (0.9) .

Screen Shot 2021-01-21 at 20 54 56

orenyodfat:
- [ ] Something the locks can be shown even the user is not login .
- [ ] `Please Login` should be something like 'Please Login to your wallet to see your locks' 

Screen Shot 2021-01-21 at 21 00 54

orenyodfat:
convert const A + B from realNumbers 
so in this case
-  const A should be 85000 rep 
- const B should be 0.9 
Screen Shot 2021-01-23 at 0 24 39

orenyodfat:
wrong (active) date 

Screen Shot 2021-01-23 at 0 27 06

roienatan:
> wrong active/inactive date 
> 
> in this case it should be `active` 
> 
> 
> 
> Screen Shot 2021-01-23 at 0 27 06
> 
> 

The lock has ended so it's inactive. Maybe it's not the best place to put it.

orenyodfat (approved):

PR #2383, opened 2021-01-05
whitelist PrimeDAO Token - II

dkent600:
Trying again


roienatan:
See https://github.com/daostack/alchemy/pull/2384

